PR #10, opened 2015-10-30
Add a batch script that will build all the projects and run all the t…

SurferJeffAtGoogle:
…ests.

Will be invoked from Jenkins.


remi:
:+1:

PR #18, opened 2015-12-05
Adding Auth sample

jsimonweb:


auth/Program.cs
```diff
@@ -0,0 +1,140 @@
...
+    public class AuthSample
+    {
+        private const string usage = @"Usage:AuthSample <bucket_name>";
+        // [START build_service]
+        /// <summary>
+        /// Creates an authorized Cloud Storage client service using Application 
+        /// Default Credentials.
+        /// </summary>
+        /// <returns>an authorized Cloud Storage client.</returns>
+        public static async Task<StorageService> CreateAuthorizedClientAsync()
```
SurferJeffAtGoogle: The new policies we adopted say no async methods in csharp-docs samples.
Also, drop private and static where ever possible.

SurferJeffAtGoogle:
Update buildAndRunTests.bat.

auth/Program.cs
remi: `CodeFormatter.exe` adds explicit visibility keywords.  Is there a `/rule:` option to *not* do this?

auth/Program.cs
```diff
@@ -0,0 +1,141 @@
...
+        }
+        // [END build_service]
+
+        // [START list_storage_bucket_contents]
+        /// <summary>
+        /// List the contents of a Cloud Storage bucket.
+        /// </summary>
+        /// <param name="bucket">the name of the Cloud Storage bucket.</param>
+        ///<returns>a list of the contents of the specified bucket.</returns>
+        public Objects ExecuteApiRequest(
```
SurferJeffAtGoogle: Can we rename this to something more specific like ListBuckets()?  If you're trying to communicate that this function is the one that actually calls our API, then maybe name it CallListBuckets()?

SurferJeffAtGoogle:
LGTM.

PR #26, opened 2016-05-20
Add an Export Table sample.

SurferJeffAtGoogle:
For https://cloud.google.com/bigquery/exporting-data-from-bigquery

Also, drop the BigQueryUtil because that only made the sample more confusing.


jsimonweb:
LGTM

PR #29, opened 2016-05-31
Adding Cloud Vision label detection sample

jsimonweb:


vision/labelDetection/Program.cs
```diff
@@ -0,0 +1,141 @@
...
+                        + label.Score + ")");
+ 
+            }
+            else
+            {
+                Console.WriteLine("No labels found.");
+            }
+            // [END parse_response]
+            Console.WriteLine("Press any key...");
+            Console.ReadKey();
```
SurferJeffAtGoogle: This is redundant with SamplesUtil.InvokeMain().  I removed SamplesUtil.InvokeMain() from another sample because I think it adds more distraction than it's worth.   Let's do the same here:  Keep Console.WriteLine() and Console.ReadKey(), and make this MainFunction the true Main().

SurferJeffAtGoogle:
Looks very nice.

SurferJeffAtGoogle:
LGTM.

PR #42, opened 2016-07-15
Copy BuildTools from getting-started-dotnet and add runTests.ps1 file…

SurferJeffAtGoogle:
…s in

subdirectories.

This means we will start running some tests that haven't been running
because they weren't MSTests.


jsimonweb:
LGTM

PR #43, opened 2016-07-15
Use _$folder$ suffix instead of / suffix to tag directories.

SurferJeffAtGoogle:
Follow the convention used by hadoop and other tools.

Also, make the test more resilient to semantically meaningless changes
by sorting the output before comparing.


jsimonweb:
LGTM

PR #57, opened 2016-09-06
Add pubsub

jsimonweb:


pubsub/api/PubsubTest.cs
```diff
@@ -0,0 +1,410 @@
...
+            string subscriptionId, 
+            SubscriberClient subscriber, 
+            PullResponse response)
+        {
+            string subscriptionName =
+                SubscriberClient.FormatSubscriptionName(_projectId, 
+                subscriptionId);
+            // [START pull_messages]
+            subscriber.Acknowledge(subscriptionName, 
+                response.ReceivedMessages.Select(m => m.AckId));
```
SurferJeffAtGoogle: Nice use of LINQ here.

pubsub/api/PubsubTest.cs
```diff
@@ -0,0 +1,410 @@
...
+            // [END pull_messages]
+        }
+
+        public Topic GetTopic(string topicId, PublisherClient publisher)
```
SurferJeffAtGoogle: GetTopic() and GetSubscription() each have only one call site.  Will later samples invoke them too?

pubsub/api/PubsubTest.cs
```diff
@@ -0,0 +1,410 @@
...
+        // [END retry]
+
+        private static bool IsEmptyResponse(PullResponse response)
+        {
+            foreach (var result in response.ReceivedMessages)
+                return false;
+            return true;
+        }
+
+        private static bool IsEmptyTopicsList(IEnumerable<Topic> topics)
```
SurferJeffAtGoogle: I think readers will be ok with seeing `topics.Count == 0`.  What do you think of dropping this function?

pubsub/api/PubsubTest.cs
```diff
@@ -0,0 +1,410 @@
...
+        {
+            foreach (var topic in topics)
+                return false;
+            return true;
+        }
+
+        private static bool IsEmptySubscriptionsList(IEnumerable<Subscription>
```
SurferJeffAtGoogle: ditto

pubsub/api/PubsubTest.cs
```diff
@@ -0,0 +1,410 @@
...
+            string topicName = PublisherClient.FormatTopicName(_projectId, topicId);
+            string subscriptionName =
+                SubscriberClient.FormatSubscriptionName(_projectId, subscriptionId);
+            RetryRpc(() =>
+            {
+                //Create Topic only on second try
+                if (tryCount++ == 1)
+                {
+                    //Create Topic
+                    publisher.CreateTopic(topicName);
```
SurferJeffAtGoogle: I think I'd rather see the two API calls in two separate calls to RetryRpc().  That should also obviate tryCount.

pubsub/api/PubsubTest.cs
jsimonweb: Yes was planning to use these functions as part of testing IAM permissions for subscriptions and for a topics.

pubsub/api/PubsubTest.cs
```diff
@@ -0,0 +1,412 @@
...
+            string topicName = PublisherClient.FormatTopicName(_projectId, topicId);
+            string subscriptionName =
+                SubscriberClient.FormatSubscriptionName(_projectId, subscriptionId);
+            try
+            {
+                RetryRpc(() =>
+                {
+                    // Subscribe to Topic
+                    // This should fail since Topic has not yet been created
+                    subscriber.CreateSubscription(subscriptionName, topicName,
```
SurferJeffAtGoogle: I see why we want to have this code in the test, so that we can test the retry behavior and make sure an exception gets appropriately thrown.

However, I'm not sure this code should live between the // [START retry] and // [END retry] tags.

pubsub/api/PubsubTest.cs
```diff
@@ -0,0 +1,412 @@
...
+                });
+            }
+            catch (AggregateException exceptions)
+            {
+                if (exceptions.InnerExceptions.Count() == _retryCount)
+                {
+                    RetryRpc(() =>
+                    {
+                        //Create Topic
+                        publisher.CreateTopic(topicName);
```
SurferJeffAtGoogle: What does publisher.CreateTopic() do if the topic already exists?

pubsub/api/PubsubTest.cs
```diff
@@ -0,0 +1,445 @@
...
+                    // [START retry]
+                    try
+                    {
+                        RetryRpc(() =>
+                        {
+                            //Create Topic
+                            publisher.CreateTopic(topicName);
+                        });
+                    }
+                    catch (AggregateException ae)
```
SurferJeffAtGoogle: If the topic already exists, this code will try to create it 3 times.  Instead, it should observe the StatusCode=AlreadyExists in the first exception and not try twice more.

Also, I think the API could be better here.  We should not have to parse text strings to observe this.  Text strings notoriously change from version to version, and are generally fragile.  File a bug on https://github.com/GoogleCloudPlatform/google-cloud-dotnet and ask for a better API.

pubsub/api/PubsubTest.cs
```diff
@@ -0,0 +1,444 @@
...
+                        (e.Status.StatusCode != StatusCode.AlreadyExists)
+                    {
+                        // A StatusCode of "AlreadyExists" is ok.  
+                        // It means the Subscription already exists. 
+                        // Otherwise throw exception.
+                        throw;
+                    }
+                }
+            }
+        }
```
SurferJeffAtGoogle: Yikes.  The exception handling and retry code in this sample now totals nearly of 100 lines of code.  That's a lot for users to digest.  I think we can simplify it if we actually drop the RetryRpc() function altogether, and inline the exception handling code.  Or do you think that will make the code even messier?

SurferJeffAtGoogle: See https://github.com/SurferJeffAtGoogle/csharp-docs-samples/commit/56810a7826528d4c5bd68bcda5d5a8bc1430ca51

jsimonweb: In-lining the exception handling code might make it even longer b/c currently RetryRpc() allows CreateTopic and CreateSubscription to share the RPC looping/retry functionality. I'll try out in-lining to see if something more succinct is possible while still demonstrating RPC retries (which is a required code sample section in the current documentation).

SurferJeffAtGoogle:
LGTM.

PR #75, opened 2016-10-25
More storage samples with tests.  Still not complete.

SurferJeffAtGoogle:
I refactored a lot of the test code so that the test bucket is
automatically cleaned up at the end of the test.

Also, tests were flaky because there was some kind of race condition
around Console.Out.  So, I changed the code so that you pass Console.Out
to Main().  No more race condition.


jsimonweb:
LGTM

PR #79, opened 2016-11-01
Demonstrate using .AsPages().

SurferJeffAtGoogle:
DO NOT MERGE.

### DIFF
--- a/pubsub/api/PubsubTest.cs
+++ b/pubsub/api/PubsubTest.cs
@@ -21,6 +21,8 @@ using Google.Protobuf;
 using Grpc.Core;
 using System.Collections.Generic;
 using Xunit;
+using Google.Api.Gax;
+using System.Diagnostics;
 
 namespace GoogleCloudSamples
 {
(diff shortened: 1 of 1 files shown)


pubsub/api/PubsubTest.cs
```diff
@@ -335,6 +337,32 @@ public void TestListTopics()
...
+        public void TestListTopicPages()
+        {
+            string topicId = "testTopicForListingTopics";
+            string topicName = PublisherClient.FormatTopicName(_projectId, topicId);
+            for (int i = 0; i < 5; ++i)
+            {
+                CreateTopic($"{topicId}{i}", _publisher);
+            }
+            string projectName = PublisherClient.FormatProjectName(_projectId);
+            var pages = _publisher.ListTopics(projectName).AsPages();
```
SurferJeffAtGoogle: Did we discuss using the word lazily here?

pubsub/api/PubsubTest.cs
```diff
@@ -201,53 +203,53 @@ public void DeleteTopic(string topicId, PublisherClient publisher)
             string topicName = PublisherClient.FormatTopicName(_projectId, topicId);
             string subscriptionName =
                 SubscriberClient.FormatSubscriptionName(_projectId, subscriptionId);
-            var delayMs = _retryDelayMs;
+            var delayMs = _retryDelayMs;
```
jskeet: Haven't seen this sample before, but you realize retry is built into the library and can be adjusted with settings?

pubsub/api/PubsubTest.cs
jskeet: I don't think so in general. (We did for Datastore and Bigquery, I think.)

This would only change to AsResponses, in the current PR.

SurferJeffAtGoogle:
I figured out retry was built into the library a couple weeks ago.  Haven't
had a chance to update the samples yet...

PR #87, opened 2016-11-08
Logging samples with Google.Logging.V2

jsimonweb:


logging/api/QuickStart/QuickStart.cs
```diff
@@ -0,0 +1,76 @@
...
+using Google.Api;
+
+namespace GoogleCloudSamples
+{
+    public class QuickStart
+    {
+        static void Main(string[] args)
+        {
+            // Your Google Cloud Platform project ID.
+            string projectId = "YOUR-PROJECT-ID";
```
SurferJeffAtGoogle: Let's add a check detecting YOUR-PROJECT-ID and printing a message that says you need to edit the code.

logging/api/LoggingTest/LoggingTest.cs
```diff
@@ -0,0 +1,239 @@
...
+                output = consoleApp.StandardOutput.ReadToEnd();
+                consoleApp.WaitForExit();
+                return output;
+            }
+
+            [Fact]
+            public void TestQuickStartConsoleApp()
+            {
+                string output;
+                string filePath = "..\\..\\..\\QuickStart\\bin\\Debug\\QuickStart.exe";
```
SurferJeffAtGoogle: Use an @"string" and then you don't need to escape the slashes.

logging/api/Program.cs
```diff
@@ -0,0 +1,237 @@
...
+            var results = client.ListLogEntries(projectIds, logName, "timestamp desc");
+            foreach (var row in results)
+            {
+                if (row != null && !String.IsNullOrEmpty(row.TextPayload.Trim()))
+                {
+                    _out.WriteLine($"{row.TextPayload.Trim()}");
+                }
+                else
+                {
+                    results.GetEnumerator().Dispose();
```
SurferJeffAtGoogle: Instead, let's change line 97 to 

using (var results = client.ListLogEntries(projectIds, logName, "timestamp desc"))
{

logging/api/LoggingTest/LoggingTest.cs
```diff
@@ -0,0 +1,239 @@
...
+
+        protected static void AssertSucceeded(ConsoleOutput output)
+        {
+            Assert.True(0 == output.ExitCode,
+                $"Exit code: {output.ExitCode}\n{output.Stdout}");
+        }
+
+        public class LoggingTest : BaseTest, IDisposable
+        {
+            public void Dispose() { }
```
SurferJeffAtGoogle: If Dispose() is empty, then we can just drop it and stop implementing IDisposable.

logging/api/QuickStart/QuickStart.cs
jsimonweb: Done.

logging/api/LoggingTest/LoggingTest.cs
jsimonweb: Done.

logging/api/Program.cs
SurferJeffAtGoogle: I downloaded the code and tried running it, and it blocks forever even with the work-around.  :-(

logging/api/LoggingTest/LoggingTest.cs
```diff
@@ -0,0 +1,237 @@
...
+                string message = "Example log entry.";
+                // Try creating a log entry.
+                var created = Run("create-log-entry", logId, message);
+                AssertSucceeded(created);
+                // Pause for 5 seconds before trying to get newly added log entry.
+                Thread.Sleep(5000);
+                // Retrieve the log entry just added, using the logId as a filter.
+                var results = Run("list-log-entries", logId);
+                // Confirm returned log entry contains expected value.
+                Assert.Contains(message, results.Stdout);
```
SurferJeffAtGoogle: If the assertion fails, then the Run("delete-log") statement below will never run, and the test will leak a log.

This pattern repeats in the logs below.  Adding try .. finally .. to every test would be tedious.  Instead, you could add a list of garbage as a member variable for the class, and then implement Dispose() to delete the garbage.

jsimonweb: Done.

logging/api/LoggingTest/LoggingTest.cs
```diff
@@ -64,13 +68,50 @@ protected static void AssertSucceeded(ConsoleOutput output)
...
+            {
+                try
+                {
+                    // Delete all logs created from running the tests.
+                    foreach (string log in _logsToDelete)
+                    {
+                        Run("delete-log", log);
+                    }
+                }
+                catch (RpcException ex)
```
SurferJeffAtGoogle:
```
catch (RpcException ex) when (x.Status.StatusCode == StatusCode.NotFound) {}
```

logging/api/LoggingTest/LoggingTest.cs
```diff
@@ -64,13 +68,50 @@ protected static void AssertSucceeded(ConsoleOutput output)
...
+                }
+                try
+                {
+                    // Delete all the log sinks created from running the tests.
+                    foreach (string sink in _sinksToDelete)
+                    {
+                        Run("delete-sink", sink);
+                    }
+                }
+                catch (RpcException ex)
```
SurferJeffAtGoogle: catch(...) when (...) {}

logging/api/LoggingTest/LoggingTest.cs
jsimonweb: Done.

logging/api/LoggingTest/LoggingTest.cs
jsimonweb: Done

PR #131, opened 2017-01-04
Implement ASP.NET session storage with datastore.

SurferJeffAtGoogle:


sessionstate/README.md
```diff
@@ -0,0 +1,30 @@
+# Cloud Storage Sample
```
jsimonweb: Google Cloud Datastore Sample

PR #143, opened 2017-01-21
Update sample to include PubSub IAM

jsimonweb:


pubsub/api/PubsubTest.cs
```diff
@@ -46,25 +47,20 @@ public class PubsubTest
...
-                MaxDelay = TimeSpan.FromSeconds(3)
-            };
-            return new CallSettings()
-            {
-                Timing = CallTiming.FromRetry(new RetrySettings()
-                {
-                    RetryBackoff = backoff,
-                    TimeoutBackoff = backoff,
-                    RetryFilter = (RpcException e) => (
+            TimeSpan delay = TimeSpan.FromMilliseconds(500);
```
SurferJeffAtGoogle: Add a comment here.

jsimonweb: Added comment.

PR #156, opened 2017-02-07
Scripts to create an Always On Availability Group in Google Cloud

sqlthinker:
The scripts will automatically create a SQL Server Availability Group

### DIFF
--- a/compute/sqlserver/powershell/create-availability-group.ps1
+++ b/compute/sqlserver/powershell/create-availability-group.ps1
@@ -1,0 +1,409 @@
+﻿#Requires -Version 5
+# Copyright(c) 2016 Google Inc.
+#
+# Licensed under the Apache License, Version 2.0 (the "License"); you may not
+# use this file except in compliance with the License. You may obtain a copy of
+# the License at
+#
+# http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS, WITHOUT
+# WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied. See the
+# License for the specific language governing permissions and limitations under
+# the License.
+
+##############################################################################
+#.SYNOPSIS
+# Set up a two node SQL Server Availability Group
+#
+#.DESCRIPTION
+# This script will set up a two node Availability Group in SQL Server.
+# It would normally be called by the script
+# create-sql-instance-availability-group.ps1, but it can also be called by 
+# itself. Before running this script you must install 'Cloud SDK for Windows' 
+# which includes the Powershell CmdLet for Google Cloud.
+# See https://googlecloudplatform.github.io/google-cloud-powershell
+# The script expects the file parameters-config.ps1 to exist in the same 
+# directory.
+#
+#.NOTES
+# AUTHOR: Anibal Santiago - @SQLThinker
+#
+#.EXAMPLE
+# create-availability-group.ps1
+# Before running the script make you need to modify the file 
+# parameters-config.ps1 with the parameters specific to your configuration
+##############################################################################
+
+$ErrorActionPreference = 'Stop'
+Set-Location $PSScriptRoot
+
+################################################################################
+# Read the parameters for this script. They are found in the file 
+# parameters-config.ps1. We assume it is in the same folder as this script
+################################################################################
+if (!($cred)) {
+  . ".\parameters-config.ps1"
+}
+
+# Only continue if SQL Server is installed in both nodes
+$node1_sql_server = Get-Service -ComputerName $node1 | 
+  Where Name -Like 'MSSQLSERVER*'
+$node2_sql_server = Get-Service -ComputerName $node2 | 
+  Where Name -Like 'MSSQLSERVER*'
+
+If ( !($node1_sql_server) -Or !($node2_sql_server) ) {
+  Throw 'Need to install SQL Server in both nodes before running this script.'
+}
+
+# Only continue if script is ran from a computer that is member of the domain
+if ( !( Get-ADDomain | Where Forest -eq $domain ) ) {
+  Throw "Need to run script from a computer member of the '$domain' domain."
+}
+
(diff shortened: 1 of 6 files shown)


SurferJeffAtGoogle:
Thanks for the PR, Anibal.  It's going to take me awhile to review.  I'll probably reply with my first comments early next week.

compute/sqlserver/powershell/create-availability-group.ps1
```diff
@@ -0,0 +1,409 @@
...
+$node1_sql_server = Get-Service -ComputerName $node1 | 
+  Where Name -Like 'MSSQLSERVER*'
+$node2_sql_server = Get-Service -ComputerName $node2 | 
+  Where Name -Like 'MSSQLSERVER*'
+
+If ( !($node1_sql_server) -Or !($node2_sql_server) ) {
+  Throw 'Need to install SQL Server in both nodes before running this script.'
+}
+
+# Only continue if script is ran from a computer that is member of the domain
```
SurferJeffAtGoogle: Why?  Is it a security concern?  (Answer in comment.)

compute/sqlserver/powershell/create-availability-group.ps1
```diff
@@ -0,0 +1,409 @@
...
+  Get-SmbShare | Where-Object -Property Name -eq $share_name | 
+    Remove-SmbShare -Force
+  New-SMBShare –Name $share_name –Path $sql_backup –FullAccess "$($node1)`$","NT SERVICE\MSSQLSERVER" | Out-Null
+} -ArgumentList $sql_backup, $share_name, $domain, $svc_acct, $node1
+
+
+# Enable CredSSP in local computer
+Write-Host "$(Get-Date) Enable CredSSP Client in local machine"
+Enable-WSManCredSSP Client -DelegateComputer * -Force | Out-Null
+Start-Sleep 15
```
SurferJeffAtGoogle: Why sleep?  What happens if the operation doen't complete in 15 minutes?

compute/sqlserver/powershell/create-sql-instance-availability-group-prerequisites.ps1
```diff
@@ -0,0 +1,384 @@
...
+#.NOTES
+# AUTHOR: Anibal Santiago - @SQLThinker
+##############################################################################
+
+
+################################################################################
+# Specify the parameters
+################################################################################
+# Network configuration
+$network      = 'wsfcnet'      # Name of the Custom network
```
SurferJeffAtGoogle: It's not clear if users are expected to update these parameters with their own settings.

compute/sqlserver/powershell/create-sql-instance-availability-group-prerequisites.ps1
```diff
@@ -0,0 +1,384 @@
...
+
+Write-Host " "
+Write-Host "IMPORTANT: You are now ready to run the script  create-sql-instance-availability-group.ps1"
+Write-Host "           It is recommended to run it from an instance that is a member of the domain: $DomainName"
+Write-Host "           if you want to automatically setup the SQL Server Availability Group."
+
+# Remove firewall rule to allow current IP
+#Get-GceFirewall -FirewallName "allow-winrm-current-ip" | Remove-GceFirewall
+
+##### Cleanup - Run with caution as everything will be deleted #####
```
SurferJeffAtGoogle: Could we put the cleanup code into its own function or script?

compute/sqlserver/powershell/create-sql-instance-availability-group-prerequisites.ps1
sqlthinker: We have a document that goes step by step how to create a HA Availability Group. The document references the script create-sql-instance-availability-group.ps1 as a way to automate the steps. But, it also says that before running the script, they need to do some pre-requisites. We realized that some customers are not familiar with GCP and will not know how to do those pre-requisites. That is why we created this "-prerequisites" script for them. It matches the same values we used in our documentation. We expected that it will be used by clients following the instructions document and that will use exactly the same values we used in the doc.

compute/sqlserver/powershell/create-sql-instance-availability-group-prerequisites.ps1
sqlthinker: Agree. I'll remove it from here and put it in a new script with the same name but with "-cleanup.ps1" at the end.

SurferJeffAtGoogle:
Like the bug fixes in f867696, please make changes in this PR as additional commits.

compute/sqlserver/powershell/create-availability-group.ps1
SurferJeffAtGoogle: Ahh.  Ok.  Can you copy that into a comment?

compute/sqlserver/powershell/create-availability-group.ps1
SurferJeffAtGoogle: Ok.  Sounds like a reasonable solution.  Please say something about the random error in a comment.

compute/sqlserver/powershell/create-sql-instance-availability-group-prerequisites.ps1
SurferJeffAtGoogle: If the doc is live, please add a link to it here.

sqlthinker:
Added more comments based on feedback. Also added the -s to the calls to Start-Sleep for readability. Fixed some bugs where we checked the wrong network name when validating if the network already exists.

sqlthinker:
To test you can run the script "create-sql-instance-availability-group.Tests.ps1". It uses the Pester module to validate that the Availability Group was successfully created. But, be aware that the script needs to be run from a computer member of the domain. It is the only way to create the cluster because of the "double-hop" issue when creating the cluster using WinRM and CredSSP. You can modify the script "parameters-config.ps1" and specify the password for the account to add the nodes to the domain. The default value is "$null" which means the script will prompt the user for a password, which may be an issue if you are trying to automate it.

PR #169, opened 2017-02-14
Add KMS IAM code samples

jsimonweb:


kms/api/Program.cs
```diff
@@ -427,6 +504,247 @@ public static object RestoreCryptoKeyVersion(string projectId, string location,
...
+                            Console.WriteLine($"  Member: {member.ToString()}");
+                        });
+                    });
+                }
+                else
+                {
+                    Console.WriteLine($"Empty IAM policy found for CryptoKey: {parent}");
+                }
+            }
+            catch (Google.GoogleApiException e)
```
SurferJeffAtGoogle: The try .. catch code falls inside the doc tags and will therefore appear in the documentation page, which clouds the sample a bit, and it doesn't need to.  We could put the try ... catch in a function higher in the stack.  Or, just remove it entirely (my preference.)  In that case, when it fails, the user still sees a stack trace and message and the return value will be nonzero.  Test code that expects a failure needs to change a little bit.  But test code that expects success would remain unchanged.

kms/api/Program.cs
```diff
@@ -427,6 +504,247 @@ public static object RestoreCryptoKeyVersion(string projectId, string location,
...
+                    .SetIamPolicyRequest(cloudKms, setIamPolicyRequest, parent);
+                var setIamPolicyResult = request.Execute();
+                var updateResult = cloudKms.Projects.Locations.KeyRings.CryptoKeys.GetIamPolicy(parent).Execute();
+                updateResult.Bindings.ToList().ForEach(response =>
+                {
+                    Console.WriteLine($"Role: {response.Role}");
+
+                    response.Members.ToList().ForEach(memberFound =>
+                    {
+                        Console.WriteLine($"  Member: {memberFound.ToString()}");
```
SurferJeffAtGoogle: I don't think it's necessary to call ToString() here.

PR #225, opened 2017-03-30
Updated to new API

riyadparvez:


riyadparvez:
I signed it!

SurferJeffAtGoogle:
I can't find a record of your signature in our signing tool.

pubsub/api/QuickStart/Program.cs
```diff
@@ -14,7 +14,7 @@
 
 // [START pubsub_quickstart]
 
-using Google.Cloud.PubSub.V1;
+using Google.Apis.Pubsub.v1;
```
SurferJeffAtGoogle: This is actually an older API.  It takes more lines of code to use it.

PR #254, opened 2017-04-28
Catch entity-sentiment availability.

jsimonweb:


language/api/Analyze/Program.cs
```diff
@@ -210,27 +210,43 @@ private static void WriteEntitySentiment(IEnumerable<Entity> entities)
...
-                ExtractDocumentSentiment = true,
-                ExtractEntities = true,
-                ExtractEntitySentiment = true
-            });
-            Console.WriteLine($"Language: {response.Language}");
-            WriteSentiment(response.DocumentSentiment, response.Sentences);
-            WriteSentences(response.Sentences, response.Tokens);
-            WriteEntities(response.Entities);
-            WriteEntitySentiment(response.Entities);
+                AnalyzeEverything(text, false);
```
jskeet: If there's any problem *other* than sentiment analysis not being available, this is going to recurse forever. Basically we're assuming one particular cause, where there could be any number of real causes.

language/api/Analyze/Program.cs
```diff
@@ -210,27 +210,43 @@ private static void WriteEntitySentiment(IEnumerable<Entity> entities)
...
         {
             var client = LanguageServiceClient.Create();
-            var response = client.AnnotateText(new Document()
+            try
             {
-                Content = text,
-                Type = Document.Types.Type.PlainText
-            },
-            new Features()
+                var response = client.AnnotateText(new Document()
```
jskeet: Document.FromPlainText(text) is your friend - that's why I provided it.

PR #262, opened 2017-05-04
Create openapi-appengine.yaml

p-buse:


p-buse:
It has a different _host_ field. I will add a comment to the exisiting openapi.yaml instead.

PR #267, opened 2017-05-23
Add a sample for adding git commit id to log entries

Deren-Liao:
This sample will be linked by the user documentation for Logs Viewer feature inside our Google Cloud Tools for Visual Studio extension.
The main purpose is to help user to know how to add the git revision id if they decide to use lower level logging API Google.Cloud.Logging.V2 instead of using Log4Net.


logging/add-git-id/App_Start/RouteConfig.cs
```diff
@@ -0,0 +1,39 @@
+﻿/*
```
SurferJeffAtGoogle: Move the code into
/logging/api/WithGitId/...

logging/add-git-id/Controllers/LoggingSampleController.cs
```diff
@@ -0,0 +1,51 @@
...
+
+        // GET: /LoggingSample/Log/1
+        [HttpGet]
+        public ActionResult Log(int? count)
+        {
+            if (count > 0 && count <= 100)
+            {
+                for (int i = 0; i < count; ++i)
+                {
+                    StackdriverLogWritter.WriteLog("Test. Replace with your own message.");
```
SurferJeffAtGoogle: Is there a type in the name LogWritter?  Should it be LogWriter?

Make the text something like "Hello Stackdriver."

logging/add-git-id/Controllers/LoggingSampleController.cs
```diff
@@ -0,0 +1,51 @@
...
+                }
+                return Content($"Inserted {count} logs");
+            }
+            else
+            {
+                throw new ArgumentOutOfRangeException("Pleae limit count in range of 1 ~ 100");
```
SurferJeffAtGoogle: Or maybe default to a small number like 3?

logging/add-git-id/Global.asax.cs
```diff
@@ -0,0 +1,18 @@
+﻿using System;
```
SurferJeffAtGoogle: Add copyright.

logging/add-git-id/Properties/AssemblyInfo.cs
```diff
@@ -0,0 +1,35 @@
+﻿using System.Reflection;
```
SurferJeffAtGoogle: Add copyright.

logging/add-git-id/StackdriverLogWritter.cs
```diff
@@ -0,0 +1,102 @@
...
+using Google.Protobuf;
+
+namespace GoogleCloudSamples
+{
+    public static class StackdriverLogWritter
+    {
+        private const string LogId = "YOUR-LOG-ID";
+        private const string ProjectId = "YOUR-PROJECT-ID";
+
+        [MethodImpl(methodImplOptions: MethodImplOptions.NoInlining)]
```
SurferJeffAtGoogle: Why?

logging/add-git-id/StackdriverLogWritter.cs
```diff
@@ -0,0 +1,102 @@
...
+                    labels.Add(SourceContext.GitRevisionIdLogLabel, gitId);
+                }
+            }
+            catch (Exception ex) when (
+                ex is SecurityException
+                || ex is InvalidProtocolBufferException
+                || ex is InvalidJsonException
+                || ex is UnauthorizedAccessException)
+            {
+                // This is best-effort only, exceptions from reading/parsing the source_context.json are ignored.
```
SurferJeffAtGoogle: If this fails, then this sample fails to show users what is was designed to show them, no?  In that case, exceptions should not be ignored.  Probably the simplest option is to not catch them at all.

logging/add-git-id/Controllers/LoggingSampleController.cs
Deren-Liao: This is protective code. The UI already set limit to 100.   Just in case somebody hack the input (which not likely), this protect the site creating huge number of log entries.

logging/add-git-id/StackdriverLogWritter.cs
Deren-Liao: I am NOT sure if this is inline , then would the caller information be messed up

logging/add-git-id/StackdriverLogWritter.cs
Deren-Liao: If user incorporate the sample into their production code, I don't want the logger blow up their products.   I added Debug.Writeline(..)  , hopefully it helps in some extent.

Deren-Liao:
PTAL.
I added a test, and README.md.  To let the test run, I upgraded the nuget packages versions so that all projects using same versions.

SurferJeffAtGoogle:
I got this error when I tried to build:
```
Restoring NuGet packages...
To prevent NuGet from restoring packages during build, open the Visual Studio Options dialog, click on the Package Manager node and uncheck 'Allow NuGet to download missing packages during build.'
NuGet package restore finished.
1>------ Build started: Project: WithGitId, Configuration: Debug Any CPU ------
1>Build started 5/30/2017 11:12:05 AM.
1>C:\Users\Jeffrey Rennie\gitrepos\dotnet-docs-samples\logging\api\WithGitId\WithGitId.csproj(5,3): error MSB4019: The imported project "C:\Users\Jeffrey Rennie\gitrepos\dotnet-docs-samples\logging\api\WithGitId\packages\Microsoft.CodeDom.Providers.DotNetCompilerPlatform.1.0.0\build\Microsoft.CodeDom.Providers.DotNetCompilerPlatform.props" was not found. Confirm that the path in the <Import> declaration is correct, and that the file exists on disk.
1>
1>Build FAILED.
1>
1>Time Elapsed 00:00:00.00
========== Build: 0 succeeded, 1 failed, 0 up-to-date, 0 skipped ==========
```

I searched for the error text and found:
https://stackoverflow.com/questions/32254439/nuget-packages-are-missing

SurferJeffAtGoogle:
Did you build this project with VS2017 or 2015?

Deren-Liao:
I built it with VS2015.
Just now, I tried to build with VS2015 and VS2017. I am getting this error
 --- from a fresh clone.   I mean it worked when I have the packages
locally.
Let me take a closer look.

Deren-Liao:
Yes, that's the problem.  I changed the solution file path, copied project
folder over.  The packages binaries under project folder were moved
together . So it built successfully locally on my machine.

I fixed the problem and tested with clean clone on both VS2015 and VS2017

logging/add-git-id/StackdriverLogWritter.cs
SurferJeffAtGoogle: 'tis noble to not blow up user products.  :-)

I tried downloading and running the sample.  I set and the project id and log name and pressed F5.  But my logging statements contain no git revision ids.  I found it hard to find the output of Debug.Writeline().  I think many users will not see it at all.

If you still think an exception is too dangerous, can we use a Debug.Assert() instead?  It won't execute at all in a production build, but will halt the program in a debug build.

Deren-Liao:
I will add that.
To be clear, git id won't be added in F5.  You have to publish it so that
the source_context.json is generated.

Deren-Liao:
I added Assert.
Some users may just copy/paste the code. I really don't want a log
exception breaks their product by any means.
If you want to try get the git id added, please use our extension to
publish it to a GCE ASP.NET image VM.

Thank you for all the feedback.  I appreciate!

logging/api/WithGitId/StackdriverLogWriter.cs
```diff
@@ -0,0 +1,105 @@
...
+                resource: resource, 
+                labels: entryLabels, 
+                entries: new[] { logEntry }, 
+                callSettings: null);
+            return logEntry.ToString();
+        }
+
+        private static void TryAddGitRevisionId(IDictionary<string, string> labels)
+        {
+            try
```
SurferJeffAtGoogle: This code can be a little tighter:
```cs
            try
            {
                var gitId = SourceContext.AppSourceContext?.Git?.RevisionId;
                Debug.Assert(!String.IsNullOrWhiteSpace(gitId), "No valid git revision id found.",
                    "Make sure you have source-context.json published with the application.");
                labels.Add(SourceContext.GitRevisionIdLogLabel, gitId);
            }
            catch (Exception ex) when (
                ex is SecurityException
                || ex is InvalidProtocolBufferException
                || ex is InvalidJsonException
                || ex is UnauthorizedAccessException)
            {
                // This is best-effort only, exceptions from reading/parsing the source_context.json are ignored.
                Debug.Fail("Exception at TryAddGitRevisionId.", ex.ToString());
            }
```

Deren-Liao:
Done.
Please take a look again.

SurferJeffAtGoogle:
Thank you Deren!

Deren-Liao:
Thanks!

PR #288, opened 2017-07-17
Add Spanner sample

jsimonweb:


spanner/api/Program.cs
```diff
@@ -0,0 +1,965 @@
...
+                 ) PRIMARY KEY (SingerId)",
+                 @"CREATE TABLE Albums (
+                     SingerId     INT64 NOT NULL,
+                     AlbumId      INT64 NOT NULL,
+                     AlbumTitle   STRING(MAX)
+                 ) PRIMARY KEY (SingerId, AlbumId),
+                 INTERLEAVE IN PARENT Singers ON DELETE CASCADE"};
+            // Make the request.
+            var response =
+                await databaseAdminClient.UpdateDatabaseDdlAsync(
```
benwulfe: whats wrong with using a spanner ddl command for this?
you know you can create a connection string that only has projectid + instanceid (no db name).

spanner/api/Program.cs
```diff
@@ -0,0 +1,965 @@
...
+            // Create client.
+            DatabaseAdminClient databaseAdminClient =
+                await DatabaseAdminClient.CreateAsync();
+            // Initialize request argument(s).
+            InstanceName parent = new InstanceName(projectId, instanceId);
+            string createStatement =
+                $"CREATE INDEX AlbumsByAlbumTitle ON Albums(AlbumTitle)";
+            // Make the request.
+            var createDatabaseRequest =
+                await databaseAdminClient.UpdateDatabaseDdlAsync(
```
benwulfe: same here, you can use connection.CreateDdlCommand(...)

spanner/api/Program.cs
```diff
@@ -0,0 +1,965 @@
...
+            DatabaseAdminClient databaseAdminClient =
+                await DatabaseAdminClient.CreateAsync();
+            // Initialize request argument(s).
+            InstanceName parent = new InstanceName(projectId, instanceId);
+            string createStatement =
+                $"CREATE INDEX AlbumsByAlbumTitle2 ON Albums(AlbumTitle) "
+                + "STORING (MarketingBudget)";
+            // Make the request.
+            var createDatabaseRequest =
+                await databaseAdminClient.UpdateDatabaseDdlAsync(
```
benwulfe: connection.CreateDdlCommand

spanner/api/Program.cs
```diff
@@ -0,0 +1,965 @@
...
+            // Create client.
+            DatabaseAdminClient databaseAdminClient =
+                await DatabaseAdminClient.CreateAsync();
+            // Initialize request argument(s).
+            InstanceName parent = new InstanceName(projectId, instanceId);
+            string alterStatement =
+                $"ALTER TABLE Albums ADD COLUMN MarketingBudget INT64";
+            // Make the request.
+            var createDatabaseRequest =
+                await databaseAdminClient.UpdateDatabaseDdlAsync(
```
benwulfe: connection.CreateDdlCommand

spanner/api/Program.cs
```diff
@@ -0,0 +1,965 @@
...
+            string endTitle = "Goo";
+            string connectionString =
+            $"Data Source=projects/{projectId}/instances/{instanceId}"
+            + $"/databases/{databaseId}";
+            // Create connection to spanner.
+            using (var connection = new SpannerConnection(connectionString))
+            {
+                var cmd = connection.CreateSelectCommand(
+                    "SELECT AlbumId, AlbumTitle, MarketingBudget FROM Albums@ "
+                    + "{FORCE_INDEX=AlbumsByAlbumTitle} "
```
benwulfe: we should get someone from spanner to comment  -- but i would assume it would do the correct thing and use that index if you used that column in orderby... yes?

spanner/api/Program.cs
```diff
@@ -0,0 +1,965 @@
...
+            // Create connection to spanner.
+            using (var connection = new SpannerConnection(connectionString))
+            {
+                var cmd = connection.CreateSelectCommand(
+                    "SELECT SingerId, AlbumId, AlbumTitle FROM Albums");
+                using (var reader = await cmd.ExecuteReaderAsync())
+                {
+                    while (await reader.ReadAsync())
+                    {
+                        Console.WriteLine($"SingerId : {reader["SingerId"]} "
```
benwulfe: we've put a lot of work into reader.GetFieldValue<T>("columnname").
where T is the type you want the column converted to.  This is a better approach than using the default type we choose for you (via the bracket operator) because if you ever do a schema change, your code has a good chance of still working.
eg.
 you store a column as FLOAT64, but only ever store ints.
 you get the value via:  reader.GetFieldValue<long>("widget_count");
 realizing your error, you switch the column type in spanner to "INT64" -- your code will work without modification.
 However, if you had done long count = Convert.ToInt64((double)reader["widget_count"]);  it will result in a cast exception because the default type we return will change to long.

benwulfe: reader.GetFieldValue{T}  that is

spanner/api/Program.cs
```diff
@@ -0,0 +1,965 @@
...
+        public static async Task QueryDataWithTransactionAsync(
+    string projectId, string instanceId, string databaseId)
+        {
+            // [START read_only_transaction]
+            string connectionString =
+            $"Data Source=projects/{projectId}/instances/{instanceId}"
+            + $"/databases/{databaseId}";
+            // Gets a transaction object that captures the database state
+            // at a specific point in time.
+            using (TransactionScope scope = new TransactionScope(
```
benwulfe: unfortunately, you will need to do another version of this for .net core -- where this type doesnt exist and you need to set .Transaction manually on each command.  (bummer, I know)

spanner/api/Program.cs
```diff
@@ -0,0 +1,965 @@
...
+                                $"SingerId : {reader["SingerId"]}"
+                                + $" AlbumId : {reader["AlbumId"]}"
+                                + $" AlbumTitle : {reader["AlbumTitle"]}");
+                        }
+                    }
+                }
+                // Commit the transaction to Spanner table.
+                scope.Complete();
+                // Yield Task thread back to the current context.
+                await Task.Yield();
```
benwulfe: can we somehow mark this Task.Yield for deletion with a TODO?  I'll make sure this turd is gone before GA.

spanner/api/Program.cs
```diff
@@ -0,0 +1,965 @@
...
+            string connectionString =
+            $"Data Source=projects/{projectId}/instances/{instanceId}"
+            + $"/databases/{databaseId}";
+            // Create connection to spanner.
+            using (var connection = new SpannerConnection(connectionString))
+            {
+                Console.WriteLine("Connected...");
+                var cmd = connection.CreateSelectCommand(
+                    "SELECT SingerId, AlbumId, AlbumTitle FROM Albums");
+                using (var reader = await cmd.ExecuteReaderAsync())
```
benwulfe: this is the same as the query one right?
maybe we can change it up a little bit and show how you can do parallel queries by running executereaderasync from several commands on the same connection... ?
you can do multiple reads and writes concurrently on the same connection -- but concurrent writes will consume extra sessions.

spanner/api/Program.cs
```diff
@@ -0,0 +1,965 @@
...
+                var cmdLookup =
+                    connection.CreateSelectCommand("SELECT * FROM Albums");
+                using (var reader = await cmdLookup.ExecuteReaderAsync())
+                {
+                    while (await reader.ReadAsync())
+                    {
+                        if (reader.GetInt64(0) == 1 && reader.GetInt64(1) == 1)
+                        {
+                            cmd.Parameters["SingerId"].Value =
+                                reader.GetInt64(0);
```
benwulfe: can we use the field name overloads instead of ordinal?  the ordinal is a bit brittle

spanner/api/Program.cs
```diff
@@ -0,0 +1,965 @@
...
+            // This sample transfers 200,000 from the MarketingBudget 
+            // field of the second Album to the first Album. Make sure to run
+            // the addColumn and writeDataToNewColumn samples first,
+            // in that order.
+
+            string connectionString =
+            $"Data Source=projects/{projectId}/instances/{instanceId}"
+            + $"/databases/{databaseId}";
+
+            using (TransactionScope scope = new TransactionScope(
```
benwulfe: + .net core version...

spanner/api/Program.cs
```diff
@@ -0,0 +1,965 @@
...
+                    cmd.Parameters["SingerId"].Value = 1;
+                    cmd.Parameters["AlbumId"].Value = 1;
+                    cmd.Parameters["MarketingBudget"].Value = firstBudget;
+                    await cmd.ExecuteNonQueryAsync();
+
+                    // Commit the transaction to Spanner table.
+                    scope.Complete();
+                    // Yield Task thread back to the current context.
+                    await Task.Yield();
+                    Console.WriteLine("Transaction complete.");
```
benwulfe: you should be using topaz as well here to handle retries:

https://www.codeproject.com/Articles/769634/Transient-Fault-Handling-Application-Block-Simplif

you can use the extension method exception.IsTransientSpannerFault() for your implementation of a detection strategy.
let me know if you have more questions about this and I can go over it in more detail.

spanner/api/Program.cs
```diff
@@ -0,0 +1,965 @@
...
+                    }
+                    // Read the first album's budget.
+                    cmdLookup = connection.CreateSelectCommand(
+                    "SELECT * FROM Albums WHERE SingerId = 1 and AlbumId = 1");
+                    using (var reader = await cmdLookup.ExecuteReaderAsync())
+                    {
+                        while (await reader.ReadAsync())
+                        {
+                            firstBudget =
+                                decimal.Parse(reader["MarketingBudget"]
```
benwulfe: reader.GetFieldValue<decimal>("MarketingBudget")

benwulfe: damn github always removes my template params:

reader.GetFieldValue{decimal}("MarketingBudget")

spanner/api/Program.cs
```diff
@@ -0,0 +1,965 @@
...
+                cmd.Parameters["SingerId"].Value = 2;
+                cmd.Parameters["AlbumId"].Value = 2;
+                cmd.Parameters["AlbumTitle"].Value = "Ambient Sistrum Sirens";
+                await cmd.ExecuteNonQueryAsync();
+                Console.WriteLine("Inserted data.");
+                // Insert fifth row.
+                cmd.Parameters["SingerId"].Value = 2;
+                cmd.Parameters["AlbumId"].Value = 3;
+                cmd.Parameters["AlbumTitle"].Value = "Congo Carols";
+                await cmd.ExecuteNonQueryAsync();
```
benwulfe: do you have a bulk insert example?
also the docs at some point should say that while you *can* do a Task.WhenAll(...multiple invocations of executequery...) it will consume extra sessions.

spanner/api/Program.cs
```diff
@@ -0,0 +1,965 @@
...
+            }
+            // [END insert_data]
+        }
+
+        public static async Task<Database> CreateDatabaseAsync(
+            string projectId, string instanceId, string databaseId)
+        {
+            // [START create_database]
+            // Create client
+            DatabaseAdminClient databaseAdminClient =
```
benwulfe: this should also be connection.CreateDdl...

spanner/api/Program.cs
```diff
@@ -0,0 +1,965 @@
...
+            // [END create_database]
+        }
+
+        public static async Task<object> UpdateDatabaseAsync(string projectId,
+            string instanceId, string databaseId, string tableName,
+            string ddlStatement)
+        {
+            // [START update_database]
+            // Create client
+            DatabaseAdminClient databaseAdminClient =
```
benwulfe: CreateDdl

spanner/api/QuickStart/Program.cs
```diff
@@ -0,0 +1,56 @@
...
+            using (var connection = new SpannerConnection(connectionString))
+            {
+                // Execute a simple SQL statement.
+                var cmd = connection.CreateSelectCommand(
+                    @"SELECT ""Hello World"" as test");
+                using (var reader = cmd.ExecuteReader())
+                {
+                    while (await reader.ReadAsync())
+                    {
+                        Console.WriteLine(reader.GetValue(0));
```
benwulfe: GetFieldValue<string>("test")

benwulfe: GetFieldValue{string}("test")

benwulfe:
Looking good so far!

A few missing samples on:
  DataAdapter
  Topaz
  .net core transactions

spanner/api/Program.cs
vkedia: I don't think spanner would allow you to change FLOAT64 to INT64 or vice versa. So in practice this might not matter. I think the only type transformation we allow is bytes to string but I am not sure.

spanner/api/Program.cs
```diff
@@ -0,0 +1,965 @@
...
+        }
+
+
+        public static async Task QueryDataWithIndexAsync(
+            string projectId, string instanceId, string databaseId)
+        {
+            // [START query_data_with_index]
+            // [START read_data_with_index]
+            string startTitle = "Aardvark";
+            string endTitle = "Goo";
```
vkedia: can you take these two as args with these being the default values.

spanner/api/Program.cs
vkedia: Spanner might not always chose the index so general recommendation is to specify force_index. Also this is the exact query we use in other languages so I would stick to this.

spanner/api/Program.cs
vkedia: I am guessing this would use TimestampBound.Strong?

spanner/api/Program.cs
```diff
@@ -0,0 +1,965 @@
...
+                        while (await reader.ReadAsync())
+                        {
+                            Console.WriteLine(
+                                $"SingerId : {reader["SingerId"]}"
+                                + $" AlbumId : {reader["AlbumId"]}"
+                                + $" AlbumTitle : {reader["AlbumTitle"]}");
+                        }
+                    }
+                }
+                // Commit the transaction to Spanner table.
```
vkedia: I don't think the below complete() actually does anything on the backend since read only transactions dont need to be committed. So you can just omit this comment. It is kind of misleading.

spanner/api/Program.cs
```diff
@@ -0,0 +1,965 @@
...
+                        {"AlbumId", SpannerDbType.Int64},
+                        {"MarketingBudget", SpannerDbType.Int64},
+                    });
+                var cmdLookup =
+                    connection.CreateSelectCommand("SELECT * FROM Albums");
+                using (var reader = await cmdLookup.ExecuteReaderAsync())
+                {
+                    while (await reader.ReadAsync())
+                    {
+                        if (reader.GetInt64(0) == 1 && reader.GetInt64(1) == 1)
```
vkedia: can you use column names here?

vkedia: Actually why do you need to lookup this row? You can assume it is there. It will keep the sample simpler.

spanner/api/Program.cs
vkedia: +1

spanner/api/Program.cs
```diff
@@ -0,0 +1,965 @@
...
+            $"Data Source=projects/{projectId}/instances/{instanceId}"
+            + $"/databases/{databaseId}";
+            // Gets a transaction object that captures the database state
+            // at a specific point in time.
+            using (TransactionScope scope = new TransactionScope(
+                TransactionScopeAsyncFlowOption.Enabled))
+            {
+                // Create connection to spanner.
+                using (var connection = new SpannerConnection(
+                    connectionString))
```
vkedia: @benwulfe I think this needs to be opened to specify a read only mode.

spanner/api/Program.cs
benwulfe: we had a lot of issues in firebase because users did not specify in code what type they were expecting and had to determine the magic underlying type the client library decided to return.
This isn't about what the server does.  This is about having the user use a method that declares what type he is expecting versus just casting it which is not as robust.  It also requires the value to be boxed, which is a performance penalty.

Also SqlDataReader provides this very method, so its a good idea from a competitive point of view.

So for example.

Do NOT do this:    double val = (double)reader["SomeColumn"];  <== boxes value, will easily break on schema changes.

DO this:  double val = reader.GetFieldValue{double}("SpannerColumn");  <== no boxing, more robust conversion in the client library.

we will do conversions for several types.  For example, complex types like structs and lists, you can pass in a concrete custom type derived from generics IList{T} and IDictionary{T}.
We'll also do date conversions to an underlying string format.

Most this work is actually done by the .net frameworks Convert.ToX().  We just enumerate thru all the various permutations so you need to think less about the types you are getting.

spanner/api/Program.cs
```diff
@@ -0,0 +1,965 @@
...
+            // [START create_database]
+            // Create client.
+            DatabaseAdminClient databaseAdminClient =
+                await DatabaseAdminClient.CreateAsync();
+            // Initialize request argument(s).
+            InstanceName parent = new InstanceName(projectId, instanceId);
+            string createStatement = $"CREATE DATABASE `{databaseId}`";
+            // Make the request.
+            var createDatabaseRequest =
+                await databaseAdminClient.CreateDatabaseAsync(parent,
```
benwulfe: this is using the vkit api which is not as idiomatic as the ado.net wrappers.  I cannot see why the vkit api doesnt let you simply pass the db name other than its a straight wrapper generated on the proto.  We could enhance it.

However, the ado.net version lets you pass any ddl statement -- it isn't specific to createdatabase, etc.

I would prefer all the samples to use the ado.net version as it will complicate matters if developers start taking direct dependencies on our vkit layer, which we have not vetted as much.

spanner/api/Program.cs
benwulfe: we should try to support this.  i'm not sure I've tested multiple ddl statements however.

spanner/api/Program.cs
benwulfe: yes

spanner/api/Program.cs
benwulfe: yes you can specify this via connection.OpenWithTimeBoundSettings.  However, this name is absolutely horrible and I want to change it in my next PR to OpenAsReadOnly() with an optional arg of timestampbound which defaults to strong.
Just a name change...
but yes -- you need to do an explicit open here to switch to readonly mode as readwrite is the default.

spanner/api/Program.cs
benwulfe: +1

jsimonweb:
Feedback comments addressed. Please review the latest changes.

spanner/api/Program.cs
```diff
@@ -0,0 +1,989 @@
...
+            // Initialize request argument(s).
+            string connectionString =
+                $"Data Source=projects/{projectId}/instances/"
+                + $"{instanceId}/databases/{databaseId}";
+            string createStatement =
+                $"ALTER TABLE Albums ADD COLUMN MarketingBudget INT64";
+            // Make the request.
+            using (var connection = new SpannerConnection(connectionString))
+            {
+                var createCmd = connection.CreateDdlCommand(createStatement);
```
benwulfe: shouldn't this local be called "updateCmd" instead of "create" since its an ALTER TABLE?

spanner/api/Program.cs
```diff
@@ -0,0 +1,989 @@
...
+            $"Data Source=projects/{projectId}/instances/{instanceId}"
+            + $"/databases/{databaseId}";
+            // Create connection to Cloud Spanner.
+            using (var connection = new SpannerConnection(connectionString))
+            {
+                var cmd = connection.CreateSelectCommand(
+                    "SELECT AlbumId, AlbumTitle, MarketingBudget FROM Albums@ "
+                    + "{FORCE_INDEX=AlbumsByAlbumTitle2} "
+                    + $"WHERE AlbumTitle >= '{startTitle}' "
+                    + $"AND AlbumTitle < '{endTitle}'");
```
benwulfe: do you also have a sample somewhere that shows parameters used in the select statement as opposed to inline values?

spanner/api/Program.cs
```diff
@@ -0,0 +1,989 @@
...
+                        {
+                            Console.WriteLine(
+                                $"SingerId : {reader["SingerId"]}"
+                                + $" AlbumId : {reader["AlbumId"]}"
+                                + $" AlbumTitle : {reader["AlbumTitle"]}");
+                        }
+                    }
+                }
+                //scope.Complete();
+                return;
```
benwulfe: transient errors (like the service being unavailable) can also affect reads and topaz should be used here for retries as well.

spanner/api/Program.cs
```diff
@@ -0,0 +1,989 @@
...
+                    "SELECT * FROM Albums WHERE SingerId = 2 AND AlbumId = 2");
+                    // Excecute the select query.
+                    using (var reader = await cmdLookup.ExecuteReaderAsync())
+                    {
+                        while (await reader.ReadAsync())
+                        {
+                            // Read the second album's budget.
+                            secondBudget =
+                                decimal.Parse(reader["MarketingBudget"]
+                                    .ToString());
```
benwulfe: you are missing culture settings -- should use invariant culture.  This is another reason to use GetFieldValue{decimal}("MarketingBudget").  You'll do that in your next PR?

spanner/api/Program.cs
```diff
@@ -0,0 +1,989 @@
...
+                }
+            }
+            return;
+            // [END query_data_with_new_column]
+        }
+
+        internal class CustomTransientErrorDetectionStrategy
+            : ITransientErrorDetectionStrategy
+        {
+            public bool IsTransient(Exception ex)
```
benwulfe: you can shorten this by using this expression body.

public bool IsTransient(Exception ex) => ex.IsTransientSpannerFault();

spanner/api/Program.cs
```diff
@@ -0,0 +1,989 @@
...
+            return 0;
+        }
+
+        public static object ReadWriteWithTransaction(string projectId,
+            string instanceId, string databaseId)
+        {
+            try
+            {
+                int retryCount = 3;
+                var retryStrategy = new Incremental(retryCount, TimeSpan.FromSeconds(2), TimeSpan.FromSeconds(2));
```
benwulfe: we normally use ExponentialBackoff for GCP services.  We should use that instead of Incremental.

spanner/api/Program.cs
```diff
@@ -0,0 +1,989 @@
...
+                var retryPolicy = new RetryPolicy<CustomTransientErrorDetectionStrategy>(retryStrategy);
+                var response = retryPolicy.ExecuteAction(async () => await ReadWriteWithTransactionAsync(projectId, instanceId, databaseId));
```
benwulfe: i wonder if the sample would be smaller if we inlined topaz?

spanner/api/Program.cs
```diff
@@ -0,0 +1,989 @@
...
+            Console.WriteLine($"Operation status: {response.Status}");
+            return 0;
+        }
+
+        public static object ReadWriteWithTransaction(string projectId,
+            string instanceId, string databaseId)
+        {
+            try
+            {
+                int retryCount = 3;
```
benwulfe: does retryCount need to be a local?  I dont see it used anywhere else.

spanner/api/Program.cs
```diff
@@ -0,0 +1,989 @@
...
+        }
+
+        public static object QueryNewColumn(string projectId,
+string instanceId, string databaseId)
+        {
+            var response = Task.Run(async () =>
+            await QueryNewColumnAsync(
+                projectId, instanceId, databaseId));
+            Console.WriteLine("Waiting for operation to complete...");
+            Task.WaitAll(response);
```
benwulfe: can't this just be:

QueryNewColumnAsync(projectId, instanceId, databaseId).Wait();

instead of task.run?

spanner/api/Program.cs
```diff
@@ -181,122 +189,111 @@ class ReadWriteWithTransactionOptions
...
-            // Create client.
-            DatabaseAdminClient databaseAdminClient =
-                await DatabaseAdminClient.CreateAsync();
             // Initialize request argument(s).
-            InstanceName parent = new InstanceName(projectId, instanceId);
-            string alterStatement =
+            string connectionString =
+                $"Data Source=projects/{projectId}/instances/"
+                + $"{instanceId}/databases/{databaseId}";
+            string createStatement =
```
benwulfe: "alter"Statement

spanner/api/Program.cs
```diff
@@ -181,122 +189,111 @@ class ReadWriteWithTransactionOptions
...
-            var updateDatabaseResponse =
-                await createDatabaseRequest.PollUntilCompletedAsync();
+            using (var connection = new SpannerConnection(connectionString))
+            {
+                var createCmd = connection.CreateDdlCommand(createStatement);
+                await createCmd.ExecuteNonQueryAsync();
+            }
             Console.WriteLine("Added the MarketingBudget column.");
-            return updateDatabaseResponse.Result;
+            return;
```
benwulfe: you dont need this "return" line.

spanner/api/Program.cs
```diff
@@ -403,9 +399,13 @@ AlbumTitle   STRING(MAX)
...
-                // Create connection to spanner.
-                using (var connection = new SpannerConnection(
-                    connectionString))
+                // Create connection to Cloud Spanner.
+                var connection = new SpannerConnection(
+                    connectionString);
+                // Open connection as read only within the scope 
+                // of the current transaction.
+                connection.OpenAsReadOnly();
+                using (connection)
```
benwulfe: you should do the using when you declare it -- as:

using (var connection = new SpannerConnection(connectionString)) {

    await connection.OpenAsReadOnlyAsync().ConfigureAwait(false);

benwulfe: (also use the async version of Open)

spanner/api/Program.cs
```diff
@@ -793,10 +746,40 @@ AlbumTitle   STRING(MAX)
...
+                Console.WriteLine("Or you can exclude both to run the "
+                    + "query with default values.");
+                return 0;
+            }
+            else if (!string.IsNullOrEmpty(startTitle)
+                && !string.IsNullOrEmpty(endTitle))
+            {
+                // Both startTitle and endTitle provided, include them both in
+                // the method call to QueryDataWithIndexAsync.
+                response = Task.Run(async () => await QueryDataWithIndexAsync(
```
benwulfe: this can just be  response = QueryDataWithIndexAsync(....);

shouldnt need task.run

spanner/api/Program.cs
```diff
@@ -840,11 +823,41 @@ AlbumTitle   STRING(MAX)
...
+                Console.WriteLine("Or you can exclude both to run the "
+                    + "query with default values.");
+                return 0;
+            }
+            else if (!string.IsNullOrEmpty(startTitle)
+                && !string.IsNullOrEmpty(endTitle))
+            {
+                // Both startTitle and endTitle provided, include them both in
+                // the method call to QueryDataWithIndexAsync.
+                response = Task.Run(async () =>
```
benwulfe: ditto here on task.run and its a few other places too

spanner/api/Program.cs
```diff
@@ -0,0 +1,989 @@
...
+            Console.WriteLine($"Operation status: {response.Status}");
+            Console.WriteLine($"Created sample database {databaseId} on "
+                + $"instance {instanceId}");
+            return 0;
+        }
+
+        public static object InsertSampleData(string projectId,
+string instanceId, string databaseId)
+        {
+            var response = Task.Run(async () => await InsertSampleDataAsync(
```
evildour: You don't need `async/await` in cases like this when when the only task to await is a return at the end.

spanner/api/Program.cs
```diff
@@ -0,0 +1,989 @@
...
+                && string.IsNullOrEmpty(endTitle))
+            {
+                // Only startTitle provided, show message that user must 
+                // provide both startTitle and endTitle or exclude them 
+                // from their command.
+                Console.WriteLine("Please provide values for both the start "
+                    + "of the title index and the end of the title index.");
+                Console.WriteLine("Or you can exclude both to run the "
+                    + "query with default values.");
+                return 0;
```
evildour: Should this return an error code?

spanner/api/Program.cs
```diff
@@ -0,0 +1,989 @@
...
+            Console.WriteLine("Waiting for operation to complete...");
+            Task.WaitAll(response);
+            Console.WriteLine($"Response status: {response.Status}");
+            return 0;
+        }
+
+        public static object QueryDataWithStoringIndex(string projectId,
+            string instanceId, string databaseId, string startTitle, string endTitle)
+        {
+            var response = new Task(() => { });
```
evildour: `= Task.CompletedTask` perhaps?

spanner/api/Program.cs
```diff
@@ -0,0 +1,989 @@
...
+                && string.IsNullOrEmpty(endTitle))
+            {
+                // Only startTitle provided, show message that user must 
+                // provide both startTitle and endTitle or exclude them 
+                // from their command.
+                Console.WriteLine("Please provide values for both the start "
+                    + "of the title index and the end of the title index.");
+                Console.WriteLine("Or you can exclude both to run the "
+                    + "query with default values.");
+                return 0;
```
evildour: Return error code?

spanner/api/Program.cs
benwulfe: thanks Mike!

spanner/api/Program.cs
```diff
@@ -0,0 +1,989 @@
...
+        }
+
+
+        public static async Task QueryDataWithIndexAsync(
+            string projectId, string instanceId, string databaseId,
+            string startTitle = "Aardvark", string endTitle = "Goo")
+        {
+            Console.WriteLine("Running QueryDataWithIndex...");
+            // [START query_data_with_index]
+            // [START read_data_with_index]
```
benwulfe: I think he is using this sample for both query and read since c# only supports query.

spanner/api/Program.cs
```diff
@@ -0,0 +1,989 @@
...
+        }
+
+        public static object CreateSampleDatabase(string projectId,
+    string instanceId, string databaseId)
+        {
+            var response = Task.Run(async () =>
+                await CreateSampleDatabaseAsync(
+                    projectId, instanceId, databaseId));
+            Console.WriteLine("Waiting for operation to complete...");
+            Task.WaitAll(response);
```
benwulfe: in this case:   CreateSampleDatabaseAsync(projectId, instanceId, databaseId).Wait();

benwulfe: Jonathan -- I'm sure you know, but just in case -- "async" when declared on a method actually has no effect on how its used or its signature.  "async Task Foo()" and "Task Foo()" are treated the same way when consumed.  async only tells the compiler to create await continuations inside the method.

spanner/api/Program.cs
```diff
@@ -0,0 +1,989 @@
...
+                + $"{instanceId}/databases/{databaseId}";
+            string createStatement =
+                $"CREATE INDEX AlbumsByAlbumTitle2 ON Albums(AlbumTitle) "
+                + "STORING (MarketingBudget)";
+            // Make the request.
+            using (var connection = new SpannerConnection(connectionString))
+            {
+                var createCmd = connection.CreateDdlCommand(createStatement);
+                await createCmd.ExecuteNonQueryAsync();
+            }
```
jsimonweb: Done.

spanner/api/Program.cs
jsimonweb: Sticking with older approach for creating Tasks. TODO: revisit this in a future update.

PR #290, opened 2017-07-21
Adding a C# example for Cloud ML Engine's online prediction API

rhaertel80:


rhaertel80:
Fixed my email address to avoid CLA bot warning.

ml_engine/online_prediction/predict.cs
```diff
@@ -0,0 +1,117 @@
...
+
+        static async Task RunAsync()
+        {
+            client.BaseAddress = new Uri("https://ml.googleapis.com/v1/");
+            client.DefaultRequestHeaders.Accept.Clear();
+            client.DefaultRequestHeaders.Accept.Add(new MediaTypeWithQualityHeaderValue("application/json"));
+
+            try
+            {
+                Person person = new Person
```
nikhilk: Minor comment -- I'd expect you should be able to use `dynamic` here rather than have to define a Person class and use strong typing, just for filling in a bunch of fields.

ml_engine/online_prediction/predict.cs
```diff
@@ -0,0 +1,117 @@
...
+
+        // [START predict_json]
+        static async Task<dynamic> Predict(String project, String model, List<Person> instances, String version = null)
+        {
+            GoogleCredential credential = await GoogleCredential.GetApplicationDefaultAsync();
+            var bearer_token = await credential.UnderlyingCredential.GetAccessTokenForRequestAsync();
+            client.DefaultRequestHeaders.Authorization = new AuthenticationHeaderValue("Bearer", bearer_token);
+
+            var version_suffix = version == null ? "" : $"/version/{version}";
+            var model_uri = $"projects/{project}/models/{model}{version_suffix}";
```
nikhilk: From a style point of view, C# code doesn't use underscores. This would be modelUri.

ml_engine/online_prediction/predict.cs
rhaertel80: I originally did that, but then thought user in production probably has structured data. If, in fact, anonymous types are pretty ubiquitous, happy to change.

nikhilk: Well, you're already using dynamic in dealing with the response, as well as an anonymous type in constructing the request (`new { instances = instances }`) itself ... hence this bit might as well do the same.

ml_engine/online_prediction/predict.cs
```diff
@@ -0,0 +1,117 @@
...
+            catch (Exception e)
+            {
+                Console.WriteLine(e.Message);
+            }
+
+            Console.ReadLine();
+        }
+
+        // [START predict_json]
+        static async Task<dynamic> Predict(String project, String model, List<Person> instances, String version = null)
```
nikhilk: In the total nit category -- was slightly bugging me to see model and version parameters separated apart. Signature could be (instances, project, model, version). :)

PR #292, opened 2017-07-24
Add openapi-appengine.yaml file for App Engine deployments.

p-buse:
After conversation with the Endpoints team, this was the most effective solution we came up with to guide users through setting up Endpoints on App Engine. Having users uncomment code proved to be difficult. If you have another idea that could work better, please let me know. Otherwise, this will have to do.

### DIFF
--- a/endpoints/getting-started/openapi.yaml
+++ b/endpoints/getting-started/openapi.yaml
@@ -6,9 +6,6 @@ info:
   version: "1.0.0"
 host: "echo-api.endpoints.YOUR-PROJECT-ID.cloud.goog"
 # [END swagger]
-# For App Engine deployments, delete the above "host:" line and remove the "# "
-# from the following line. Then change YOUR-PROJECT-ID to your project id.
-# host: "YOUR-PROJECT-ID.appspot.com"
 consumes:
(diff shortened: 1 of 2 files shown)


endpoints/getting-started/openapi-appengine.yaml
```diff
@@ -0,0 +1,108 @@
+# [START swagger]
+swagger: "2.0"
+info:
+  description: "A simple Google Cloud Endpoints API example."
+  title: "Endpoints Example"
+  version: "1.0.0"
+host: "YOUR-PROJECT-ID.appspot.com"
```
SurferJeffAtGoogle: Please update https://github.com/p-buse/dotnet-docs-samples/blob/master/endpoints/getting-started/src/IO.Swagger/deployTests.ps1

p-buse: Done. It doesn't look like there is a .NET getting started example for GCE or GKE. Do we need openapi.yaml at all?

p-buse: I don't have write access to this repo. Could you please merge?

PR #306, opened 2017-08-02
Added EnableWordTimeOffset=true to LongRunningRecognize (Async file)

dlaqab:
Added EnableWordTimeOffset=true to LongRunningRecognize (Async file)
Added EnableWordTimeOffset=false to Sync samples

### DIFF
--- a/speech/api/Recognize/Recognize.cs
+++ b/speech/api/Recognize/Recognize.cs
@@ -99,6 +99,7 @@ namespace GoogleCloudSamples
                 Encoding = encoding,
                 SampleRateHertz = bitRate,
                 LanguageCode = "en",
+                EnableWordTimeOffsets = false,
             }, RecognitionAudio.FromFile(filePath));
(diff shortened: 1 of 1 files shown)


dlaqab:
I am covered under my employer (Google)

speech/api/Recognize/Recognize.cs
```diff
@@ -99,6 +99,7 @@ public class Recognize
                 Encoding = encoding,
                 SampleRateHertz = bitRate,
                 LanguageCode = "en",
+                EnableWordTimeOffsets = false,
```
SurferJeffAtGoogle: Why show users to set a value when it's the same as the default?

remi:
The EnableWordTimeOffset=false parameter in these samples is extraneous to the task these samples are intended to demonstrate. Closing.

remi:
Note: samples which demonstrate Enable Word Time Offsets were added to dotnet-docs-samples in #310

PR #309, opened 2017-08-03
Update auth sample.

SurferJeffAtGoogle:


auth/Program.cs
```diff
@@ -15,18 +15,72 @@
...
+                Console.WriteLine(bucket.Name);
+            }
+        }
+
+        void ExplicitAuthComputeEngine()
+        {
+            // Explicitly use Compute Engine credentials. These credentials are
+            // available on Compute Engine, App Engine Flexible, and Container Engine.
+            var computeCredential = new ComputeCredential();
+            // How to instantiate?  This gives me a compile time error.
```
SurferJeffAtGoogle: jskeet,  How can I instantiate a GoogleCredential explicitly from a compute engine credential?

jskeet: I don't know that there's any way of doing that at the moment. If you're on ComputeEngine and nothing else is configured, `GoogleCredential.GetApplicationDefaultCredentials` will give you back a wrapped `ComputeCredential`, but I don't know of a way of doing it explicitly.

@chrisdunelm: Anything you know of?

chrisdunelm: There is currently no way to do this.
The [constructor you're trying to use](https://github.com/google/google-api-dotnet-client/blob/master/Src/Support/Google.Apis.Auth/OAuth2/GoogleCredential.cs#L45) is `internal`; and there's no `From...` method that accepts an `ICredential` or `ComputeCredential`.

We *could* make this possible, but I'm not very keen to. In real use I doubt anyone is wanting to do this.

PR #313, opened 2017-08-15
add new tags for netcore versions of transaction samples.

benwulfe:
read_only_transaction_core
read_write_transaction_core

also adds a new tag for showing code on how to use the transient fault app block to handle transaction retries.
read_write_retry

(only one is needed as its the same in netcore and net45)

### DIFF
--- a/spanner/api/Program.cs
+++ b/spanner/api/Program.cs
@@ -176,6 +176,8 @@ namespace GoogleCloudSamples.Spanner
         public string instanceId { get; set; }
         [Value(2, HelpText = "The ID of the database where the sample database resides.", Required = true)]
         public string databaseId { get; set; }
+        [Value(3, HelpText = "The platform code to execute (netcore or net45).", Required = false)]
+        public string platform { get; set; }
     }
 
     [Verb("readWriteWithTransaction", HelpText = "Update data in the sample Cloud Spanner database table using a read-write transaction.")]
@@ -187,11 +189,14 @@ namespace GoogleCloudSamples.Spanner
         public string instanceId { get; set; }
         [Value(2, HelpText = "The ID of the database where the sample database resides.", Required = true)]
         public string databaseId { get; set; }
+        [Value(3, HelpText = "The platform code to execute (netcore or net45).", Required = false)]
+        public string platform { get; set; }
     }
 
     public class Program
     {
         static readonly ILog s_logger = LogManager.GetLogger(typeof(Program));
+        private static readonly string NetCorePlatform = "netcore";
 
         enum ExitCode : int
         {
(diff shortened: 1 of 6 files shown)


spanner/api/Program.cs
```diff
@@ -928,34 +1091,55 @@ internal class CustomTransientErrorDetectionStrategy
             return ExitCode.Success;
         }
 
-        public static object QueryDataWithTransaction(string projectId,
-            string instanceId, string databaseId)
+        public static object QueryDataWithTransaction(string projectId, string instanceId, string databaseId, string platform)
```
SurferJeffAtGoogle: Code lines within doc tags [START ...] [END ...] should be less than 80 characters long.  There are other lines in this PR that are > 80 chars too.

spanner/api/Program.cs
```diff
@@ -928,34 +1091,55 @@ internal class CustomTransientErrorDetectionStrategy
...
-                var retryPolicy =
-                    new RetryPolicy<CustomTransientErrorDetectionStrategy>
-                        (RetryStrategy.DefaultExponential);
-                var response = retryPolicy.ExecuteAsync(async () =>
-                await ReadWriteWithTransactionAsync(
-                    projectId, instanceId, databaseId));
                 s_logger.Info("Waiting for operation to complete...");
-                Task.WaitAll(response);
+                var response = platform == NetCorePlatform
+                    ? Task.Run(async () =>
```
SurferJeffAtGoogle: I like how you put the 4.6 vs core code in two separate functions, like you did on line 1103.  Can you do that here too?

spanner/api/Program.cs
```diff
@@ -187,11 +189,14 @@ class ReadWriteWithTransactionOptions
...
         [Value(2, HelpText = "The ID of the database where the sample database resides.", Required = true)]
         public string databaseId { get; set; }
+        [Value(3, HelpText = "The platform code to execute (netcore or net45).", Required = false)]
+        public string platform { get; set; }
     }
 
     public class Program
     {
         static readonly ILog s_logger = LogManager.GetLogger(typeof(Program));
+        private static readonly string NetCorePlatform = "netcore";
```
SurferJeffAtGoogle: I forget, will const work here?

It will have exactly the same effect, but be a stronger hint to the user that it's a constant.

spanner/api/Program.cs
```diff
@@ -176,6 +176,8 @@ class QueryDataWithTransactionOptions
         public string instanceId { get; set; }
         [Value(2, HelpText = "The ID of the database where the sample database resides.", Required = true)]
         public string databaseId { get; set; }
+        [Value(3, HelpText = "The platform code to execute (netcore or net45).", Required = false)]
+        public string platform { get; set; }
```
SurferJeffAtGoogle: Let's set this to the default value so the reader knows which is chosen by default:
```
public string platform { get; set; } = "net45";
```

spanner/api/Program.cs
```diff
@@ -187,11 +189,14 @@ class ReadWriteWithTransactionOptions
...
```
SurferJeffAtGoogle: Let's set this to the default value so the reader knows which is chosen by default:
```
public string platform { get; set; } = "net45";
```

benwulfe:
I'll give the tests a go.  shouldnt be hard if there are tests covering the net45 version

spanner/api/Program.cs
benwulfe: no const doesnt work for strings -- has to be static readonly.

spanner/api/Program.cs
benwulfe: I wanted to keep the sample snippet for read_write_retry clean.  It should focus solely on how to do retry and including the both the Core and net45 versions within a single line will introduce needless complexity.

It also uses Task.Run to allow use of "async" in the sample.

benwulfe:
I've added tests and they pass.  However, with the way the commands and tests are structured, you need to have previously set up the sample database with the following commands:
(replace spanneref with your own projectid)

spanner createSampleDatabase spanneref my-instance my-database
spanner insertSampleData spanneref my-instance my-database
spanner addColumn spanneref my-instance my-database
spanner writeDataToNewColumn spanneref my-instance my-database

I recommend making the tests runnable from scratch -- but that's a bit of work and we need to get Sandra unblocked asap.

PR #330, opened 2017-09-13
TwelveFactor shows how to load appsettings.json from Google Cloud Sto…

SurferJeffAtGoogle:
…rage.


appengine/flexible/TwelveFactor/Services/CloudStorageFileProvider.cs
```diff
@@ -0,0 +1,259 @@
...
+
+    class CloudStorageChangeToken : IChangeToken
+    {
+        List<ChangeCallback> _callbacks = new List<ChangeCallback>();
+        bool _hasChanged = false;
+
+        public bool HasChanged { 
+            get { return _hasChanged; }
+            set {
+                _hasChanged = true;
```
ivannaranjo: Looks like this property should be protected by a lock, since it will be called from separate threads and could cause trouble.

appengine/flexible/TwelveFactor/Services/CloudStorageFileProvider.cs
```diff
@@ -0,0 +1,259 @@
...
+            if (_pollingInterval == null) {
+                return null;
+            }
+            _logger.LogDebug("Watching {0}", filter);
+            var path = SplitObjectPath(filter);
+            var token = new CloudStorageChangeToken();
+            var obj = GetObject(path);
+            long? generation = obj?.Generation;
+            Task.Run(async () => {
+                while (true) {
```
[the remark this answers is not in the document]
ivannaranjo: Looking at the `CloudStorageChangeToken` class below it seems that this lambda actually should be an implementation details of that class.

appengine/flexible/TwelveFactor/Services/CloudStorageFileProvider.cs
```diff
@@ -0,0 +1,259 @@
...
+        public bool ActiveChangeCallbacks => true;
+
+        public IDisposable RegisterChangeCallback(Action<object> callback,
+            object state)
+        {
+            var disposable = new ChangeCallback() {
+                Callback = callback,
+                State = state
+            };
+            _callbacks.Add(disposable);
```
ivannaranjo: The `_callbacks` field should be protected by a lock since it can also be read from a different thread when the `HasChanged` property is set.

appengine/flexible/TwelveFactor/Services/CloudStorageFileProvider.cs
ivannaranjo: You should verify that `value` is indeed `true`. Maybe this should be a readonly property with a method for teh change? The `IChangeToken` interface only specifies a getter.

appengine/flexible/TwelveFactor/Services/CloudStorageFileProvider.cs
```diff
@@ -0,0 +1,259 @@
...
+            return disposable;
+        }
+
+        class ChangeCallback : IDisposable {
+            public Action<object> Callback { get; set; }
+            public object State {get; set;}
+
+            private object _lock = new object();
+            private bool _done = false;
```
ivannaranjo: Nit, instead of having a boolean `_done` you could just null out the `Callback` property when you raise it. That will make your implementation of `Dispose` below make more sense.

appengine/flexible/TwelveFactor/Services/CloudStorageFileProvider.cs
```diff
@@ -0,0 +1,259 @@
...
+                        Callback(State);
+                    }
+                }                
+            }
+        };
+    }
+
+    class CloudStorageFileInfo : IFileInfo
+    {
+        readonly Google.Apis.Storage.v1.Data.Object _cloudStorageObject;
```
ivannaranjo: Nit, you could use a name alias to make this more readable:
```C#
using GcsObject = Google.Apis.Storage.v1.Data.Object;
```

Then you could just refer to `GcsObject` in your code.

appengine/flexible/TwelveFactor/Services/CloudStorageFileProvider.cs
```diff
@@ -0,0 +1,259 @@
...
+
+        public DateTimeOffset LastModified => 
+            _cloudStorageObject.Updated.Value;
+
+        public bool IsDirectory => false;
+
+        public Stream CreateReadStream()
+        {
+            // Json config files are small, and will easily fit in memory.
+            var stream = new MemoryStream();
```
ivannaranjo: Nit, you could cache the contents in the `CloudStorageFileInfo` instance, so you avoid downloading the contents every time the `CreateReadStream()` method is called. Of course this might not make any difference if the method is only called once.

ivannaranjo: Is not really about the size, is more about the latency.

appengine/flexible/TwelveFactor/Startup.cs
```diff
@@ -0,0 +1,172 @@
...
+        IConfigurationRoot BuildConfiguration(IHostingEnvironment env, 
+            string cloudStorageBucket)
+        {
+            var builder = new ConfigurationBuilder()
+                .SetBasePath(env.ContentRootPath)
+                .AddJsonFile("appsettings.json", optional: true,
+                     reloadOnChange: true);
+            if (null != cloudStorageBucket) {
+                builder.AddJsonFile(_cloudStorage, string.Format(
+                    "{0}/aspnet-configs/appsettings.json", cloudStorageBucket),
```
ivannaranjo: Nit, why not use the string interpolation? As in:
```C#
$"{cloudStorageBucket}/aspnet-configs/appsettings.json"
```

appengine/flexible/TwelveFactor/Startup.cs
```diff
@@ -0,0 +1,172 @@
...
+    public class Startup
+    {
+        // Polling a file on Cloud Storage at the rate of
+        // once every 2 minutes will cost approximately $0.01 per month.
+        // See https://cloud.google.com/storage/pricing#network-pricing
+        CloudStorageFileProvider _cloudStorage = new CloudStorageFileProvider(
+            TimeSpan.FromMinutes(2));
+        public Startup(IHostingEnvironment env)
+        {
+            // Have to build the configuration twice.
```
ivannaranjo: Ouch, is there no better way to handle this? Creating the configuration twice seems like a huge inconvenience. Maybe we have to find a way for the `Google.Api.Gax.Platform.Instance()` to return the right value when running locally?

My problem with this comes from the fact that we're doing this so the user can override the project id in the local `appsettings.{env}.json` file. We should find a better way of doing this. Could we simulate this with environment variables or something?

appengine/flexible/TwelveFactor/Startup.cs
```diff
@@ -0,0 +1,172 @@
...
+        }
+
+        private string GetVersion()
+        {
+            var instance = Google.Api.Gax.Platform.Instance();
+            // The source version of the service. See https://cloud.google.com/error-reporting/docs/formatting-error-messages#FIELDS.version.
+            var versionId =
+                instance?.GaeDetails?.VersionId ??
+                Configuration["Google:ErrorReporting:Version"];
+            if (string.IsNullOrEmpty(versionId))
```
ivannaranjo: This is quite harsh isn't it? Why do we need to crash the app if there's no version? We should have a suitable default that can be found in the logs/traces easily. Show a warning that the version is not set.

appengine/flexible/TwelveFactor/Startup.cs
```diff
@@ -0,0 +1,172 @@
...
+        }
+
+        private string GetServiceName()
+        {
+            var instance = Google.Api.Gax.Platform.Instance();
+            // An identifier of the service. See https://cloud.google.com/error-reporting/docs/formatting-error-messages#FIELDS.service.
+            var serviceName =
+                instance?.GaeDetails?.ServiceId ??
+                Configuration["Google:ErrorReporting:ServiceName"];
+            if (string.IsNullOrEmpty(serviceName))
```
ivannaranjo: This is quite harsh, I think that we need to rethink our stance and perhaps log a warning?

appengine/flexible/TwelveFactor/TwelveFactor.csproj
```diff
@@ -0,0 +1,41 @@
+<Project Sdk="Microsoft.NET.Sdk.Web">
+
+  <PropertyGroup>
+    <TargetFramework>netcoreapp1.0</TargetFramework>
```
ivannaranjo: This is quite an old app. I know the VS templates are old but we should not creating new apps targetting .NET Core 1.0. I suggest you move to 2.0 to showcase this.

ivannaranjo:
LGTM.

For what is worth, we use _around_ 100 chars limit on the line width, but is more of a guideline than a hard rule.

Agreed on the not supporting .NET Core 2.0, but at least 1.1? .NET Core 1.0 is **really** old.

PR #336, opened 2017-09-19
Upgrade to dotnet 2.0.

SurferJeffAtGoogle:
And upgrade Google.Cloud.Diagnostics.AspNetCore to latest.


SurferJeffAtGoogle:
It's more complicated that this.  This doesn't configure correctly.
Links:
https://www.stevejgordon.co.uk/getting-started-with-asp-net-core-2-0-preview-1
https://github.com/aspnet/MetaPackages/blob/rel/2.0.0-preview1/src/Microsoft.AspNetCore/WebHost.cs
https://docs.microsoft.com/en-us/aspnet/core/migration/1x-to-2x/

PR #350, opened 2017-10-10
Update Bigquery streaming sample.

jsimonweb:


bigquery/api/GettingStarted/BigqueryTest/BigqueryTest.cs
```diff
@@ -48,6 +49,15 @@ public class BigQueryTest
             }
         };
 
+        internal class CustomTransientErrorDetectionStrategy
+            : ITransientErrorDetectionStrategy
+        {
+            public bool IsTransient(Exception ex)
```
SurferJeffAtGoogle:
```cs
 public bool IsTransient(Exception ex) => ex is InvalidOperationException;
```

bigquery/api/GettingStarted/BigqueryTest/BigqueryTest.cs
```diff
@@ -533,18 +537,36 @@ public void TestImportDataFromFile()
         public void TestImportDataFromStream()
         {
             string datasetId = "datasetForTestImportDataFromStream";
-            string newTableId = "tableForTestImportDataFromStream";
+            Random rnd = new Random();
+            string tableNameSuffix = rnd.Next(4200000).ToString();
```
SurferJeffAtGoogle: Why pass 4200000 here?

bigquery/api/GettingStarted/BigqueryTest/BigqueryTest.cs
```diff
@@ -533,18 +537,36 @@ public void TestImportDataFromFile()
...
+                        newTableId, query, _client);
+                    var row = results.First();
+                    valueToTest = row["title"].ToString();
+                });
+            }
+            catch (Exception)
+            {
+                // All of the retries failed.
+            }
+            Assert.Equal(gcsUploadTestWord, valueToTest);
```
SurferJeffAtGoogle: If this assert fails, then DeleteTable and DeleteDataset won't be called below.  The test will leak resources.

jsimonweb:
Addressed comments. Please review.

PR #351, opened 2017-10-11
Move auth2 to replace auth.

jsimonweb:


SurferJeffAtGoogle:
As we discussed, please update the README with instructions for Visual Studio 2015 users.

PR #394, opened 2017-11-21
Update sentiment value in Language sample tests.

jsimonweb:


language/api/AnalyzeTest/AnalyzeTest.cs
```diff
@@ -116,15 +116,15 @@ public void EntitySentimentTest()
         {
             var output = Run("entity-sentiment", s_text);
             Assert.Contains("Entity Sentiment:", output.Stdout);
-            Assert.Contains("Santa Claus Conquers the Martians (31%)", output.Stdout);
+            Assert.Contains("Santa Claus Conquers the Martians (32%)", output.Stdout);
```
SurferJeffAtGoogle: Let's use a regex to ignore the number here.  Will this work?

```cs
Assert.Matches(new Regex(@"Santa Claus Conquers the Martians \([0-9]+%\)", RegexOptions.IgnoreCase), output.Stdout);
```

PR #396, opened 2017-11-22
Update Bigquery sample to .NET Core 2.0

jsimonweb:


tswast:
I see many results for `bigquery/api/GettingStarted/BigqueryTest/BigqueryTest.cs` when I codesearch for this file that has moved.

SurferJeffAtGoogle:
Indeed.  The good news is that because the existing links contain commit hashes, so the links in the live docs are not broken.  Jonathan will prepare a CL that points to the new file locations.

PR #397, opened 2017-11-30
Update Video Intelligence samples to use V1

remi:
Bump package for Video Intelligence to V1

### DIFF
--- a/video/api/Analyze/Program.cs
+++ b/video/api/Analyze/Program.cs
@@ -13,7 +13,7 @@
 // the License.
 
 using CommandLine;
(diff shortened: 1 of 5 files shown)


video/api/Analyze/Program.cs
```diff
@@ -115,25 +115,25 @@ public static object AnalyzeLabelsGcs(string uri)
         }
 
         // [START analyze_labels]
-        static void PrintLabels(string labelName,
+        static void PrintLabels(string labelName,
```
remi: FYI I'm not sure why some of these lines show in the diff, I used VS Code to edit 🤷🏼‍♀️

PR #406, opened 2017-12-15
Update instructions on main README.md

simonz130:
Main README.md didn't clarify that there may be additional customizations required to run specific samples. Fixed that.

### DIFF
--- a/README.md
+++ b/README.md
@@ -32,6 +32,8 @@ But no sample will work until you:
     4.  Give your project a name.
     5.  Make a note of the project ID, which might be different from the project
         name. The project ID is used in commands and in configurations.
+	6.  Follow README.md instructions of corresponding sample you try to run - in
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -32,6 +32,8 @@ But no sample will work until you:
     4.  Give your project a name.
     5.  Make a note of the project ID, which might be different from the project
         name. The project ID is used in commands and in configurations.
+    6.  Follow README.md instructions of corresponding sample you try to run - in
```
SurferJeffAtGoogle: This additional step is necessary and helpful to users.  Thanks for adding it.

The phrase "of corresponding sample" doesn't really tell users where to find the other README.md.  They may still be left wondering.  

I'd phrase it something like.

In each sample directory, a README.md contains additional steps necessary to run the sample.  Don't forget to follow those instructions too.

simonz130: Fixed.

PR #417, opened 2018-01-03
DO NOT MERGE

SurferJeffAtGoogle:
Examples of the ideal auth coding experience for C# developers.

Includes:
1.  Deprecating zero-argument constructors for services.  All services
    constructors require a GoogleCredential.

    The problem with zero-argument constructors is that when a reader
    simply sees
        var speech = SpeechClient.Create();
    they have no clue that there is any authentication happening or
    what they might need to do to authenticate.

    Instead, if they see:
        var credential = GoogleCredential.FromEnvironment();
        var speech = SpeechClient.Create(credential);

    They see exactly how authenticaton happens.  What's more, it's
    simple to discover other ways to authenticate, especially with an
    IDE.  I put my cursor on GoogleCredential and press F12, and my
    IDE shows me all the other ways of authenticating, i.e., FromFile(),
    FromStream(), etc.

2.  Renaming Credential.GetApplicationDefault() to
    Credential.FromEnvironment().  Users who are new to GCP have no idea
    what GetApplicationDefault() means.  At least FromEnvironment()
    tells the reader where credentials are coming from.

3.  Using a fluent coding style for adding scopes.  It reduces lines of
    code and clutter.

4.  Adding a ProjectId property to GoogleCredential.  Users are still
    puzzled as to where the project id comes from, and modifying
    code or environment variables to set it is a hassle.  See buganizer
    71516166 for a recent report.

### DIFF
--- a/auth/AuthSample/Program.cs
+++ b/auth/AuthSample/Program.cs
@@ -325,10 +325,7 @@ namespace GoogleCloudSamples
         {
             var credential = GoogleCredential.FromFile(jsonPath)
                 .CreateScoped(LanguageServiceClient.DefaultScopes);
-            var channel = new Grpc.Core.Channel(
-                LanguageServiceClient.DefaultEndpoint.ToString(),
-                credential.ToChannelCredentials());
-            var client = LanguageServiceClient.Create(channel);
+            var client = LanguageServiceClient.Create(credential);
             AnalyzeSentiment(client);
             return 0;
         }
@@ -336,17 +333,15 @@ namespace GoogleCloudSamples
         public object AuthExplicitComputeEngine(string projectId)
         {
             var credential = GoogleCredential.FromComputeCredential();
-            var channel = new Grpc.Core.Channel(
-                LanguageServiceClient.DefaultEndpoint.ToString(),
-                credential.ToChannelCredentials());
-            var client = LanguageServiceClient.Create(channel);
+            var client = LanguageServiceClient.Create(credential);
             AnalyzeSentiment(client);
             return 0;
         }
 
         public object AuthImplicit(string projectId)
         {
-            var client = LanguageServiceClient.Create();
+            var credential = GoogleCredential.FromEnvironment();
+            var client = LanguageServiceClient.Create(credential);
             AnalyzeSentiment(client);
             return 0;
         }
--- a/auth/AuthSample/Program.cs
+++ b/auth/AuthSample/Program.cs
@@ -96,14 +96,14 @@ namespace GoogleCloudSamples
         // This is the preferred way of authenticating.
         ///////////////////////////////////////////////
         // [START auth_cloud_implicit]
-        public object AuthImplicit(string projectId)
+        public object AuthImplicit()
         {
             // If you don't specify credentials when constructing the client, the
             // client library will look for credentials in the environment.
-            var credential = GoogleCredential.GetApplicationDefault();
+            var credential = GoogleCredential.FromEnvironment();
             var storage = StorageClient.Create(credential);
             // Make an authenticated API request.
-            var buckets = storage.ListBuckets(projectId);
+            var buckets = storage.ListBuckets(credential.ProjectId);
             foreach (var bucket in buckets)
             {
                 Console.WriteLine(bucket.Name);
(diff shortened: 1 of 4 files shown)


jonparrott:
> Deprecating zero-argument constructors for services. All services
constructors require a GoogleCredential.

I'm okay with this for C# (and possibly even Java), but I don't think it makes sense for the dynamic languages. The boilerplate is small, but I think it just adds unnecessary overhead.

> Renaming Credential.GetApplicationDefault() to
Credential.FromEnvironment().

Totally cool with this. In Python it's `google.auth.default()` which is think is pretty succinct, but `google.auth.from_env()` would be just as valid.

> Adding a ProjectId property to GoogleCredential.

One of the things here that concerns me is that projects are completely separate from credentials and are a distinctly separate resource. You can have credentials that belong to Project A access resources in Projects B, C, and D. Treating them as a singular entity is problematic. In Python, `google.auth.default()` returns `Tuple[Credentials, Optional[str]]` of `credentials, project_id`. Our *client* constructors come in two flavors:

* `Client(credentials=Optional[Credentials], project=Optional[str])` for hand-written clients (mostly storage clients) where not specifying the credentials will use `google.auth.default()` to get credentials and project from the environment. Specifying credentials *requires* specifying the project.

* `Client(credentials=Optional[Credentials])` for autogenerated clients and non-storage clients. The project id, if it's ever needed, is specified in the rpc, for example `pubsub.list_topics(project='my-project')`.

However, if `GoogleCredentials` is only ever used as the concrete return type of `FromEnvironment()` I'm cool with it.

(Aside: A lot of the other languages have `GoogleCredentials` but we murdered that in Python)

datastore/api/QuickStart/QuickStart.cs
```diff
@@ -22,11 +23,11 @@ public class QuickStart
     {
         public static void Main(string[] args)
         {
-            // Your Google Cloud Platform project ID
-            string projectId = "YOUR-PROJECT-ID";
-
+            var credential = GoogleCredential.FromEnvironment();
             // Instantiates a client
-            DatastoreDb db = DatastoreDb.Create(projectId);
+            DatastoreDb db = DatastoreDb.Create(credential);
```
simonz130: Isn't projectId embedded in JWT that is used when you generate credential using *.FromEnvironment method?
If yes, we shouldn't have an overload that allows specifying project id - not only it's a duplicate information, but customer may specify id that is incompatible with provided credential.

auth/AuthSample/Program.cs
```diff
@@ -90,14 +90,14 @@ public class CloudLibrary : AuthLibrary
...
+        public object AuthImplicit()
         {
             // If you don't specify credentials when constructing the client, the
             // client library will look for credentials in the environment.
-            var credential = GoogleCredential.GetApplicationDefault();
+            var credential = GoogleCredential.FromEnvironment();
             var storage = StorageClient.Create(credential);
             // Make an authenticated API request.
-            var buckets = storage.ListBuckets(projectId);
+            var buckets = storage.ListBuckets(credential.ProjectId);
```
simonz130: Why is projectId needed after you already specified the credential?

pubsub/api/QuickStart/QuickStart.cs
```diff
@@ -17,25 +17,20 @@
...
-            string projectId = "YOUR-PROJECT-ID";
-            // [END pubsub_quickstart]
-            Debug.Assert(projectId != "YOUR-PROJECT" + "-ID",
-                "Edit Program.cs and replace YOUR-PROJECT-ID with your project id.");
-            // [START pubsub_quickstart]
+            PublisherClient publisher = PublisherClient.Create(credential);
 
             // The name for the new topic
-            var topicName = new TopicName(projectId, "my-new-topic");
+            var topicName = new TopicName(credential.ProjectId, "my-new-topic");
```
simonz130: Why do we need to specify the projectId to create a topic name? Topic name is data entity that is provided to publisher's CreateTopic method. When you hold "publisher" instance, you already authenticated using credential that was passed in static factory method of PublisherClient. Therefore, the API call should know which project Id to work against when creating the topic.

PR #419, opened 2018-01-05
Add BigQuery Data Transfer API quickstart

tswast:
This sample lists the data sources enabled for a project.

### DIFF
--- a/bigquery/data-transfer/api/QuickStart/Program.cs
+++ b/bigquery/data-transfer/api/QuickStart/Program.cs
@@ -1,0 +1,46 @@
+﻿/*
+ * Copyright (c) 2018 Google LLC
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License"); you may not
+ * use this file except in compliance with the License. You may obtain a copy of
+ * the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS, WITHOUT
+ * WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied. See the
+ * License for the specific language governing permissions and limitations under
+ * the License.
+ */
+// [START bigquery_data_transfer_quickstart]
+
+using System;
+using Google.Api.Gax;
+using Google.Cloud.BigQuery.DataTransfer.V1;
+
+namespace QuickStart
+{
+    class Program
+    {
+        static void Main(string[] args)
+        {
+            // Instantiates a client
+            DataTransferServiceClient client = DataTransferServiceClient.Create();
+
+            // Your Google Cloud Platform project ID
+            string projectId = "YOUR-PROJECT-ID";
+
+            ProjectName project = new ProjectName(projectId);
+            PagedEnumerable<ListDataSourcesResponse, DataSource> sources = client.ListDataSources(ParentNameOneof.From(project));
+            Console.WriteLine("Supported Data Sources:");
+            foreach (DataSource source in sources)
+            {
+                Console.WriteLine($"{source.DataSourceId}: {source.DisplayName} ({source.Description})");
+            }
+            Console.WriteLine("\nPress any key...");
+            Console.ReadKey();
+        }
+    }
+}
+// [END bigquery_data_transfer_quickstart]
--- a/bigquery/data-transfer/api/QuickStart/QuickStart.cs
+++ b/bigquery/data-transfer/api/QuickStart/QuickStart.cs
@@ -19,11 +19,11 @@ using System;
 using Google.Api.Gax;
 using Google.Cloud.BigQuery.DataTransfer.V1;
 
-namespace QuickStart
+namespace GoogleCloudSamples
 {
-    class Program
+    public class QuickStart
     {
-        static void Main(string[] args)
+        public static void Main(string[] args)
         {
             // Instantiates a client
             DataTransferServiceClient client = DataTransferServiceClient.Create();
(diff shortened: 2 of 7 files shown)


bigquery/data-transfer/api/QuickStart/Program.cs
```diff
@@ -0,0 +1,46 @@
...
+        static void Main(string[] args)
+        {
+            // Instantiates a client
+            DataTransferServiceClient client = DataTransferServiceClient.Create();
+
+            // Your Google Cloud Platform project ID
+            string projectId = "YOUR-PROJECT-ID";
+
+            ProjectName project = new ProjectName(projectId);
+            PagedEnumerable<ListDataSourcesResponse, DataSource> sources = client.ListDataSources(ParentNameOneof.From(project));
```
SurferJeffAtGoogle: Because this code lies between START and END tags, keep the line length below 80 characters.

bigquery/data-transfer/api/QuickStart/Program.cs
```diff
@@ -0,0 +1,46 @@
...
+            Console.WriteLine("Supported Data Sources:");
+            foreach (DataSource source in sources)
+            {
+                Console.WriteLine($"{source.DataSourceId}: {source.DisplayName} ({source.Description})");
+            }
+            Console.WriteLine("\nPress any key...");
+            Console.ReadKey();
```
SurferJeffAtGoogle: I see that https://github.com/GoogleCloudPlatform/dotnet-docs-samples/blob/47aa6ac0ad2b9db68c484828441cfc583ee341f7/bigquery/api/BigquerySample/Program.cs contains a call to ReadKey(), so you followed the pattern and added it here.

But, we have stopped calling ReadKey() at the end of quick starts.  The BigquerySample was the last left over.  Please remove it here.

bigquery/data-transfer/api/QuickStart/Program.cs
SurferJeffAtGoogle: How do you like
```
var sources = client.ListDataSources(ParentNameOneof.From(project));
```
?

bigquery/data-transfer/api/QuickStart/QuickStart.csproj
```diff
@@ -0,0 +1,12 @@
+<Project Sdk="Microsoft.NET.Sdk">
+
+  <PropertyGroup>
+    <OutputType>Exe</OutputType>
+    <TargetFramework>netcoreapp2.0</TargetFramework>
+  </PropertyGroup>
+
+  <ItemGroup>
+    <PackageReference Include="Google.Cloud.BigQuery.DataTransfer.V1" Version="1.0.0-beta01" />
```
SurferJeffAtGoogle: Does this sample require beta01?  Or will it also work with 1.0.0?

bigquery/data-transfer/api/QuickStart/Program.cs
tswast: Done.

tswast: I like it. At first I worried they wouldn't be able to find the types easily, but I think this makes it easier to find the method we are calling so that they can look it up.

bigquery/data-transfer/api/QuickStart/Program.cs
tswast: Done.

bigquery/data-transfer/api/QuickStart/QuickStart.csproj
tswast: 1.0.0 doesn't exit yet.

> Install the Google.Cloud.BigQuery.DataTransfer.V1 package from NuGet. Add it to your project in the normal way (for example by right-clicking on the project in Visual Studio and choosing "Manage NuGet Packages..."). Please **ensure you enable pre-release packages** (for example, in the Visual Studio NuGet user interface, check the "Include prerelease" box).

https://googlecloudplatform.github.io/google-cloud-dotnet/docs/Google.Cloud.BigQuery.DataTransfer.V1/index.html

tswast:
Looked like the Kokoro failure was on lint with the new test class. I re-ran Format-Code.

tswast:
@SurferJeffAtGoogle It looks like Kokoro is happy finally.

bigquery/data-transfer/api/QuickStart/QuickStart.cs
```diff
@@ -0,0 +1,46 @@
...
+{
+    public class QuickStart
+    {
+        public static void Main(string[] args)
+        {
+            // Instantiates a client
+            DataTransferServiceClient client = DataTransferServiceClient.Create();
+
+            // Your Google Cloud Platform project ID
+            string projectId = "YOUR-PROJECT-ID";
```
simonz130: please, add an explicit check that will throw exception if user runs your sample straight away and forgets to replace projectId value with real value.

bigquery/data-transfer/api/QuickStartTest/QuickStartTest.cs
```diff
@@ -0,0 +1,31 @@
...
+ */
+
+using Xunit;
+
+namespace GoogleCloudSamples
+{
+    public class QuickStartTest
+    {
+        [Fact]
+        public void TestMain()
```
simonz130: Please, name the test so that it's clear from the name what are you testing.

bigquery/data-transfer/api/QuickStartTest/QuickStartTest.cs
```diff
@@ -0,0 +1,31 @@
...
+        {
+            // Main() will throw an exception if something fails.
+            QuickStart.Main(new string[] { });
```
simonz130: Typically, you wouldn't write tests this way for real application. I'd suggest you extract main's method into a separate method, then test that method. Main method will call the same method too. It may look like we add more unnecessary work here, but it will mimic real world better.

bigquery/data-transfer/api/QuickStart/QuickStart.cs
tswast: That would complicate the sample (much undesired for a "quickstart") and is inconsistent with other quickstart samples in this repo such as https://github.com/GoogleCloudPlatform/dotnet-docs-samples/blob/master/pubsub/api/QuickStart/QuickStart.cs

bigquery/data-transfer/api/QuickStartTest/QuickStartTest.cs
tswast: This file is exactly copied from https://github.com/GoogleCloudPlatform/dotnet-docs-samples/blob/master/datastore/api/QuickStartTest/QuickStartTest.cs

PR #427, opened 2018-01-17
Update appengine READMEs to require VS 2017, because they do.

SurferJeffAtGoogle:
Change-Id: I83976b21730f9821d7e7da7f91d85aafe8640a9d

### DIFF
--- a/appengine/flexible/Analytics/README.md
+++ b/appengine/flexible/Analytics/README.md
@@ -10,37 +10,20 @@ when running in Google App Engine Flexible Environment.
 2.  Install the [Google Cloud SDK](https://cloud.google.com/sdk/).  The Google Cloud SDK
     is required to deploy .NET applications to App Engine.
 
-2.  Install the [.NET Core SDK, version 1.1](https://github.com/dotnet/core/blob/master/release-notes/download-archives/1.1.4-download.md).
-
-2.  Visual Studio is *optional*.  Solution files can be opened with Visual
-    Studio 2017 or later.  An old, unmaintained branch of samples that work
-    with Visual Studio 2015 is 
-    [here](https://github.com/GoogleCloudPlatform/dotnet-docs-samples/tree/vs2015).
-
-3.  [Google Cloud Tools for Visual Studio](
-	https://marketplace.visualstudio.com/items?itemName=GoogleCloudTools.GoogleCloudPlatformExtensionforVisualStudio)
-    make it easy to deploy to App Engine.  Install them if you are running Visual Studio.
-
-4.  [Create a Google Analytics Property and obtain the Tracking ID](
-    https://support.google.com/analytics/answer/1042508?ref_topic=1009620).
+3.  Install the [.NET Core SDK, version 1.1](https://github.com/dotnet/core/blob/master/release-notes/download-archives/1.1.4-download.md).
 
 4.  Edit [appsettings.json](appsettings.json).  Replace `your-google-analytics-tracking-id` with your google analytics tracking id.
 
-## Run Locally
+## ![PowerShell](../.resources/powershell.png)Using PowerShell
 
-### ![PowerShell](../.resources/powershell.png)Using PowerShell
+### Run Locally
+
 ```psm1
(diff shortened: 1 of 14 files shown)


appengine/flexible/Analytics/README.md
```diff
@@ -10,11 +10,12 @@ when running in Google App Engine Flexible Environment.
 2.  Install the [Google Cloud SDK](https://cloud.google.com/sdk/).  The Google Cloud SDK
     is required to deploy .NET applications to App Engine.
 
-2.  Install the [.NET Core SDK, version LTS](https://www.microsoft.com/net/download/core#/lts).
+2.  Install the [.NET Core SDK, version 1.1](https://github.com/dotnet/core/blob/master/release-notes/download-archives/1.1.4-download.md).
 
-2.  Visual Studio is not required, but to build and run .NET *core* applications,
-    Visual Studio 2015 users need to download and install 
-	[.NET Core tools](https://www.microsoft.com/net/core#windowsvs2015).
+2.  Visual Studio is *optional*.  Solution files can be opened with Visual
```
simonz130: This is confusing. Visual Studio Code is also Visual Studio. You probably mean Visual Studio 2017 Professional and up here.
I think we should provide instructions for both users of Visual Studio Code as well as full Visual Studio. For Code solution files are irrelevant.

Perhaps, you can have two sections - one for Visual Studio Code and another for VS 2017.
Also maybe worth having instructions for build that are uniform (dotnet build from command line), instead of relying on building from visual studio.

appengine/flexible/CloudSql/README.md
```diff
@@ -10,11 +10,12 @@ when running in Google App Engine Flexible Environment.
 2.  Install the [Google Cloud SDK](https://cloud.google.com/sdk/).  The Google Cloud SDK
     is required to deploy .NET applications to App Engine.
 
-2.  Install the [.NET Core SDK, version LTS](https://www.microsoft.com/net/download/core#/lts).
+2.  Install the [.NET Core SDK, version 1.1](https://github.com/dotnet/core/blob/master/release-notes/download-archives/1.1.4-download.md).
 
-2.  Visual Studio is not required, but to build and run .NET *core* applications,
-    Visual Studio 2015 users need to download and install 
-	[.NET Core tools](https://www.microsoft.com/net/core#windowsvs2015).
+2.  Visual Studio is *optional*.  Solution files can be opened with Visual
```
simonz130: Ditto (and I guess it's the same for the rest of the files)

appengine/flexible/Analytics/README.md
SurferJeffAtGoogle: That's interesting.  I haven't heard the point of view that "Visual Studio Code is also Visual Studio" before.  In mind, they have always been separate, unrelated entities.  Well, unrelated except for the company than builds them.

Visual Studio *Code* users can use the Powershell instructions below.  My assumption is that VSCode users are more comfortable at the command line than VS users.  Does that match your expectations and experience?

How would you rephrase these 2 paragraphs?

simonz130: I don't think "VSCode" users are better with command line than "VS" users. You probably got this feeling because VSCode is very limited in functionality compared to full version and therefore requires you to do a lot more from the command line, which is easy to open embedded in VSCode as opposed to full version.

I'd suggest putting Visual Studio instructions last. Put the required prerequisites first:

You can open the sources in either [VSCode](https://code.visualstudio.com/) or [Visual Studio 2017](https://www.visualstudio.com/downloads/).
If you work with earlier versions of Visual Studio, there is a deprecated branch of samples that work with Visual Studio 2015 [here](https://github.com/GoogleCloudPlatform/dotnet-docs-samples/tree/vs2015).

* If you use <b>VSCode</b>, open folder <<"folder name here">>
* If you use <b>Visual Studio 2017</b>, open <<"mention relevant sln or csproj file">>

==
In my opinion this way it's clearer that you distinguish between VSCode and Visual Studio. You can also put instructions on how to build from command line.

PR #431, opened 2018-01-23
Update pubsub sample to use beta16 release

chrisdunelm:
This release renames various classes. See http://googlecloudplatform.github.io/google-cloud-dotnet/docs/Google.Cloud.PubSub.V1/ for details.

### DIFF
--- a/pubsub/api/PubsubSample/Program.cs
+++ b/pubsub/api/PubsubSample/Program.cs
@@ -195,7 +195,7 @@ namespace GoogleCloudSamples
         public static object CreateTopic(string projectId, string topicId)
         {
             // [START create_publisher_client]
-            PublisherClient publisher = PublisherClient.Create();
+            PublisherServiceApiClient publisher = PublisherServiceApiClient.Create();
             // [END create_publisher_client]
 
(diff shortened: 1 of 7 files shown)


pubsub/api/PubsubSample/Program.cs
```diff
@@ -195,7 +195,7 @@ public class Pubsub
         public static object CreateTopic(string projectId, string topicId)
         {
             // [START create_publisher_client]
-            PublisherClient publisher = PublisherClient.Create();
+            PublisherServiceApiClient publisher = PublisherServiceApiClient.Create();
```
simonz130: minor: It would be easier if the cases when the type of the variable is clear from the assignment, to use "var".
For example,

var publisher = PublisherServiceApiClient.Create();

That way refactoring are easier, you write less code without sacrificing the readability.

chrisdunelm: True, but having explicit types (not `var`) can be useful in samples, as it makes the type clearer for users who are not familiar with our code.
And this code shouldn't be changed again. We expect this to be the final beta release before the GA.

PR #433, opened 2018-01-24
Adds IoT manager sample.

gguuss:


iot/api/CloudIotTest/CloudIotTest.cs
```diff
@@ -0,0 +1,334 @@
...
+{
+    // <summary>
+    /// Runs the sample app's methods and tests the outputs
+    // </summary>
+    public class CommonTests
+    {
+        private static readonly string s_projectId = Environment.GetEnvironmentVariable("GOOGLE_PROJECT_ID");
+        private static readonly string s_regionId = "us-central1";
+        private static readonly string s_serviceAccount = "serviceAccount:cloud-iot@system.gserviceaccount.com";
+
```
simonz130: nit: please, remove 2 empty lines

iot/api/CloudIotTest/CloudIotTest.cs
```diff
@@ -0,0 +1,334 @@
...
+namespace GoogleCloudSamples
+{
+    // <summary>
+    /// Runs the sample app's methods and tests the outputs
+    // </summary>
+    public class CommonTests
+    {
+        private static readonly string s_projectId = Environment.GetEnvironmentVariable("GOOGLE_PROJECT_ID");
+        private static readonly string s_regionId = "us-central1";
```
simonz130: Should this be specified by customer to work for them?
Similar comment for line 30.
If yes, please, replace the value with something like "SPECIFY_GCP_REGION_HERE"...

iot/api/CloudIotTest/CloudIotTest.cs
```diff
@@ -0,0 +1,334 @@
...
+        private readonly RetryRobot _retryRobot = new RetryRobot()
+        {
+            RetryWhenExceptions = new[] { typeof(Xunit.Sdk.XunitException) }
+        };
+
+        private void Eventually(Action action) => _retryRobot.Eventually(action);
+
+        private void CreatePubSubTopic(String name)
+        {
+            var publisher = PublisherClient.Create();
```
simonz130: Let's be consistent.
If you use "var" for static type specifier in line 53, you should also do that in line 54.

iot/api/CloudIotTest/CloudIotTest.cs
```diff
@@ -0,0 +1,334 @@
...
+        {
+            PublisherClient publisher = PublisherClient.Create();
+            TopicName topicName = new TopicName(s_projectId, name);
+            publisher.DeleteTopic(topicName);
+        }
+
+        [Fact]
+        public void TestCreateDeleteRegistry()
+        {
+            var registryId = "testcreatereg";
```
simonz130: It can be a problem if two people will run the same set of tests in parallel using the same project Id. We should add a degree of uniqueness to object names if we use real GCP APIs and not mocks.

iot/api/CloudIotTest/test/data/ec_public.pem
```diff
@@ -0,0 +1,4 @@
+-----BEGIN PUBLIC KEY-----
+MFkwEwYHKoZIzj0CAQYIKoZIzj0DAQcDQgAES5Mo86JNproHuRgdDxwadgXcMB9L
```
simonz130: Instead of pushing the changelist with sample cert, should we rather put instructions on how to create a sample cert as part of the demo setup?

iot/api/CloudIotTest/CloudIotTest.cs
```diff
@@ -28,7 +28,7 @@ public class CommonTests
         private static readonly string s_projectId = Environment.GetEnvironmentVariable("GOOGLE_PROJECT_ID");
         private static readonly string s_regionId = "us-central1";
         private static readonly string s_serviceAccount = "serviceAccount:cloud-iot@system.gserviceaccount.com";
-
+        private static readonly string s_testID = Guid.NewGuid().ToString().Split("-")[0];
```
simonz130: There is existing method (RandomName) to create randomness in this codebase:
https://github.com/GoogleCloudPlatform/dotnet-docs-samples/blob/master/testutil/TestUtil.cs

I'll leave it up to your judgement whether to leave it as is or utilize TestUtil.

iot/api/CloudIotTest/CloudIotTest.cs
simonz130: If this shouldn't be specified by the customer, please, add a comment to explain why it's always us-cental1.

iot/api/CloudIotTest/CloudIotTest.cs
SurferJeffAtGoogle: Using var on line 53 and TopicName on line 54 doesn't strike me as difficult to read.  What are the benefits to consistently using typenames or var?

simonz130: When the static type of the variable is obvious from the expression, "var" is preferable IMO because:
  1. You write less code without sacrificing the readability
  2. If you later need to refactor the code, there will be less visible changes (text diffs)

This is in line with the [guidelines](https://docs.microsoft.com/en-us/dotnet/csharp/programming-guide/inside-a-program/coding-conventions) that are commonly used by MS around "var". (Search for "Implicitly Typed Local Variables" in the doc).

iot/api/CloudIotTest/CloudIotTest.cs
gguuss: +1 -- Done.  The only potential issue before with collisions was if the teardown failed, the tests would potentially fail on next run.  Using part of a GUID should prevent this.

iot/api/CloudIotTest/CloudIotTest.cs
gguuss: Done.

iot/api/CloudIotTest/CloudIotTest.cs
gguuss: Nope, this is just for testing, using the hardcoded values should be totally acceptable.

iot/api/CloudIotTest/CloudIotTest.cs
gguuss: Done.

iot/api/CloudIotTest/test/data/ec_public.pem
gguuss: This is just used for testing, the README.md in the folder should deter folks from registering devices (they would be unusable without the private keys).

PR #438, opened 2018-02-03
Simplify IAP sample.

SurferJeffAtGoogle:


identity-aware-proxy/IAPClient/IAPClient.cs
```diff
@@ -48,44 +50,32 @@ class IAPClient
...
-                    (int)result.StatusCode, result.ReasonPhrase,
-                    responseContent));
-            }
-            string token = JsonConvert.DeserializeObject<IapResponse>(
-                responseContent).IdToken;
+                Assertion = jwtBasedAccessToken
+            };
+            var result = req.ExecuteAsync(saCredential.HttpClient,
+                saCredential.TokenServerUrl, CancellationToken.None,
+                saCredential.Clock).Result;
```
simonz130: I suggest we make this sample code async method and don't use .Result() but rather await on the result of the request.

identity-aware-proxy/IAPClient/IAPClient.cs
```diff
@@ -128,9 +118,8 @@ class IAPClient
 
             // Encryption algorithm must be RSA SHA-256, according to
             // https://developers.google.com/identity/protocols/OAuth2ServiceAccount
-            SecurityKey key = new RsaSecurityKey(
-                Pkcs8.DecodeRsaParameters(privateKey));
-            var signingCredentials = new SigningCredentials(key,
+            var signingCredentials = new SigningCredentials(
```
simonz130: This is awesome!
Since now the sample is pretty straightforward, we can make it frictionless for the customers by exposing an API that gets service account and client Id as input and returns the token as output. Can we implement it in the auth library code as a separate change?

identity-aware-proxy/IAPClient/IAPClient.cs
SurferJeffAtGoogle: Let's discuss this more at tomorrow's meeting.

identity-aware-proxy/IAPClient/IAPClient.cs
SurferJeffAtGoogle: That would make things easier for users.  However, it's a lower priority than other stuff I need to do this quarter.  If you want to create a pull request, feel free.

PR #439, opened 2018-02-05
Fix csproj to not point to .net core 1.x

simonz130:
A few minor things that prevented successful build


SurferJeffAtGoogle:
Did you intend to pull into master?

simonz130:
Hm... There are two commits that should go together to master from my
forked branch after they are approved and tests are passing. Should it be
done differently?

PR #445, opened 2018-02-08
Fix a race condition around deleting topics and subscriptions.

SurferJeffAtGoogle:


pubsub/api/PubsubTest/PubsubTest.cs
```diff
@@ -42,15 +42,60 @@ public class PubsubTest : IDisposable
...
+        /// <summary>
+        /// Handle a special race condition that can occur when deleting
+        /// something:
+        /// 1. Delete request times out.
+        /// 2. Delete operation continues on server and succeeds.
+        /// 3. Later requests to delete the same entity see NotFound error.
+        /// </summary>
+        /// <param name="delete">The delete operation to run.</param>
+        /// <returns>An action to run inside Eventually().</returns>
+        Action HandleDeleteRace(Action delete)
```
simonz130: Normally, I'd test that this delegate is not null, however not sure if you should do it in sample code...

pubsub/api/PubsubTest/PubsubTest.cs
```diff
@@ -42,15 +42,60 @@ public class PubsubTest : IDisposable
...
+                }
+            };
+        }
+
         void IDisposable.Dispose()
         {
             foreach (string topicId in _tempTopicIds)
             {
-                Eventually(() => Run("deleteTopic", _projectId, topicId));
+                Eventually(HandleDeleteRace(() =>
```
simonz130: We should not throw exceptions in Dispose method - even if it's a sample code.
Dispose is executed in a finally block which is usually executed in the context of resources cleanup.
https://msdn.microsoft.com/en-us/library/bb386039.aspx
Also, Dispose should implement a logic to prevent freeing resources twice (if it's called by mistake twice) and still, even if called twice, it shouldn't throw exception.
https://msdn.microsoft.com/en-us/library/system.idisposable.dispose(v=vs.110).aspx

In your case HandleDeleteRace can throw exception.

pubsub/api/PubsubTest/PubsubTest.cs
SurferJeffAtGoogle: If someone passes in a null delegate, the test fails, we see an error report and fix it.  Which is exactly the same as what we'd have to do if we added a null test here.

pubsub/api/PubsubTest/PubsubTest.cs
SurferJeffAtGoogle: xUnit reuses/abuses the `IDisposable` interface for the purposes of test clean up.  It's described here: https://xunit.github.io/docs/shared-context.html

I wish xUnit had defined a separate interface instead of reusing `IDisposable` so it would be less confusing.  In short, the usual rules for `Dispose()` are overkill and do not apply to xUnit tests.  

In summary:
1. The xUnit framework is the only thing that will ever call this `Dispose()` method.
2. The xUnit framework is will call the method exactly once.
3. The xUnit framework is will recover if an exception is thrown.

simonz130: Yeah, I just saw that and I really don't like what they did - they say to initialize everything per test in constructor, which is also antipattern and they call it idiomatic solution... If exception is thrown, the test framework catches that but the callstack and result reporting is misleading. The problem is that your test could succeed, test cleanup failed and it will show that the test failed with callstack from Dispose(). There should be clear distinction between test failure vs. test setup/cleanup failure.  The only upside from abusing Dispose and doing initialization in c-tor is that you can mark fields as readonly.

Anyway, I don't want to block you as I understand that the problem stems from xUnit.
I'll try to think about a way to make xUnit test initialization more tester friendly without abusing language features.

pubsub/api/QuickStart/QuickStart.cs
```diff
@@ -22,16 +22,19 @@ namespace GoogleCloudSamples
...
         {
             // Instantiates a client
             PublisherServiceApiClient publisher = PublisherServiceApiClient.Create();
 
             // Your Google Cloud Platform project ID
             string projectId = "YOUR-PROJECT-ID";
             // [END pubsub_quickstart]
-            Debug.Assert(projectId != "YOUR-PROJECT" + "-ID",
-                "Edit Program.cs and replace YOUR-PROJECT-ID with your project id.");
+            if (projectId == "YOUR-PROJECT" + "-ID")
```
simonz130: I like it but it looks like not all the samples follow this pattern. Should we fix all of them later on so all of them are consistent?

pubsub/api/PubsubTest/PubsubTest.cs
```diff
@@ -42,15 +42,60 @@ public class PubsubTest : IDisposable
...
+                    Run("deleteTopic", _projectId, topicId)));
             }
             foreach (string subscriptionId in _tempSubscriptionIds)
             {
-                Eventually(() => Run("deleteSubscription", _projectId, subscriptionId));
+                Eventually(HandleDeleteRace(() =>
+                    Run("deleteSubscription", _projectId, subscriptionId)));
```
simonz130: I think you should reverse the order of subscriptions deletion and topic deletion.
This is what I've found in documentation [here](https://cloud.google.com/pubsub/docs/admin#delete_a_topic):
**When you delete a topic, its subscriptions are not deleted, and the subscription's message backlog is available for subscribers. After a topic is deleted, its subscriptions have the topic name _deleted-topic_.**

pubsub/api/PubsubTest/PubsubTest.cs
```diff
@@ -42,15 +42,60 @@ public class PubsubTest : IDisposable
...
+            };
+        }
+
         void IDisposable.Dispose()
         {
             foreach (string topicId in _tempTopicIds)
             {
-                Eventually(() => Run("deleteTopic", _projectId, topicId));
+                Eventually(HandleDeleteRace(() =>
+                    Run("deleteTopic", _projectId, topicId)));
```
simonz130: What you can do is catch the exception you throw in "HandleDeleteRace" here and if exception is thrown, you can do:
Assert.True(false, "Cleanup of the topic named xxx has failed. More details: xxx")
(Apparently, xunit doesn't support Assert.Fail...)

The reason I think this is better than just let the exception propagate:

1) When looking at the failing test (even though it's test cleanup that failed, corresponding test is marked as failed...) the message will say the real reason of what happened as opposed to "Exception of type xxx was thrown". Will make it easier to troubleshoot even without delving into the code.
2) It will look as best effort to follow best practices within xunit implementation constraints.

PR #448, opened 2018-02-14
SignUrl for Cloud CDN

simonz130:
Sign URL sample for Cloud CDN using .NET Core


signUrl/README.md
```diff
@@ -0,0 +1,56 @@
...
+## Build and Run
+
+1.  **Follow the instructions in the [root README](../../README.md)**.
+
+2.  Follow the instructions here
+    [https://cloud-dot-devsite.googleplex.com/cdn/docs/signed-urls#creatingkeys](https://cloud-dot-devsite.googleplex.com/cdn/docs/signed-urls#creatingkeys)
+    to create the keys that will be used to sign the URLs, upload them and give relevant permissions.
+
+3.  From a Powershell command line, run to see command line arguments.
+    ```
```
ahmetb: there's probably an indentation error here. it doesn't show right on the preview.

signUrl/SignUrlSample/Program.cs
```diff
@@ -0,0 +1,119 @@
...
+
+        private static long ToUnixTime(DateTime date)
+        {
+            var epoch = new DateTime(1970, 1, 1, 0, 0, 0, DateTimeKind.Utc);
+            return Convert.ToInt64((date - epoch).TotalSeconds);
+        }
+
+        private static byte[] Base64UrlDecode(string arg)
+        {
+            string s = arg;
```
ahmetb: new variable declaration is not necessary

signUrl/SignUrlSample/Program.cs
```diff
@@ -0,0 +1,119 @@
...
+
+            long unixTimestampExpiration = ToUnixTime(expirationTime);
+
+            char queryParam = string.IsNullOrEmpty(builder.Query) ? '?' : '&';
+            builder.Query += $"{queryParam}Expires={unixTimestampExpiration}&KeyName={keyName}".ToString();
+
+            // Key is passed as base64url encoded
+            byte[] decodedKey = Base64UrlDecode(encodedKey);
+
+            // Computes SHA-1 hash of the URL using the key
```
ahmetb: HMAC SHA-1

signUrl/README.md
```diff
@@ -0,0 +1,56 @@
...
+
+  --name          The name of the key to use when encrypting url
+
+  --path          The path to the key to use when encrypting url
+
+  -e              Url expiration in UTC formatted as: YYYY-MM-DDThh:mm:ss
+
+  --help          Display this help screen.
+
+  --version       Display version information.
```
ahmetb: does `--version` actually do anything?  if it does, is it necessary?

signUrl/README.md
```diff
@@ -0,0 +1,56 @@
...
+   SignUrlSample 1.0.0
+   Copyright (C) 2018 SignUrlSample
+   ERROR(S):
+   A required value not bound to option name is missing.
+
+  --name          The name of the key to use when encrypting url
+
+  --path          The path to the key to use when encrypting url
+
+  -e              Url expiration in UTC formatted as: YYYY-MM-DDThh:mm:ss
```
ahmetb: --name & --path are full but this is short (-e), looks incoherent.

signUrl/README.md
simonz130: It's built into CommandLineParser library. This and --help are always automatically added. Version prints the assembly version. You can't remove it.

ahmetb: got it. in that case we can keep.

signUrl/README.md
```diff
@@ -0,0 +1,56 @@
...
+3.  From a Powershell command line, run to see command line arguments.
+    ```
+   PS C:\dev\dotnet-docs-samples\signUrl\SignUrlSample> dotnet run
+   Using launch settings from C:\dev\dotnet-docs-samples\signUrl\SignUrlSample\Properties\launchSettings.json...
+   SignUrlSample 1.0.0
+   Copyright (C) 2018 SignUrlSample
+   ERROR(S):
+   A required value not bound to option name is missing.
+
+  --name          The name of the key to use when encrypting url
```
ahmetb: should be `--key-name`

signUrl/README.md
```diff
@@ -0,0 +1,56 @@
...
+
+  --path          The path to the key to use when encrypting url
```
ahmetb: should be ideally `--key-path` or `--key-file`

signUrl/README.md
```diff
@@ -0,0 +1,56 @@
...
+
+## Build and Run
+
+1.  **Follow the instructions in the [root README](../../README.md)**.
+
+2.  Follow the instructions here
+    [https://cloud-dot-devsite.googleplex.com/cdn/docs/signed-urls#creatingkeys](https://cloud-dot-devsite.googleplex.com/cdn/docs/signed-urls#creatingkeys)
+    to create the keys that will be used to sign the URLs, upload them and give relevant permissions.
+
+3.  From a Powershell command line, run to see command line arguments.
```
ahmetb: Powershell &rarr; PowerShell

signUrl/README.md
```diff
@@ -0,0 +1,56 @@
...
+    https://www.microsoft.com/net/core) or later.  That means using
+[Visual Studio 2017](
+    https://www.visualstudio.com/), or the command line.
+
+## Build and Run
+
+1.  **Follow the instructions in the [root README](../../README.md)**.
+
+2.  Follow the instructions here
+    [https://cloud-dot-devsite.googleplex.com/cdn/docs/signed-urls#creatingkeys](https://cloud-dot-devsite.googleplex.com/cdn/docs/signed-urls#creatingkeys)
```
ahmetb: do not link to internal urls

signUrl/README.md
```diff
@@ -0,0 +1,56 @@
+# Google Cloud Signing URL Sample
```
ahmetb: other samples seem to be following Pascal case (i.e. `SignUrl`), is it applicable here?

signUrl/SignUrlSample/Program.cs
simonz130: If I remove this assignment, I'll have to change input parameter value in the method, which is bad practice if you don't explicitly declare this as intent by making the method signature require passing parameter by reference:
<b>Base64UrlDecode(ref string arg);</b>

In this case we return byte array, so no point in changing method signature.
Even though strings are immutable in C# and have value type semantics, still it's better not to mutate input parameter.

signUrl/SignUrlSample/Program.cs
simonz130: Fixed, thanks!

signUrl/README.md
simonz130: Fair point. Fixed so that now we support both short and long forms.

signUrl/README.md
simonz130: ok

signUrl/README.md
simonz130: ok

signUrl/README.md
simonz130: Good catch. Thanks!

signUrl/README.md
simonz130: It's just the folder name, where it always starts with small letter. The rest is following the same convension.

ahmetb: actually I think this should be in `cdn/signedUrl` directory, not `signedUrl`.

signUrl/SignUrlSample/Program.cs
ahmetb: Maybe I'm recalling this wrong about C#, but here's an experiment: https://dotnetfiddle.net/o9EV5F

When you pass a string to a function, and modify its value, its value in the outer scope remains the same.

cdn/signUrl/SignUrlSample/Program.cs
```diff
@@ -0,0 +1,119 @@
...
+
+        private static long ToUnixTime(DateTime date)
+        {
+            var epoch = new DateTime(1970, 1, 1, 0, 0, 0, DateTimeKind.Utc);
+            return Convert.ToInt64((date - epoch).TotalSeconds);
+        }
+
+        private static byte[] Base64UrlDecode(string arg)
+        {
+            string s = arg;
```
ahmetb: (in case my other comment gets lost through changes)

I created a little experiment here and this variable declaration here does not seem necessary https://dotnetfiddle.net/o9EV5F,  I may be wrong but I think I'm right.

cdn/signUrl/SignUrlSample/Program.cs
```diff
@@ -0,0 +1,119 @@
...
+            s = s.Replace('-', '+'); // 62nd char of encoding
+            s = s.Replace('_', '/'); // 63rd char of encoding
+
+            return Convert.FromBase64String(s); // Standard base64 decoder
+        }
+
+        private static string Base64UrlEncode(byte[] inputBytes)
+        {
+            var output = Convert.ToBase64String(inputBytes);
```
ahmetb: not sure why this entire method cannot be

```
return Convert.ToBase64String(inputBytes).Replace('+', '-')
                           .Replace('/', '_');
```

cdn/signUrl/SignUrlSample/Program.cs
```diff
@@ -0,0 +1,119 @@
...
+
+            output = output.Replace('+', '-')      // 62nd char of encoding
+                           .Replace('/', '_');     // 63rd char of encoding
+
+            return output;
+        }
+
+        private static byte[] ComputeHash(byte[] secretKey, string signatureString)
+        {
+            var enc = Encoding.ASCII;
```
ahmetb: used only once hardcode

cdn/signUrl/SignUrlSample/Program.cs
```diff
@@ -0,0 +1,119 @@
...
+            using (HMACSHA1 hmac = new HMACSHA1(secretKey))
+            {
+                hmac.Initialize();
+
+                byte[] buffer = enc.GetBytes(signatureString);
+
+                byte[] hashValue = hmac.ComputeHash(buffer);
```
ahmetb: unnecessary declaration, just return

cdn/signUrl/SignUrlSample/Program.cs
```diff
@@ -0,0 +1,119 @@
...
+        }
+
+        private static byte[] ComputeHash(byte[] secretKey, string signatureString)
+        {
+            var enc = Encoding.ASCII;
+            using (HMACSHA1 hmac = new HMACSHA1(secretKey))
+            {
+                hmac.Initialize();
+
+                byte[] buffer = enc.GetBytes(signatureString);
```
ahmetb: also probably unnecessary declaration, you can put this inside `ComputeHash(...)`

simonz130:
Re: When you pass a string to a function, and modify its value, its value
in the outer scope remains the same.

By default, parameters are passed by value in C# - unless you use "out" or
"ref" parameters. string is a reference type but it's immutable type, so
every change you make to original string will create a new string with
changed value, point existing reference to that string and let GC discard
the old value. So when passing a string as input to a method, you pass a
copy of a reference to that string (or you can say that the reference is
being passed by value). In other words, when passing a string to a method,
you pass a copy of a reference to that string so when you later on do
re-assignment to that variable, you in fact re-reference to a new place in
memory holding a new value but you are not changing original string. This
what explains your experiment. Even though technically, I can use the same
variable passed to the method to make necessary modifications, it's
considered a bad practice to do so - unless your method is explicitly
designed to mutate input parameters. This is not the case in my method, so
I prefer to stick with best practices here.

ahmetb:
Understood. My problem with the current code is that some of the declarations seem unnecessary as we're often just processing and returning a string, the whole method can be a single `return` stmt, or even can be inlined in `SignUrl()` method.

I'll leave it to you and the language experts on this repo, to decide on code cleanliness. I think our Style/Readability Guide probably offers a solution for these sort of "preference" matters. I recommend consulting to the Style Guide.

cdn/signUrl/README.md
```diff
@@ -0,0 +1,55 @@
+# Google Cloud Signing URL Sample
+
+This sample shows how to create Cloud CDN signed Url.
```
SurferJeffAtGoogle: Many readers will not know what CDN means.  Define it here.  Rename the directory something longer than cdn/.  Add a link to the documentation page.

cdn/signUrl/SignUrlSample.Tests/SignUrlTests.cs
```diff
@@ -0,0 +1,61 @@
+// Copyright 2018 Google Inc. All rights reserved.
```
SurferJeffAtGoogle: Copy and paste the exact text we use in all our other copyright statements.

cdn/signUrl/SignUrlSample.Tests/SignUrlTests.cs
```diff
@@ -0,0 +1,61 @@
...
+
+    /// <summary>
+    /// Tests signing Urls
+    /// </summary>
+    public class SignUrlTests
+    {
+        /// <summary>
+        /// Key used to sign all test Urls
+        /// </summary>
+        private static readonly string s_signingKey = "nZtRohdNF9m3cKM24IcK4w==";
```
SurferJeffAtGoogle: I guess this is why this PR is DO NOT MERGE?  What damage can be done with a leaked signing key?

cdn/signUrl/SignUrlSample/Program.cs
```diff
@@ -0,0 +1,128 @@
+﻿// Copyright 2018 Google Inc. All rights reserved.
```
SurferJeffAtGoogle: Copyright ditto.

cdn/signUrl/SignUrlSample/Program.cs
```diff
@@ -0,0 +1,128 @@
...
+            if (string.IsNullOrEmpty(signedUrl))
+            {
+                return 1;
+            }
+
+            Console.WriteLine($"Signed URL: {signedUrl}");
+            return 0;
+        }
+
+        // [START CreateSignedUrl]
```
SurferJeffAtGoogle: On which web page will this code appear?

cdn/signUrl/SignUrlSample/Program.cs
```diff
@@ -0,0 +1,128 @@
...
+        /// <param name="expirationTime">Expiration time of the signature</param>
+        /// <returns>Signed Url that is valid until {expirationTime}</returns>
+        public static string CreateSignedUrl(string url, string keyName, string encodedKey, DateTime expirationTime)
+        {
+            var builder = new UriBuilder(url);
+
+            long unixTimestampExpiration = ToUnixTime(expirationTime);
+
+            char queryParam = string.IsNullOrEmpty(builder.Query) ? '?' : '&';
+            builder.Query += $"{queryParam}Expires={unixTimestampExpiration}&KeyName={keyName}".ToString();
```
SurferJeffAtGoogle: Do we need to `Uri.EscapeDataString()` the `keyName`?

cdn/signUrl/SignUrlSample/Program.cs
SurferJeffAtGoogle: To me, it's 6 vs. half dozen.

cdn/signUrl/SignUrlSample/Program.cs
SurferJeffAtGoogle: +1

cdn/signUrl/SignUrlSample.Tests/SignUrlTests.cs
ahmetb: Nope, the feature isn't launched yet, we do not want to make the sample discoverable too easily.

ahmetb: No damage can be done with the signing key. You need to have a LB that the key is added to.

cdn/signUrl/SignUrlSample/Program.cs
ahmetb: likely `/cdn/docs/signed-urls` (doesn't exist yet)

cdn/signUrl/SignUrlSample/Program.cs
ahmetb: No, we asked the product team and they told the values for this param is url safe.

cdn/signUrl/SignUrlSample/Program.cs
simonz130: @SurferJeffAtGoogle , not sure what do you mean by your comment. Can you explain?

cdn/signUrl/SignUrlSample/Program.cs
simonz130: ok

cdn/signUrl/SignUrlSample/Program.cs
simonz130: I could write the entire method in single line like this:
<b>return new HMACSHA1(secretKey).ComputeHash(Encoding.ASCII.GetBytes(signatureString));</b>
However I don't think it would be good sample this way.
People reading this may have different skill sets and the sample should be easy to follow, debug and understand to all of them. We are not in the competition on the smallest number of lines.
I prefer "unnecessary" declarations for better clarity vs. smaller number of lines of code - especially in sample code.

cdn/signUrl/README.md
simonz130: Agree with Ahmet on the folder, and agree with Jeff that we should add a comment on what CDN means with link. Adding a comment.

cdn/signUrl/SignUrlSample/Program.cs
ahmetb: SGTM

ahmetb:
LGTM

cdn/signUrl/SignUrlSample.Tests/SignUrlTests.cs
simonz130: Done

cdn/signUrl/SignUrlSample.Tests/SignUrlTests.cs
simonz130: It's randomly generated key that you need to upload like Ahmet pointed out. We need a reference point to be able to sign urls in test and compare to baseline, so we have to use some key.

cdn/signUrl/SignUrlSample/Program.cs
simonz130: fixed. Apparently, we have different copyright formats for different languages. Previous was copied from another sample (not dotnet).

ahmetb:
This is now OK to merge. If there are no outstanding comments, can someone merge it?

simonz130:
Yeah, will do

PR #453, opened 2018-02-28
Add Spanner BatchQuery sample.

jsimonweb:


spanner/api/Spanner/Program.cs
```diff
@@ -1343,6 +1378,177 @@ await connection.CreateDdlCommand($"DROP TABLE {table}")
             return ExitCode.Success;
         }
 
+        public static object BatchInsertRecords(string projectId,
+            string instanceId, string databaseId)
+        {
+            var response =
```
simonz130: I think developers can find this confusing.
By Microsoft conventions, when the method is called XXXAsync, it returns a Task or Task<T>. Therefore, if the method is async, you can do <b>var response = await BatchInsertRecordAsync(...);</b> and the expectation is that the method will return the T that Task running the async method returns.
In your case you write the method as synchronous, therefore, it should be <b>var response<i>Task</i> = BatchInsertRecordsAsync...</b>

spanner/api/Spanner/Program.cs
```diff
@@ -1155,12 +1190,12 @@ await connection.CreateDdlCommand($"DROP TABLE {table}")
             }
             // Call method QueryDataWithIndexAsync() based on whether
             // user provided startTitle and endTitle values are empty or null.
-            Task response = string.IsNullOrEmpty(startTitle) ?
-                // startTitle not provided, exclude startTitle and endTitle from
-                // method call to use default values for both.
-                QueryDataWithIndexAsync(projectId, instanceId, databaseId) :
-                // Both startTitle and endTitle provided, include them both in
-                // the method call to QueryDataWithIndexAsync.
+            Task response = string.IsNullOrEmpty(startTitle) ?
```
simonz130: Similar comment to below

spanner/api/Spanner/Program.cs
```diff
@@ -1343,6 +1378,177 @@ await connection.CreateDdlCommand($"DROP TABLE {table}")
...
+                var cmd = connection.CreateSelectCommand(
+                    @"SELECT MAX(SingerId) as SingerId FROM Singers");
+                using (var reader = await cmd.ExecuteReaderAsync())
+                {
+                    while (await reader.ReadAsync())
+                    {
+                        long parsedValue;
+                        bool result = Int64.TryParse(
+                            reader.GetFieldValue<string>("SingerId"), out parsedValue);
+                        if (result) maxSingerId = parsedValue;
```
simonz130: minor:
I suggest not to have one liners like this - it confuses the readers and is prone to bugs. 
Instead, we can do:
if (result)
{
   maxSingerId = parsedValue;
}

spanner/api/Spanner/Program.cs
```diff
@@ -1343,6 +1378,177 @@ await connection.CreateDdlCommand($"DROP TABLE {table}")
...
+            using (var connection = new SpannerConnection(connectionString))
+            {
+                await connection.OpenAsync();
+
+                for (int i = 0; i < 100; i++)
+                {
+                    using (var tx = await connection.BeginTransactionAsync())
+                    using (var cmd = connection.CreateInsertCommand("Singers", new SpannerParameterCollection
+                    {
+                        {"SingerId", SpannerDbType.String},
```
simonz130: nit: please, add one space after { and before }

spanner/api/Spanner/Program.cs
```diff
@@ -1343,6 +1378,177 @@ await connection.CreateDdlCommand($"DROP TABLE {table}")
...
+            }
+
+            // Batch insert 249,900 singer records into the Singers table.
+            using (var connection = new SpannerConnection(connectionString))
+            {
+                await connection.OpenAsync();
+
+                for (int i = 0; i < 100; i++)
+                {
+                    using (var tx = await connection.BeginTransactionAsync())
```
simonz130: Is it possible to specify isolation level for the transaction?
When reading it without knowing what's the default, it makes it tricky to follow. 
If there is no overload that specifies isolation level, worth adding a comment to explain what the default is and what it means.

spanner/api/Spanner/Program.cs
```diff
@@ -1343,6 +1378,177 @@ await connection.CreateDdlCommand($"DROP TABLE {table}")
...
+                    await Task.WhenAll(partitions.Select(
+                            x => DistributedReadWorkerAsync(x, transactionId)))
+                                .ConfigureAwait(false);
+                }
+                Console.WriteLine($"Done reading!  Total rows read: "
+                    + $"{s_rowsRead:N0} with {s_partitionId} partition(s)");
+            }
+        }
+
+        private static int s_partitionId;
```
simonz130: nit: All the static variables should be defined before all the methods

spanner/api/Spanner/Program.cs
```diff
@@ -1343,6 +1378,177 @@ await connection.CreateDdlCommand($"DROP TABLE {table}")
...
+        {
+            var response =
+                DistributedReadAsync(projectId, instanceId, databaseId);
+            Console.WriteLine("Waiting for operation to complete...");
+            response.Wait();
+            Console.WriteLine($"Operation status: {response.Status}");
+            return ExitCode.Success;
+        }
+
+        private static async Task DistributedReadAsync(string projectId,
```
simonz130: Please, document all relevant methods with xmldoc.

spanner/api/Spanner/Program.cs
```diff
@@ -1343,6 +1378,177 @@ await connection.CreateDdlCommand($"DROP TABLE {table}")
...
+        }
+
+        private static async Task DeleteDatabaseAsync(string projectId,
+            string instanceId, string databaseId)
+        {
+            string AdminConnectionString = $"Data Source=projects/{projectId}/"
+                + $"instances/{instanceId}";
+            using (var connection = new SpannerConnection(AdminConnectionString))
+            using (var cmd = connection.CreateDdlCommand(
+                $@"DROP DATABASE {databaseId}"))
```
simonz130: You should document the assumptions of DDL command. Typically you need extra permissions besides R/W on the DB to delete it.

spanner/api/Spanner/Program.cs
```diff
@@ -1343,6 +1378,177 @@ await connection.CreateDdlCommand($"DROP TABLE {table}")
...
+                    connection.CreateSelectCommand(
+                        "SELECT SingerId, FirstName, LastName FROM Singers"))
+                {
+                    transaction.DisposeBehavior =
+                        DisposeBehavior.CloseResources;
+                    cmd.Transaction = transaction;
+                    var partitions = await cmd.GetReaderPartitionsAsync();
+                    var transactionId = transaction.TransactionId;
+                    await Task.WhenAll(partitions.Select(
+                            x => DistributedReadWorkerAsync(x, transactionId)))
```
simonz130: Here you call a method the number of times as the number of partitions. Each method call will create a separate connection and transaction within outer connection and transaction. Is that really what you want?  

Also, the static variables - do you really need them? Can you just define them in the calling method and pass them by reference to the method that is incrementing their value? This way you don't risk the possibility that someone will use them in a wrong way.

spanner/api/SpannerTest/Tests.cs
```diff
@@ -53,6 +57,17 @@ public class SpannerTests
             Command = "Spanner"
         };
 
+        public void Dispose()
+        {
+            try
+            {
+                // Delete database created from running the tests.
+                _spannerCmd.Run("deleteSampleDatabase",
```
simonz130: What happens with the database if this fails?

spanner/api/SpannerTest/runTests.ps1
```diff
@@ -14,11 +14,12 @@
...
+Require-Platform Win*
 
 BackupAndEdit-TextFile "..\QuickStart\Program.cs" `
     @{"YOUR-PROJECT-ID" = $env:GOOGLE_PROJECT_ID} `
-{ 
+{
     dotnet restore
-    dotnet test
+    dotnet build
+    dotnet test --no-build --no-restore -v detailed
```
simonz130: This is great! Thanks for adding this optimization!

spanner/api/Spanner/Program.cs
jsimonweb: Added a comment to point to the transaction isolation documentation.

spanner/api/Spanner/Program.cs
jsimonweb: Added a comment to note the IAM Database Admin role is required.

spanner/api/Spanner/Program.cs
jsimonweb: Yes we want the partition reads to run in parallel controlled by the calling transaction.

The static variables simplify the method signature for inclusion in documentation. I'd prefer to leave as is.

spanner/api/SpannerTest/Tests.cs
jsimonweb: It will be deleted as a beginning step the next time the test is run.
https://github.com/GoogleCloudPlatform/dotnet-docs-samples/pull/453/files#diff-a502c0e97c603f710fbe44afa5200f6aR86

spanner/api/Spanner/Program.cs
jsimonweb: I'd prefer to leave this one as is since we've got adequate comments included to describe how the logic is being processed.

spanner/api/Spanner/Program.cs
jsimonweb: Done.

spanner/api/Spanner/Program.cs
jsimonweb: Done.

spanner/api/Spanner/Program.cs
jsimonweb: Moved the static variables to the top of the code snippet utilizing them so they remain included in the code snippet to which they are relevant.

spanner/api/Spanner/Program.cs
jsimonweb: Opting not to do this to maintain the existing format of the code which does not include xmldoc comments for methods. We can revisit this for future code changes.

spanner/api/Spanner/Program.cs
jsimonweb: Done.

jsimonweb:
Addressed all open review comments. Please re-review.

PR #457, opened 2018-03-05
Corrects code regions on set commands in README

gguuss:


gguuss:
Looks like spanner is failing?

nawaf4:
nnc55

PR #464, opened 2018-03-10
DLPv2 samples

simonz130:
DLPv2 samples.

Currently includes: 

Dane's samples: inspect string and file
Simon's samples: templates


dlp/api/DlpSample/InspectTemplates.cs
```diff
@@ -0,0 +1,117 @@
...
+        {
+            DlpServiceClient client = DlpServiceClient.Create();
+
+            var request = new DeleteInspectTemplateRequest
+            {
+                Name = templateName
+            };
+
+            client.DeleteInspectTemplate(request);
+            Console.WriteLine($"Template {templateName} was deleted");
```
dzlier-gcp: Do we want to verify anything about the response before claiming the template was deleted? Or does the DeleteInspectTemplate method return void?

dlp/api/DlpSample/InspectTemplates.cs
```diff
@@ -0,0 +1,117 @@
...
+
+                currentTemplatesPage = response.ReadPage(batchSize); ;
+                foreach (var template in currentTemplatesPage)
+                {
+                    Console.WriteLine("Inspect Template Info:");
+                    Console.WriteLine($"\tname: {template.Name}");
+                    Console.WriteLine($"\tdisplayName: {template.DisplayName}");
+                    Console.WriteLine($"\tdescription: {template.Description}");
+                    Console.WriteLine($"\tcreateTime: {template.CreateTime}");
+                    Console.WriteLine("Configuration:");
```
dzlier-gcp: Should this and the remaining WriteLines below also start with a \t, or did you want the indent only for the above info?

My preference would be to keep the indent all the way through, otherwise in the output there won't be much separation between the templates. At least adding a newline between the templates would be good.

dlp/api/DlpSample/InspectTemplates.cs
```diff
@@ -0,0 +1,117 @@
...
+                    if (template.InspectConfig.InfoTypes.Any())
+                    {
+                        Console.WriteLine($"Info types: {string.Join(',', template.InspectConfig.InfoTypes.Select(t => t.Name))}");
```
dzlier-gcp: Please go through and break up long lines. I've been using a line length of 120 for mine.

dlp/api/DlpSample/Program.cs
```diff
@@ -0,0 +1,116 @@
...
+    }
+
+    [Verb("inspectFile", HelpText = "Inspects a content file.")]
+    class InspectFileOptions : InspectLocalOptions
+    {
+        [Value(1, HelpText = "The path to the local file to inspect. Can be a text, JPG, or PNG.", Required = true)]
+        public string File { get; set; }
+    }
+
+    [Verb("createInspectTemplate", HelpText = "Creates a template for inspecting operations")]
```
dzlier-gcp: Nit: end HelpText with period.

dlp/api/DlpSample/Program.cs
```diff
@@ -0,0 +1,116 @@
...
+    class InspectTemplateOptions : InspectLocalOptions
+    {
+        [Value(1, HelpText = "Display name of the template", Required = true)]
```
dzlier-gcp: Nit: end HelpText with period.

dlp/api/DlpSample/Program.cs
```diff
@@ -0,0 +1,116 @@
...
+        public string DisplayName { get; set; }
+
+        [Value(2, HelpText = "Description of the template", Required = true)]
```
dzlier-gcp: Nit: end HelpText with period.

dlp/api/DlpSample/Program.cs
```diff
@@ -0,0 +1,116 @@
...
+        public string Description { get; set; }
+    }
+
+    [Verb("listTemplates", HelpText = "Lists all created inspection templates")]
```
dzlier-gcp: Nit: end HelpText with period.

dlp/api/DlpSample/Program.cs
```diff
@@ -0,0 +1,116 @@
...
+
+    [Verb("inspectFile", HelpText = "Inspects a content file.")]
+    class InspectFileOptions : InspectLocalOptions
+    {
+        [Value(1, HelpText = "The path to the local file to inspect. Can be a text, JPG, or PNG.", Required = true)]
+        public string File { get; set; }
+    }
+
+    [Verb("createInspectTemplate", HelpText = "Creates a template for inspecting operations")]
+    class InspectTemplateOptions : InspectLocalOptions
```
dzlier-gcp: This is only used for Create, so the name should match the pattern "CreateTemplateOptions"

dlp/api/DlpSample/Program.cs
```diff
@@ -0,0 +1,116 @@
...
+    }
+
+    [Verb("listTemplates", HelpText = "Lists all created inspection templates")]
+    class ListTemplatesOptions
+    {
+        [Value(0, HelpText = "The project ID to run the API call under.", Required = true)]
+        public string ProjectId { get; set; }
+    }
+
+    [Verb("deleteTemplate", HelpText = "Deletes given template by name")]
```
dzlier-gcp: Nit: end HelpText with period.

dlp/api/DlpSample/Program.cs
```diff
@@ -0,0 +1,116 @@
...
+    class DeleteTemplatesOptions
+    {
+        [Value(0, HelpText = "The project ID to run the API call under.", Required = true)]
+        public string ProjectId { get; set; }
+
+        [Value(1, HelpText = "The name of the template to delete", Required = true)]
```
dzlier-gcp: Nit: end HelpText with period.

dlp/api/DlpTest/DlpTest.cs
```diff
@@ -0,0 +1,144 @@
...
+            AssertPhoneEmailCC(_dlp.Run(verb, _projectId, value, "-m", "2"), checkPhone: true, checkEmail: true);
+            AssertPhoneEmailCC(_dlp.Run(verb, _projectId, value, "-m", "1"), checkPhone: true);
+        }
+
+        [Fact]
+        public void TestTemplates()
+        {
+            // Creation
+            ConsoleOutput output = _dlp.Run("createInspectTemplate", _projectId, "testDisplayName", "test description");
+            Assert.Contains("name: ", output.Stdout);
```
dzlier-gcp: Does the name for the created template not always match the input name?

dlp/api/DlpSample/InspectTemplates.cs
simonz130: Good point. It's not possible to do any verification as the method the library is exposing is void. I'll create an issue for library author to change the signature to return the id of deleted template

dlp/api/DlpSample/InspectTemplates.cs
simonz130: Good points. Added a new line and added tabs for sub-configurations.

dlp/api/DlpSample/InspectTemplates.cs
simonz130: Done

dlp/api/DlpSample/Program.cs
simonz130: done

dlp/api/DlpSample/Program.cs
simonz130: done

dlp/api/DlpSample/Program.cs
simonz130: done

dlp/api/DlpSample/Program.cs
simonz130: ok

dlp/api/DlpSample/Program.cs
simonz130: ok

dlp/api/DlpSample/Program.cs
simonz130: ok

dlp/api/DlpTest/DlpTest.cs
simonz130: When creating a template, you pass a "display name", and the actual name is generated by the BE.
Honestly, I don't like these tests. We are testing CLI layer instead of the actual API. However I just followed established pattern. We should have a separate discussion on whether we need CLI layer at all, and regardless, how do we test the sample code. Both are outside of this PR.

ace-n: FWIW - the Node samples test the CLI to verify that the CLI and API don't diverge over time. If you want to drop the CLI itself, that's certainly an option (though it may make it harder for our team's tools to interact with your samples) - but if we're keeping it for now, I'm personally in favor of keeping tests as "end-to-end" as possible.

dlp/api/DlpSample/InspectTemplates.cs
```diff
@@ -0,0 +1,119 @@
...
+    {
+        // [START dlp_create_inspect_template]
+        public static string CreateInspectTemplate(
+            string projectId, 
+            string displayName,
+            string description,
+            int likelihood, 
+            int maxFindings,
+            bool includeQuote,
+            string infoTypes)
```
dzlier-gcp: This never appears to get used. You can parse them into an actual List by extending DlpSampleBase and calling ParseInfoTypes(infoTypes)

This also tells me that your unit test doesn't check the infoTypes param on the call.

PR #474, opened 2018-03-15
Sponge looks for a log file called sponge_log.xml specifically.

SurferJeffAtGoogle:
So rename our logs accordingly.

### DIFF
--- a/logging/api/LoggingTest/LoggingTest.csproj
+++ b/logging/api/LoggingTest/LoggingTest.csproj
@@ -8,10 +8,10 @@
 
   <ItemGroup>
     <PackageReference Include="Google.Cloud.Logging.V2" Version="2.1.0" />
+    <PackageReference Include="JUnitTestLogger" Version="0.6.0" />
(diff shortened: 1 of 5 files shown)


logging/api/LoggingTest/LoggingTest.csproj
```diff
@@ -8,10 +8,10 @@
 
   <ItemGroup>
     <PackageReference Include="Google.Cloud.Logging.V2" Version="2.1.0" />
+    <PackageReference Include="JUnitTestLogger" Version="0.6.0" />
```
simonz130: What is the difference between XUnit and JUnit? Is it what enables sponge to capture logs correctly?

SurferJeffAtGoogle: Sponge expects https://en.wikipedia.org/wiki/XUnit, more specifically http://nose.readthedocs.io/en/latest/plugins/xunit.html.  JUnit output conforms to that specification.

It's a horribly overloaded name.  Anyway, the xunit output format is actually a csv by default.

PR #494, opened 2018-04-03
Setup script2

SurferJeffAtGoogle:


.kokoro-windows/New-BuildEnv.ps1
```diff
@@ -0,0 +1,86 @@
...
+    [System.IO.Compression.ZipFile]::ExtractToDirectory($zipfile, $outpath)
+}
+
+Set-PsDebug -Trace 1
+# Install codeformatter
+Unzip $env:KOKORO_GFILE_DIR\codeformatter.zip $installDir\codeformatter
+# Install phantomjs
+Unzip $env:KOKORO_GFILE_DIR\phantomjs-2.1.1-windows.zip $installDir
+# Install casperjs
+Unzip $env:KOKORO_GFILE_DIR\n1k0-casperjs-1.0.3-0-g76fc831.zip $installDir
```
simonz130: It could be really nice if we could deal with the versions of all the dependencies in one place instead of referencing the exact versions in multiple places. This will make updating to newest versions a lot easier.

.kokoro-windows/New-BuildEnv.ps1
```diff
@@ -0,0 +1,86 @@
...
+# Patch casperjs
+Copy-Item -Force $PSScriptRoot\..\.kokoro\docker\bootstrap.js `
+    $installDir\n1k0-casperjs-76fc831\bin\bootstrap.js
+# Install casperjs 1.1
+Unzip $env:KOKORO_GFILE_DIR\casperjs-1.1.4-1.zip $installDir
+Set-PsDebug -Off
+
+# Copy Activate.ps1 to the environment directory.
```
simonz130: IMO, better name would be install or initialize.

SurferJeffAtGoogle: I'm following the pattern of virtualenv described here:
https://virtualenv.pypa.io/en/stable/userguide/
It generates a file called activate.

.kokoro-windows/New-BuildEnv.ps1
SurferJeffAtGoogle: That would be ideal, but the motivating reason for doing it like this is that the new scripts, New-BuildEnv and Activate, must be run in different contexts.  New-BuildEnv must be run as an administrator.  Activate should definitely not be run as an administrator.  Maybe it could be defined in a common include file?

SurferJeffAtGoogle:
Thank you for the review, Simon!

PR #507, opened 2018-04-09
Firestore C# samples

ryanmats:
Ready for review!

### DIFF
--- a/firestore/api/FirestoreTest/FirestoreTest.cs
+++ b/firestore/api/FirestoreTest/FirestoreTest.cs
@@ -1,0 +1,621 @@
+/*
+ * Copyright (c) 2018 Google Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License"); you may not
+ * use this file except in compliance with the License. You may obtain a copy of
+ * the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS, WITHOUT
+ * WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied. See the
+ * License for the specific language governing permissions and limitations under
+ * the License.
+ */
+
+using System;
+using System.Collections;
+using System.Collections.Generic;
+using System.IO;
+using System.Text.RegularExpressions;
+using System.Threading.Tasks;
+using Xunit;
+using Google.Cloud.Firestore;
+
+
+namespace GoogleCloudSamples
+{
+    public class FirestoreTests
+    {
+        readonly CommandLineRunner _quickstart = new CommandLineRunner()
+        {
+            VoidMain = Quickstart.Main,
+            Command = "dotnet run"
+        };
+
+        protected ConsoleOutput RunQuickstart(params string[] args)
+        {
+            return _quickstart.Run(args);
+        }
+
+        readonly CommandLineRunner _addData = new CommandLineRunner()
+        {
+            VoidMain = AddData.Main,
+            Command = "dotnet run"
+        };
+
+        protected ConsoleOutput RunAddData(params string[] args)
+        {
+            return _addData.Run(args);
+        }
+
+        readonly CommandLineRunner _deleteData = new CommandLineRunner()
+        {
+            VoidMain = DeleteData.Main,
+            Command = "dotnet run"
+        };
+
+        protected ConsoleOutput RunDeleteData(params string[] args)
+        {
+            return _deleteData.Run(args);
+        }
+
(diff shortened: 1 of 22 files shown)


firestore/api/FirestoreTest/FirestoreTest.cs
```diff
@@ -0,0 +1,622 @@
...
+            var output = RunQuickstart("initialize-project-id", Environment.GetEnvironmentVariable("FIRESTORE_PROJECT_ID"));
+            Assert.Contains("Created Cloud Firestore client with project ID:", output.Stdout);
+        }
+
+        [Fact]
+        public void AddData1Test()
+        {
+            var output = RunQuickstart("add-data-1", Environment.GetEnvironmentVariable("FIRESTORE_PROJECT_ID"));
+            Assert.Contains("Added data to the alovelace document in the users collection.", output.Stdout);
+            DeleteCollection("users").Wait();
```
SurferJeffAtGoogle: The clean up operation needs to happen reliably even when an Assertion fails.  Make this class inherit from IDisposable and put the clean up code in the Dispose() method.

firestore/api/FirestoreTest/FirestoreTest.cs
```diff
@@ -0,0 +1,622 @@
...
+        }
+
+        [Fact]
+        public void AddData2Test()
+        {
+            var output = RunQuickstart("add-data-2", Environment.GetEnvironmentVariable("FIRESTORE_PROJECT_ID"));
+            Assert.Contains("Added data to the aturing document in the users collection.", output.Stdout);
+            DeleteCollection("users").Wait();
```
SurferJeffAtGoogle: To avoid name collisions when two instances of this test are running at the same time, we need some randomization in the names.

For datastore, we used a randomly named namespace.  Not sure what the corresponding solution will be for firestore.  Names should clearly indicate that it's part of a test.  Something like "kokoro-test-users-<randomstring>".

ryanmats: Is this for when Kokoro CI Windows + Kokoro CI Linux tests might be running at the same time?

For Firestore there isn't an equivalent of Datastore's namespaces (https://stackoverflow.com/questions/48570122/what-is-the-namespace-string-of-a-document-in-cloud-firestore-for-purposes-of).

If we try to change the collection names (i/e users-123 and users-124) we would run into a problem of not having proper indexes set up for each collection. While there aren't client library methods to add indexes, it does look like we can create them from the REST API (https://cloud.google.com/firestore/docs/reference/rest/v1beta1/projects.databases.indexes/create). However, index creation can take several minutes and would thus slow down the tests significantly.

Another option is to create different projects for the Windows and Linux tests.

Thoughts?

SurferJeffAtGoogle: Wow, then Firebase is missing a key feature.

I read the instructions in the README for creating the indexes, and it looks pretty tedious and error prone.  I think we'd be doing the user a favor if we added a command to create the indexes.  Because each test would start with empty collections, I would be very surprised if index creation actually took minutes.

ryanmats: So I was able to get programmatic index creation working!

My current plan (using testutil to generate random collection names for each test (i/e users-12345)) will require me to change the sample code functions to reflect the collection name changes. This could be somewhat problematic from the documentation point of view. The other languages' samples all use specific collection names (i/e "users" or "cities") for each sample - see Firestore Add Data here (https://cloud.google.com/firestore/docs/quickstart#add_data).

With randomly generated collection names, I'll have to change all the snippets to use a variable like 'usersCollectionId'. So the code would change from https://gist.github.com/ryanmats/66ea48d6838bcb587d5eba27a731a474 to https://gist.github.com/ryanmats/84744aa4210d8317042ab24586e7ada4. This would make the C# samples different from the other languages, which is not ideal.

I can change my samples to this format, but are there any other options?

@SurferJeffAtGoogle @jsimonweb

SurferJeffAtGoogle: Yes.  We modify runTests.ps1 so that it updates the .cs source and replaces names with random names before building and running the test.  We already have the utility function `BackupAndEdit-TextFile ` in BuildTools.psm1 to make this easier.  For an example, see https://github.com/GoogleCloudPlatform/dotnet-docs-samples/blob/master/appengine/flexible/CloudSql/runTestsWithMySql.ps1.

ryanmats:
The tests now work end-to-end with randomly generated collection names + indexes created for them programmatically.

Ready for review.

ryanmats:
thanks @jsimonweb for recommendation to use a Fixture for cleanup. All tests working now locally.

PR #509, opened 2018-04-12
Migrate CommandLineUtil's tests to xunit.

SurferJeffAtGoogle:
So we see fine-grained results in Kokoro build output.


SurferJeffAtGoogle:
> Why do we need this project and its tests?

Because the [CommandLineParser](https://www.nuget.org/packages/CommandLineParser/) library is inconvenient in two ways:
1. It can't handle more than 16 verbs.  Some of our samples have more than 16 verbs.
2. Its generic-heavy interface requires you to repeat yourself, and is error prone.  The fluent interface provided by this library never requires you to repeat yourself, and is much less error prone.

Issue #510, opened 2018-04-12
DLP tests do not run on Kokoro.


SurferJeffAtGoogle:

dzlier-gcp:
Some of the failing tests are due to [this commit](https://github.com/GoogleCloudPlatform/dotnet-docs-samples/commit/df95d5d756151cf90ca760c6d211927534871ea9#diff-5f6de6a82ea3730361c8537a2acd33db), which removed the setup code that ensures that a KMS keyring/key exists for the tests and then uses it to create a wrapped key (cipher file). @ace-n , please re-add the code so that these tests pass again.

pheonixblade9:
This is still an issue. Please fix.

amanda-tarafa:
@annadata assigning to you so you can triage. This has been open for a long time but there was no one to pick it up.
The failing tests are:
- https://github.com/GoogleCloudPlatform/dotnet-docs-samples/blob/main/dlp/api/Snippets.Tests/DeidentifyWithDateShiftTests.cs
- https://github.com/GoogleCloudPlatform/dotnet-docs-samples/blob/main/dlp/api/Snippets.Tests/DeidentifyWithFpeTests.cs

amanda-tarafa:
Closed by #2202

PR #511, opened 2018-04-12
Run some DLP tests.

SurferJeffAtGoogle:
But not yet all.

https://github.com/GoogleCloudPlatform/dotnet-docs-samples/issues/510


dlp/api/DlpTest/DlpTest.cs
```diff
@@ -135,7 +135,7 @@ public void TestDeidMask()
             Assert.Contains("My SSN is *****9127", output.Stdout);
         }
 
-        [Fact]
+        [Fact(Skip = "https://github.com/GoogleCloudPlatform/dotnet-docs-samples/issues/510")]
```
simonz130: Please, re-assign it to me.

PR #512, opened 2018-04-12
Endpoints test xunit

SurferJeffAtGoogle:


endpoints/getting-started/src/IO.SwaggerTest/UnitTest1.cs
```diff
@@ -0,0 +1,51 @@
...
+using System.Collections.Generic;
+using System.Net;
+using System.Net.Http;
+using Xunit;
+using Newtonsoft.Json;
+using System.IO;
+
+namespace IO.SwaggerTest
+{
+    public class UnitTest1
```
simonz130: Can you rename it to something more meaningful? File name as well.

endpoints/getting-started/src/IO.SwaggerTest/UnitTest1.cs
```diff
@@ -0,0 +1,51 @@
...
+    {
+        class Payload
+        {
+            public string Message { get; set; }
+        }
+
+        [Fact]
+        public async void TestEcho()
+        {
+            var request = HttpWebRequest.Create("http://localhost:7412/echo");
```
simonz130: Can you read this from configuration?

simonz130: We should use HttpClient to make http requests in .NET Core code. HttpWebRequest is old, outdated API.

endpoints/getting-started/src/runTests.ps1
```diff
@@ -12,18 +12,18 @@
...
-$url = "http://localhost:7412"
-$job = Run-Kestrel($url)
-Start-Sleep -Seconds 5
+Import-Module -DisableNameChecking ..\..\..\BuildTools.psm1
+Set-PSDebug -Trace 1
 try {
-	.\Test.ps1 "$url/echo"
+	Push-Location
+	Set-Location IO.Swagger
+	$url = "http://localhost:7412"
```
simonz130: Can we have this and the http client request in tests read this value from the same location?

BuildTools.psm1
```diff
@@ -770,31 +789,26 @@ function Move-TestResults($OutDir) {
...
+        $LASTEXITCODE = 0      
+        foreach ($suite in $x.testsuites.testsuite) {
+            $LASTEXITCODE += [int] $suite.failures 
         }
-        $sleepSeconds *= 2
+    } else {
+        $casperOut = casperjs $TestJs $Url
+    }
+    if ($LASTEXITCODE -eq 0) {
+        $casperOut | Write-Host
```
simonz130: Is Casper output tagged somehow so it's easy to recognize that it's casper output?

endpoints/getting-started/src/IO.SwaggerTest/UnitTest1.cs
```diff
@@ -0,0 +1,51 @@
...
+            request.Method = "POST";
+            request.ContentType = "application/json";
+            using (var streamWriter = new StreamWriter(request.GetRequestStream()))
+            {
+                streamWriter.Write(JsonConvert.SerializeObject(
+                    new Payload { Message = "in a bottle" }));
```
simonz130: nit: create a const string at the start of the test, then use it here and in assert later on.

SurferJeffAtGoogle: Done.

BuildTools.psm1
SurferJeffAtGoogle: Casper also creates a TestResults.xml, which gets rendered in TestFusion and Sponge.

endpoints/getting-started/src/runTests.ps1
SurferJeffAtGoogle: Done.

endpoints/getting-started/src/IO.SwaggerTest/UnitTest1.cs
SurferJeffAtGoogle: Done.

SurferJeffAtGoogle: Done.

endpoints/getting-started/src/IO.SwaggerTest/UnitTest1.cs
SurferJeffAtGoogle: Done.

SurferJeffAtGoogle:
> Do we really need casper and kestrel to test what we test? Can't we just leverage the fact it's ASP.NET Core OWIN and create integration test with the same effect?
> https://docs.microsoft.com/en-us/aspnet/core/testing/integration-testing?view=aspnetcore-2.1

For this test, casper is not needed and not used.

For other tests, we need to do much more than just fetch pages and check http status codes.  We need to click on links in pages.  We need to run the javascript and observe changes to page elements.  ASP.NET's test server does not appear to be able to do that.

BuildTools.psm1
simonz130: I probably didn't structure the question well. 
I meant if there is any way to distinguish between the traces that are created by casper and the traces that are created by other parts (like dotnet test, ASP.NET service traces, etc)? Assuming that all the traces will be captured in testResults.xml

Also, not relevant for this code review, but in general, in my opinion, it's "dotnet test" that should produce TestResults.xml via custom logger implementation we should supply it. Casper traces should be captured by the test harness (dotnet test) via custom logger.

PR #516, opened 2018-04-16
Fix ListJobs test by first creating a job to list.

dzlier-gcp:


SurferJeffAtGoogle:
Run Format-code in the dlp directory to fix the lint issues.

dzlier-gcp:
Done.

dzlier-gcp:
Not sure what is broken about the tests now. There isn't a clear failure message. Looks like one test might have just timed out? How do you trigger a test re-run?

dlp/api/DlpTest/DlpTest.cs
```diff
@@ -305,19 +308,47 @@ public void TestKMap()
...
-            string jobName = dlpJobRegex.Match(listOutput.Stdout).Value;
-            ConsoleOutput deleteOutput = _dlp.Run("deleteJob", jobName);
-            Assert.Contains($"Successfully deleted job {jobName}", deleteOutput.Stdout);
+            _retryRobot.Eventually(() =>
+            {
+                // List jobs.
+                ConsoleOutput listOutput = _dlp.Run("listJobs", CallingProjectId, "state=DONE", "RiskAnalysisJob");
+                Assert.Matches(dlpJobRegex, listOutput.Stdout);
+
+                // Delete created job.
```
SurferJeffAtGoogle: Clean up code should go in a Dispose() method of this class.  Otherwise, if line 345's assertion fails, the code will leak resources.

dzlier-gcp: I'll add a dispose method that deletes any DLP jobs found, but we still want this delete check here because it's actually testing the `deleteJob` command.

dzlier-gcp:
Failed test is video timing out again.

PR #529, opened 2018-04-25
Add Spanner Leaderboard sample.

jsimonweb:


applications/leaderboard/README.md
```diff
@@ -0,0 +1,68 @@
+# .NET Cloud Spanner Sample Leaderboard Application
```
SurferJeffAtGoogle: Could you add a brief note about why there are step1, step2, etc. directories?

applications/leaderboard/Leaderboard/Program.cs
```diff
@@ -0,0 +1,525 @@
...
+                $"Data Source=projects/{projectId}/instances/{instanceId}"
+                + $"/databases/{databaseId}";
+
+            // Get current max PlayerId.
+            Int64 maxPlayerId = 0;
+            using (var connection = new SpannerConnection(connectionString))
+            {
+                // Execute a SQL statement to get current MAX() of PlayerId.
+                var cmd = connection.CreateSelectCommand(
+                    @"SELECT MAX(PlayerId) as PlayerId FROM Players");
```
SurferJeffAtGoogle: Is this really the recommended way in spanner?  In other databases, I know this pattern creates a bottleneck on the playerId.  In SqlServer, for example, I would expect this value to be a GUID.

applications/leaderboard/Leaderboard/Program.cs
```diff
@@ -0,0 +1,525 @@
...
+                    }
+                }
+            }
+
+            // Insert 100 player records into the Players table.
+            using (var connection = new SpannerConnection(connectionString))
+            {
+                await connection.OpenAsync();
+                using (var tx = await connection.BeginTransactionAsync())
+                using (var cmd = connection.CreateInsertCommand("Players", new SpannerParameterCollection
```
SurferJeffAtGoogle: To avoid collisions, getting the max playerid would have to happen in the same transaction.  But as my comment says above, I'd expect to see a GUID or the like.

applications/leaderboard/Leaderboard/Program.cs
```diff
@@ -0,0 +1,525 @@
...
+        public static object QueryTopTenOfYear(string projectId,
+            string instanceId, string databaseId)
+        {
+            var response = QueryTopTenOfYearAsync(
+                projectId, instanceId, databaseId);
+            response.Wait();
+            return ExitCode.Success;
+        }
+
+        public static async Task QueryTopTenOfYearAsync(
```
SurferJeffAtGoogle: OfYear, OfWeek, OfMonth....  I think these 3 functions can be written as one function with one more parameter, the time span.

applications/leaderboard/LeaderboardTest/Tests.cs
```diff
@@ -0,0 +1,120 @@
...
+using System.Threading;
+using System.Threading.Tasks;
+using System.Text.RegularExpressions;
+using Xunit;
+using Xunit.Sdk;
+using Grpc.Core;
+
+namespace GoogleCloudSamples.Leaderboard
+{
+    public class LeaderboardFixture : IDisposable
```
SurferJeffAtGoogle: Nicely done!

applications/leaderboard/README.md
jsimonweb: Done.

PR #536, opened 2018-05-07
Throttle dialogflow calls.

SurferJeffAtGoogle:
Also, RetryRobot makes dialogflow tests cope with eventual consistency in the list() method.

### DIFF
--- a/dialogflow/api/Test/DialogflowTest.cs
+++ b/dialogflow/api/Test/DialogflowTest.cs
@@ -17,11 +17,14 @@ using System.Linq;
 using System.Collections.Generic;
 using System.Text.RegularExpressions;
 using Xunit;
+using System.Collections.Concurrent;
+using System.Threading.Tasks;
 
 namespace GoogleCloudSamples
 {
     public class DialogflowTest
     {
+        protected RetryRobot _retryRobot = new RetryRobot();
         public readonly string ProjectId = Environment.GetEnvironmentVariable("GOOGLE_PROJECT_ID");
         public readonly string SessionId = TestUtil.RandomName();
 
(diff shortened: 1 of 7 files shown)


beccca:
TIL about `BlockingCollection` and it's pretty darned cool!

Took me a couple mins to walk through and grok it. I can't think of any semantics which would've helped me grok it faster (except 1 comment I made about `timeSpan`, feel free to make this change but I'm LGTM!)

dialogflow/api/Test/DialogflowTest.cs
```diff
@@ -72,4 +84,50 @@ public ConsoleOutput RunWithSessionId(string command, params object[] args)
...
+    // TODO: Move this class into test helpers.
+    /// <summary>
+    /// Schedules throttling.
+    /// </summary>
+    class ThrottleTokenPool
+    {
+        private readonly TimeSpan _timeSpan;
+        private readonly BlockingCollection<ThrottleToken> _pool =
+            new BlockingCollection<ThrottleToken>();
+        public ThrottleTokenPool(int tokenCount, TimeSpan timeSpan)
```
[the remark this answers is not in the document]
SurferJeffAtGoogle: Added lots of comments.  Thank you for the review!

dialogflow/api/Test/DialogflowTest.cs
```diff
@@ -72,4 +84,65 @@ public ConsoleOutput RunWithSessionId(string command, params object[] args)
             return Run(command, arguments.ToArray());
         }
     }
+
+    // TODO: Move this class into test helpers.
+    /// <summary>
+    /// Schedules throttling.
+    /// </summary>
+    class ThrottleTokenPool
```
simonz130: Should the pool be static? Can you have multiple instances of throttler pools that in fact block requests to the same actual resource?

simonz130: I see that you create a static instance of the pool earlier, but IMO the class itself should also be static and should expose methods to initialize and get access to the throttle pool per resource type that is being throttled.

Issue #546, opened 2018-05-21
Revert PR 545 after 2018-05-23.


SurferJeffAtGoogle:
https://github.com/GoogleCloudPlatform/dotnet-docs-samples/pull/545

PR #550, opened 2018-05-22
Samples to delete alert policies and notification channels.

SurferJeffAtGoogle:
I needed these tools to clean up policies and channels left behind by tests.

Also, fix https://github.com/GoogleCloudPlatform/dotnet-docs-samples/issues/546

### DIFF
--- a/monitoring/api/AlertSample/Program.cs
+++ b/monitoring/api/AlertSample/Program.cs
@@ -136,6 +136,7 @@ namespace GoogleCloudSamples
         }
         // [END monitoring_alert_list_policies]
 
+        // [START monitoring_alert_list_channels]
         static void ListNotificationChannels(string projectId)
         {
             var client = NotificationChannelServiceClient.Create();
@@ -150,9 +151,9 @@ namespace GoogleCloudSamples
                 Console.WriteLine();
             }
         }
+        // [END monitoring_alert_list_channels]
 
         // [START monitoring_alert_backup_policies]
-        // [START monitoring_alert_list_channels]
         static void BackupPolicies(string projectId, string filePath)
         {
             var policyClient = AlertPolicyServiceClient.Create();
@@ -167,7 +168,6 @@ namespace GoogleCloudSamples
                 }, new ProtoMessageConverter()));
         }
         // [END monitoring_alert_backup_policies]
-        // [END monitoring_alert_list_channels]
 
         // [START monitoring_alert_restore_policies]
         // [START monitoring_alert_create_policy]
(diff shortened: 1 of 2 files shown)


monitoring/api/AlertSample/Program.cs
```diff
@@ -114,6 +136,21 @@ static void ListAlertPolicies(string projectId)
         }
         // [END monitoring_alert_list_policies]
 
+        static void ListNotificationChannels(string projectId)
```
simonz130: Do you have a region tag for this sample?

monitoring/api/AlertSample/Program.cs
```diff
@@ -114,6 +136,21 @@ static void ListAlertPolicies(string projectId)
...
         // [END monitoring_alert_list_policies]
 
+        static void ListNotificationChannels(string projectId)
+        {
+            var client = NotificationChannelServiceClient.Create();
+            var response = client.ListNotificationChannels(new ProjectName(projectId));
+            foreach (NotificationChannel channel in response)
+            {
+                Console.WriteLine(channel.Name);
+                if (channel.DisplayName != null)
```
simonz130: This is interesting. So the channel can have empty display name? If it's valid statement, then perhaps we may want to print something else that is always present, such as unique identifier of the channel. Otherwise for channels that don't have display name but are valid channel, you will not "list" them.

monitoring/api/AlertSample/Program.cs
```diff
@@ -359,5 +396,29 @@ static object EnablePolicies(string projectId, string filter, bool enable)
         }
         // [END monitoring_alert_enable_policies]
         // [END monitoring_alert_disable_policies]
+
+        static void DeleteAlertPolicy(string policyNameString)
```
simonz130: region tag for this and the next one?

monitoring/api/AlertSample/Program.cs
```diff
@@ -359,5 +396,29 @@ static object EnablePolicies(string projectId, string filter, bool enable)
...
+            var client = NotificationChannelServiceClient.Create();
+            NotificationChannelName channelName;
+            if (!NotificationChannelName.TryParse(channelNameString, out channelName))
+            {
+                throw new BadNameException();
+            }
+            client.DeleteNotificationChannel(channelName, force);
+            Console.WriteLine($"Deleted {channelName}.");
+        }
+        public class BadNameException : Exception { }
```
simonz130: since it's a sample code, maybe worth throwing standard exception with the right message to not complicate things?

monitoring/api/AlertSample/Program.cs
SurferJeffAtGoogle: Done.

monitoring/api/AlertSample/Program.cs
SurferJeffAtGoogle: The code already prints the channel.Name, which is always present.  If the display name is not present, the channel.Name should still tell the user that the channel is there.

monitoring/api/AlertSample/Program.cs
SurferJeffAtGoogle: Added a region tag for deleting the notification channel.  No such tag was requested (by tech writers or others) for deleting a policy.

SurferJeffAtGoogle: I just needed the code to clean up a test project.

monitoring/api/AlertSample/Program.cs
SurferJeffAtGoogle: Done.

SurferJeffAtGoogle:
Thank you for the review and the approval.  I will merge when some tests pass.

PR #552, opened 2018-05-25
Retry vision

SurferJeffAtGoogle:


vision/api/VisionTest/VisionTest.cs
```diff
@@ -63,7 +65,7 @@ public void TestFacesWithFace()
         public static byte[] ToPngBytes(string imagePath)
         {
             var stream = new MemoryStream();
-            using (var image = Image.FromFile(imagePath))
+            using (var image = System.Drawing.Image.FromFile(imagePath))
```
simonz130: nit: do you still need "using System.Drawing;" at line 17 after this?

vision/api/VisionTest/VisionTest.cs
```diff
@@ -15,6 +15,8 @@
 using System.IO;
 using Xunit;
 using System.Drawing;
+using System;
+using Google.Cloud.Vision.V1;
```
dzlier-gcp: Are there any style rules for ordering these imports? This hurts me.

Issue #434, opened 2018-01-24
warning NU1701: Package 'EnterpriseLibrary.TransientFaultHandling 6.0.1304' was restored using '.NETFramework,Version=v4.6.1' instead of the project target framework '.NETCoreApp,Version=v2.0'. This package may not be fully compatible with your project. 


SurferJeffAtGoogle:
When building the Spanner project.  Far from fatal, but we should be using https://www.nuget.org/packages/TransientFaultHandling.Core/

SurferJeffAtGoogle:
Hmm.  That library was last updated in 2012, so we may need to find another one.

simonz130:
I've had the same error for BigQuery project. Fixed it there to use Retrier from TestUtils. I'll fix this too.

simonz130:
Submitted PR: https://github.com/GoogleCloudPlatform/dotnet-docs-samples/pull/571

mjr2001us:
Warning	NU1701	Package 'Antlr 3.4.1.9004' was restored using '.NETFramework,Version=v4.6.1 in MVC5 ,netcore

PR #571, opened 2018-06-26
Fixing Spanner code to not depend on enterprise library nuget packages

simonz130:
Enterprise library retry nuget package targets .NET Framework 4.6.1. If we reference it, we won't be able to run the code on Linux. We could (potentially) depend on testUtil project that includes retryRobot implementation, but that means that Spanner project has to pull many unnecessary nuget packages for GCP.

Therefore, I've added a tiny implementation of retryRobot and used it to replace retry code from Enterprise library.

### DIFF
--- a/spanner/api/Spanner/Program.cs
+++ b/spanner/api/Spanner/Program.cs
@@ -15,16 +15,14 @@
  */
 
 using System;
+using System.Collections.Generic;
 using System.Threading;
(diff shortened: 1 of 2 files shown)


simonz130:
This also fixes https://github.com/GoogleCloudPlatform/dotnet-docs-samples/issues/434

PR #574, opened 2018-07-03
Adding Vision OCR sample

telpirion:
This pull request adds the following region tag to the C# Vision samples:

vision_async_detect_document_ocr

The Python canonical (+ test) is here:
https://github.com/GoogleCloudPlatform/python-docs-samples/blob/master/vision/cloud-client/detect/detect_pdf.py
https://github.com/GoogleCloudPlatform/python-docs-samples/blob/master/vision/cloud-client/detect/detect_pdf_test.py


vision/api/VisionTest/VisionTest.cs
```diff
@@ -195,6 +197,25 @@ public void DetectCropHint()
             Assert.Contains("\tX:", output.Stdout);
             Assert.Contains("\tY:", output.Stdout);
         }
+
+        [Fact]
+        public void DetectDocument()
+        {
+            var bucketName = System.Environment.GetEnvironmentVariable("CLOUD_STORAGE_BUCKET");
+            var outputPrefix = "OCR_PDF_TEST_OUTPUT";
+            var gcsSourceURI = $"gs://{bucketName}/HodgeConj.pdf";
```
beccca: Please put the .pdf being used to test into `./VisionTest/data/`

Also, if it is not already, please use the same pdf file as the Python tests use

Finally, my personal CLOUD_STORAGE_BUCKET does not contain this file so this will presumably fail. Please use the Storage helpers to upload this file before the tests, presuming it's not already

See: https://github.com/telpirion/dotnet-docs-samples/blob/2dd06fa770c4688370ae62c5f7de9cedf83b5fa3/vision/api/VisionTest/VisionTest.cs#L242-L266

beccca: screen shot 2018-07-16 at 12 13 28 pm

beccca: Annoyingly, this is not where it should be in the python-docs-samples (we should fix that)

https://github.com/GoogleCloudPlatform/python-docs-samples/tree/master/vision/cloud-client/detect/resources

vision/api/Detect/Program.cs
```diff
@@ -318,6 +333,109 @@ private static object DetectDocText(Image image)
...
 
+        // [START vision_async_detect_document_ocr]
+        private static object DetectDocument(string gcsSourceUri, string gcsDestinationUri)
+        {
+
+            // Supported mime_types are: 'application/pdf' and 'image/tiff'
+            string mimeType = "application/pdf";
+
+            // How many pages should be grouped into each json output file.
+            // With a file of 5 pages
```
beccca: What does "With a file of 5 pages" mean?

vision/api/Detect/Program.cs
```diff
@@ -318,6 +333,109 @@ private static object DetectDocText(Image image)
...
+
+            Console.WriteLine("Waiting for the operation to finish");
+
+            operation.PollUntilCompleted();
+
+            // Once the rquest has completed and the output has been
+            // written to GCS, we can list all the output files.
+            var storageClient = StorageClient.Create();
+
+            Uri destinationBucket = new Uri(gcsDestinationUri, UriKind.Absolute);
```
beccca:
```cs
Uri destinationBucket = new Uri(gcsDestinationUri, UriKind.Absolute);
```
⤋
```cs
var destinationBucket = new Uri(gcsDestinationUri, UriKind.Absolute);
```

Prefer implicit typing > explicit

Exceptions include times when the explicit type is both (a) unclear from the code (b) something the users should know about

vision/api/Detect/Program.cs
```diff
@@ -318,6 +333,109 @@ private static object DetectDocText(Image image)
...
+            string bucketName = destinationBucket.Host;
+            string prefix = destinationBucket.GetComponents(
```
beccca: Pretty gnarly for a sample whose intent is to demonstrate this feature only

Consider accepting bucket name & path prefix as method parameters to make code more clear

(I realize this will make the CLI's inputs differ from Python, but I approve of this to provide more clarity)

vision/api/Detect/Program.cs
```diff
@@ -318,6 +333,109 @@ private static object DetectDocText(Image image)
             return 0;
         }
 
+        // [START vision_async_detect_document_ocr]
+        private static object DetectDocument(string gcsSourceUri, string gcsDestinationUri)
```
beccca: Consider updating method signature (and CLI) to something like this:

```cs
private static object DetectDocument(string gcsSourceUri, string destinationBucket, string destinationPrefix)
```

Mentioned below in [this comment](https://github.com/GoogleCloudPlatform/dotnet-docs-samples/pull/574/files#r202859816)

vision/api/Detect/Program.cs
```diff
@@ -318,6 +333,109 @@ private static object DetectDocText(Image image)
...
+            // Supported mime_types are: 'application/pdf' and 'image/tiff'
+            string mimeType = "application/pdf";
+
+            // How many pages should be grouped into each json output file.
+            // With a file of 5 pages
+            int batchSize = 2;
+
+            var client = ImageAnnotatorClient.Create();
+
+            var feature = new Feature{
```
beccca: I found these little variables very confusing when I read the sample.

I have to mentally process [FORTY THREE LOC](https://github.com/telpirion/dotnet-docs-samples/blob/2dd06fa770c4688370ae62c5f7de9cedf83b5fa3/vision/api/Detect/Program.cs#L340-L383) ([43 LOC](https://github.com/telpirion/dotnet-docs-samples/blob/2dd06fa770c4688370ae62c5f7de9cedf83b5fa3/vision/api/Detect/Program.cs#L340-L383)) _before the API is even called._

**AND...** the sample is _backwards_. The tiniest structure (feature) is created _first_, but it's one of the _last_ parts of the request to be constructed.

The only way I could comprehend this sample was to scroll down to find the `client.AsyncBatchAnnotateFiles(requests)` and then read UPWARDS to put the puzzle pieces together

----

Oof! – So. What do we do? How can we help make this easier to comprehend?

----

Well. As a developer, I would like to start with the API call. If that's not possible, let's at least start by building up the **Request** <– _an object which a developer would _expect_ to find in a client library/API sample_

Think about reading the API Reference. You might look at an object like [AnnotateImageRequest](https://cloud.google.com/vision/docs/reference/rpc/google.cloud.vision.v1#google.cloud.vision.v1.AnnotateImageRequest) & your goal is to understand the structure of the request, right? If you look at the request object from the OUTSIDE –> Inwards things make more sense

Compare:

 - MimeType
 - BatchSize
 - Feature
 - GcsSource
 - InputConfig
 - GcsDestination
 - OutputConfig
 - Async Annotate File Request
 - Operation!

To this:

 - Async Annotate File Request
   - Input Config
    - GcsSource
    - MimeType
  - OutputConfig
    - GcsDestination
    - BatchSize
  - Features
 - Operation!

One is **flat**. The other is **hierarchical**.

This is part of why I always recommend constructing the request object _inline_, when kicking off the API method, eg:

```cs
var client = LanguageServiceClient.Create();
var response = client.AnalyzeEntitySentiment(new Document()
{
    Content = text,
    Type = Document.Types.Type.PlainText
});
```

At the end of the day, it's about whatever we think is the (a) easiest to consume (b) highlights what we want to highlight (c) does not include extraneous code (d) is not overly long, pay attention to vertical space consumption, aka LOC, as well as horizontal usage/display

----

Here's an IDEA for how I might have coded this API request (note: I haven't gotten to the GCS JSON bottom half yet)

```cs
var client = ImageAnnotatorClient.Create();

var asyncRequest = new AsyncAnnotateFileRequest
{
    InputConfig = new InputConfig
    {
        GcsSource = new GcsSource { Uri = gcsSourceUri },
        // Supported mime_types are: "application/pdf" and "image/tiff"
        MimeType = "application/pdf"
    },
    OutputConfig = new OutputConfig
    {
        GcsDestination = new GcsDestination { Uri = gcsDestinationUri },
        // How many PDF/TIFF pages should be grouped into each JSON file
        BatchSize = 2,
    }
};

asyncRequest.Features.Add(new Feature {
    Type = Feature.Types.Type.DocumentTextDetection
});

var operation = client.AsyncBatchAnnotateFiles(requests: new[]{ asyncRequest });
```

### My 2¢

vision/api/VisionTest/VisionTest.cs
```diff
@@ -195,6 +197,25 @@ public void DetectCropHint()
             Assert.Contains("\tX:", output.Stdout);
             Assert.Contains("\tY:", output.Stdout);
         }
+
+        [Fact]
```
beccca: Move this test into its own class

Every test in `CommonTests` is run 3 times with various arguments provided (once to test using local .png file, once to test using http:// url to .png, once to test using gs:// url to .png)

We don't want that :)

You could add it right below, class... `PdfOcrTests`... with one `DetectDocument()` method, as is... then you can could a line to have it upload the .pdf file before the test runs or something :)

vision/api/Detect/Program.cs
```diff
@@ -318,6 +333,109 @@ private static object DetectDocText(Image image)
...
+            var annotation = firstPageResponses.FullTextAnnotation;
+
+            // Here we print the full text from the first page.
+            // The response contains more information:
+            // annotation/pages/blocks/paragraphs/words/symbols
+            // including confidence scores and bounding boxes
+            Console.WriteLine($"Full text: \n {annotation.Text}");
+
+            // Clean up local file.
+            File.Delete(localPath);
```
beccca: Remove. Do not download a temporary file.

vision/api/Detect/Program.cs
```diff
@@ -318,6 +333,109 @@ private static object DetectDocText(Image image)
...
+            foreach (var blob in blobList)
+            {
+                Console.WriteLine(blob.Name);
+            }
+
+            // Process the first output file from GCS.
+            // Select the first JSON file from the objects in the list.
+            var output = blobList.Where(x => x.Name.Contains(".json")).First();
+            var localPath = Path.GetFileName(output.Name);
+            using (var stream = new FileStream(localPath,
```
beccca: Do not write a file – `s/FileStream/MemoryStream/`

vision/api/VisionTest/VisionTest.cs
beccca: Example: remove the current `DetectDocument` method from `CommonTests` and add the following class 
after `TutorialTests` before `CommonTests`

```cs
// TutorialTests
// ...

/// <summary>
/// Runs tests detecting text from PDF or TIFF files
/// These require uploading a .pdf to Google Cloud Storage
/// and then API results are saved to Google Cloud Storage
/// </summary>
public class PdfDocumentTests : IClassFixture<RandomBucketFixture>
{
  readonly string _bucketName;
  readonly BucketCollector _bucketCollector;
  readonly string _pdfFileName = "HodgeConj.pdf";

  public PdfDocumentTests(RandomBucketFixture bucketFixture)
  {
    _bucketName = bucketFixture.BucketName;
    _bucketCollector = new BucketCollector(_bucketName);
    var localPath = Path.Combine("data", _pdfFileName)
    _bucketCollector.CopyToBucket(localPath, _pdfFileName);
  }

  [Fact]
  public void TestDetectPdfDocument()
  {
    var outputPrefix = "OCR_PDF_TEST_OUTPUT";
    var gcsSourceURI = $"gs://{_bucketName}/{_pdfFileName}";
    var gcsDestinationURI = $"gs://{_bucketName}/{outputPrefix}/";

    var output = Run("ocr", gcsSourceURI, gcsDestinationURI);
    var storageClient = StorageClient.Create();
    var bucket = storageClient.GetBucket(_bucketName);
    var blobList = storageClient.ListObjects(bucketName, outputPrefix);

    Assert.Contains("Full text:", output.Stdout);
    Assert.Contains("Hodge conjecture", output.Stdout);
    Assert.Equal(blobList.Count(), 1);
  }
  
  public void Dispose() => _bucketCollector.Dispose();
}

// ...
// CommonTests
```

beccca: ... if this blows up, you might have to do the following, I didn't have intellisense, just typed this out ...

```cs
public void Dispose() => ((IDisposable)_bucketCollector).Dispose();
```

beccca:
As promised, here's an overview comment with a task list of recommendations to get this done quickly/minimal changes–

 - [ ] Replace AsyncAnnotateFileRequest call with [provided alternative clean, hierarchical code](https://github.com/GoogleCloudPlatform/dotnet-docs-samples/pull/574/files#r202867706)
 - [ ] Replace explicit types with `var` for variable declarations
 - [ ] Do not download JSON file –> "download" to a MemoryStream to get JSON text from GCS
 - [ ] Replace the DetectDocument test with [provided alternative that makes use of existing GCS helpers](https://github.com/GoogleCloudPlatform/dotnet-docs-samples/pull/574/files#r202895605)

That's all 🙏

vision/api/VisionTest/VisionTest.cs
beccca: FYI References
 - [GCS Test Helper file](https://github.com/GoogleCloudPlatform/dotnet-docs-samples/blob/b573a1402310f6ebc089cbdd7666211f275dbf97/testutil/TestBucket.cs) – it's only ~ 100 LOC with the following 2 helper classes
   - [RandomBucketFixture](https://github.com/GoogleCloudPlatform/dotnet-docs-samples/blob/b573a1402310f6ebc089cbdd7666211f275dbf97/testutil/TestBucket.cs#L26)
   - [BucketCollector](https://github.com/GoogleCloudPlatform/dotnet-docs-samples/blob/b573a1402310f6ebc089cbdd7666211f275dbf97/testutil/TestBucket.cs#L62)

vision/api/VisionTest/VisionTest.cs
```diff
@@ -72,6 +74,43 @@ public static byte[] ToPngBytes(string imagePath)
...
+
+        public PdfDocumentTests(RandomBucketFixture bucketFixture)
+        {
+            _bucketName = bucketFixture.BucketName;
+            _bucketCollector = new BucketCollector(_bucketName);
+            var localPath = Path.Combine("data", _pdfFileName);
+            _bucketCollector.CopyToBucket(localPath, _pdfFileName);
+        }
+
+        [Fact(Skip = "Test hangs")]
```
beccca: ... this test gets run!? 🤯

----

screen shot 2018-07-23 at 2 56 49 pm

----

w.t.f. mate

telpirion: I'm mystified 0_o

telpirion: I isolated the cause of the hang: it's line 93. 

I'm going to use a similar technique as is used in the SpeechText project: defining an Upload() method in this class that uploads the local file to the GCS bucket.

PR #579, opened 2018-07-13
Add IAM code snippets

awkoren:
Code snippets for IAM documentation: https://cloud.google.com/iam/docs/


awkoren:
I see my code failed linting -- I'm developing using .NET Core on Linux so I can't run codeformatter.exe. Is there a suggested workaround?

PR #582, opened 2018-07-24
Introducing Petshop modernization demo

simonz130:
In this demo we show how you can modernize .NET application by breaking monolith into
microservices, containerizing them, and then leveraging all the superpowers of running
containers using Kubernetes in Google Kubernetes Engine. We show how easy is it to
scale the app, handle resiliency and do production debugging.


SurferJeffAtGoogle:
Please add a README.md.

PR #595, opened 2018-08-13
Extend the timeout to call CreateUptimeCheckConfig.

SurferJeffAtGoogle:
It can take a long time, causing tests to fail.
Server-side issue tracked in bug 112552528.


monitoring/api/UptimeCheckSample/Program.cs
```diff
@@ -27,6 +27,7 @@
 using static Google.Api.MetricDescriptor.Types;
 using Google.Protobuf.WellKnownTypes;
 using System.Net;
+using Google.Api.Gax.Grpc;
```
simonz130: please, sort

SurferJeffAtGoogle: Done.

PR #607, opened 2018-08-21
Remove modernizePetshop.  liftAndShiftPetShop is the good code we wan…

SurferJeffAtGoogle:
…t to keep.


SurferJeffAtGoogle:
Somehow, we merged in two copies of the same code.  One was properly formatted.  The other wasn't.

PR #633, opened 2018-09-13
Add a note that the code samples are designed to demonstrate API call…

SurferJeffAtGoogle:
…s, and not the best engineering practices.

Change-Id: Ifd30ae6070aa4eed16b2e66fd6304e9a1871c012

### DIFF
--- a/translate/canon/README.md
+++ b/translate/canon/README.md
@@ -31,7 +31,9 @@ A collection of samples that demonstrate how to call the
 ## Designed to demonstrate Google APIs
 
 These code samples are designed to demonstrate simple code that calls Google
-APIs.  So sometimes, they don't conform to engineering best practices like
+APIs.  They're optimized for being embedded into documentation on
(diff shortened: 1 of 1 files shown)


translate/canon/README.md
```diff
@@ -28,6 +28,12 @@ A collection of samples that demonstrate how to call the
     Test execution time: 2.6078 Seconds    
 	```
 
+## Designed to demonstrate Google APIs
+
+These code samples are designed to demonstrate simple code that calls Google
+APIs.  So sometimes, they don't conform to engineering best practices like
+"don't copy and paste code."
```
simonz130: Maybe worth also mentioning that they are optimized for being embedded at cloud.google.com documentation and therefore there is a separate file per snippet.

SurferJeffAtGoogle: Done.

Thanks for the review!

PR #645, opened 2018-09-21
Add comment to pubsub publish sample

chrisdunelm:
Explaining returned Task from PublishAsync a little
Fixes https://github.com/googleapis/google-cloud-dotnet/issues/2532

### DIFF
--- a/pubsub/api/PubsubSample/Program.cs
+++ b/pubsub/api/PubsubSample/Program.cs
@@ -280,7 +280,9 @@ namespace GoogleCloudSamples
             {
                 // Record all publishing Tasks. When each Task completes
                 // without error, the message has been successfully published.
-                // In real use, don't simply store all publish Tasks forever.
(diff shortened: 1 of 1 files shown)


pubsub/api/PubsubSample/Program.cs
```diff
@@ -278,6 +278,9 @@ public static object CreateTopic(string projectId, string topicId)
             // batches.
             foreach (string text in messageTexts)
             {
+                // Record all publishing Tasks. When each Task completes
+                // without error, the message has been successfully published.
+                // In real use, don't simply store all publish Tasks forever.
```
SurferJeffAtGoogle: Please update the code comment with an answer to the question:
What should the user do "in real use"?

chrisdunelm: Done. PTAL

PR #657, opened 2018-10-09
Dlp redux

SurferJeffAtGoogle:
Update one sample https://github.com/GoogleCloudPlatform/dotnet-docs-samples/blob/master/dlp/api/DlpSample/DeIdentify.cs#L28
and its test to the proposed canonical form.


SurferJeffAtGoogle:
> First thoughts after reading through [DeidentifyMasking](https://github.com/SurferJeffAtGoogle/csharp-docs-samples/blob/dlpRedux/dlp/api/DlpSnippets/DeidentifyMasking.cs)... **2¢**
> 
> 1. Looks fabulous! tada
> 
> * I adore having 1 file per sample, so it can include the imports easily and everything! And I love that the sample uses real values, so this sample runs ok_hand without a developer having to add their own values!
> 
> 1. There is no **description** of _what the sample does_, e.g. `"De-identify sensitive data in text content"`

Added.

> * This could, for example, be added as a method comment on `DeidentiyMasking(...)`
> 
> 1. There is no **description** of the _method parameters_ the sample accepts, e.g. `projectId`

Added.
 
> * This for, for example, be added as `<param>` method comment on `DeidentiyMasking(...)`
> * For `projectId`, the user can see a default value of `"YOUR-PROJECT-ID"` provided; uncertain of the readability of this vs using param documentation, which is **familiar to developers**
> 
> 1. Why is the class `partial`? <-- something which "surprised" me when I read through; unexpected & I have no idea why _(I would have to dig around to figure it out)_

For convenience maintaining the code.  The tests are simpler because they don't have to instantiate a little new class for every test.  The [translate sample](https://github.com/GoogleCloudPlatform/dotnet-docs-samples/tree/master/translate/api/Snippets) was the former canonical sample and that's what it does.  It also makes the code easier to write.  First, I don't have to choose what to name the class.  Second, when it's time to write a new sample, I just copy an old .cs file and the class name is one less thing I have to update.

I'm surprised the partial class was so distracting.

> 1. Nit: What is `using Google.Api.Gax.ResourceNames;` needed for?
> 
> * It's a pretty gnarly `using` statement imo. "What is `Gax`? Why `Google.Api` not `Google.Cloud`?"
> * Library feedback, not so much sample

I don't see how I can improve this.

> 1. Could y'all add comments to help provide a readable outline/structure of the sample?
> 
> * It's a lot of text to look at without comments helping me break it down into sections. Comments like these would be helpful:
> * `// Instantiates a client`
> * `// Construct Request`
> * `// Execute Request`
> * `// Process Response`

Done.

> * `// Imports the Google Cloud Data Loss Prevention library`

I find such a comment does not tell me anything that the code already does not tell me.  A using statement does only one thing in C#.  I think in other languages, a comment here is more appropriate.

> 1. To my eyes, it's non-trivial/quick for me to pick out the parts of the sample that I, as a developer, care about...
> 
> * That is to say: the INPUT to the API which I'll want to edit to running the request with my own data. For this sample, the values which my eyes have trouble picking out are: `Item.Content.Value` and `InspectConfig.InfoTypes[0]`
> * Could these be moved into variables near the top of the sample method? e.g.
> 
> ```cs
> public string DeidentiyMasking(string projectId = "YOUR-PROJECT-ID")
> {
>   var infoTypeName = "US_SOCIAL_SECURITY_NUMBER";
>   var contentText = "My SSN is 372819127.";
> 
>   // ... the rest of the sample, as is .... using the variables ...
> }
> ```

Done.

SurferJeffAtGoogle:
Also, I restored all the old files so that we can merge this PR when we agree this one sample looks like the canonical sample we want.  That way, we don't have to migrate all the code to unblock others like Alix who need a stable canonical sample yesterday.

beccasaurus:
1. `partial` class

> [the class is partial] For convenience maintaining the code. The tests are simpler because they don't have to instantiate a little new class for every test.

Would it be OK for us to "instantiate a little new class for every test"? Seems like an easy fix. This way the class can be a **familiar** `public class ProductSnippets` 🙂

----

2. Do you happen to know what `Google.Api.Gax.ResourceNames` is needed for?

None of the other samples in `dotnet-docs-samples` use this namespace. And very few how-to samples (3 individual snippets) use the `Google.Api.Gax` namespace at all. 

I can take this feedback back to the C# Client Library team

----

3. `// Imports the Google Cloud Data Loss Prevention library`

The comment could be useful to sample templates in a number of ways

 - a. Clarify which namespaces being imported are required for the client library, versus SDK imports
 - b. Using the secondary comment `// Import other required libraries` helps clarify imports needed to perform additional required functionality in the sample, e.g. Base64 encoding/decoding, File reading/saving, etc.

^--- see [this currently proposed Node sample](https://github.com/googleapis/nodejs-dlp/blob/9270b3cda8283bed77c1a84dd3dc7183e64053dc/samples/dlp_inspect_file.js#L19-L24) for an example of this:

dlp/api/Snippets/DeidentifyMasking.cs
```diff
@@ -0,0 +1,80 @@
...
+// the License.
+
+// [START dlp_deidentify_masking]
+
+using Google.Api.Gax.ResourceNames;
+using Google.Cloud.Dlp.V2;
+using System;
+using System.Collections.Generic;
+
+partial class DlpSnippets
```
beccasaurus: Let's do this _(from our discussion offline)_

```cs
public class Dlp DeidentiyMasking
{
  /// <summary>...</summary>
  /// ...
  public string DeidentiyMasking(
    // ...
}
```

SurferJeffAtGoogle:
I incorporated all the comments from the meeting and added an InspectFile sample.

PR #670, opened 2018-10-23
Asset: Add quickstart code for export-assets API.

peter-zheng-g:


asset/quickstart/ExportAssetsTest/ExportAssetsTest.cs
```diff
@@ -0,0 +1,95 @@
...
+        public ExportAssetsTest(ITestOutputHelper output)
+        {
+            _testOutput = output;
+        }
+
+        [Fact]
+        public void TestExportAsests()
+        {
+            string projectId = Environment.GetEnvironmentVariable("GOOGLE_PROJECT_ID");
+            string bucketName = String.Format("{0}-for-assets", projectId);
```
SurferJeffAtGoogle: This can still cause problems if two agents are running the test at the same time.  Instead, add a dependency to ../../../../testutil/testutil.csproj and use https://github.com/GoogleCloudPlatform/dotnet-docs-samples/blob/master/testutil/TestBucket.cs which will set up and tear down an bucket with a random name.

asset/quickstart/ExportAssets/ExportAssets.cs
```diff
@@ -0,0 +1,47 @@
...
+using Google.Api.Gax.ResourceNames;
+using Google.Cloud.Asset.V1Beta1;
+using System;
+
+
+public class ExportAssets
+{
+    public static void Main(string[] args)
+    {
+        string projectId = Environment.GetEnvironmentVariable("GOOGLE_PROJECT_ID");
```
SurferJeffAtGoogle: Use a literal string here, and then replace it in runTests.ps1, like is done here: https://github.com/GoogleCloudPlatform/dotnet-docs-samples/blob/master/trace/api/runTests.ps1

SurferJeffAtGoogle:
btw, thank you for matching the existing pattern for samples.

asset/quickstart/ExportAssetsTest/ExportAssetsTest.cs
peter-zheng-g: Thanks. Not sure in which situation this can be run concurrently for exactly the same project. If we use random bucket, shall I pass the bucket name to the dev code (../ExportAssets/ExportAssets.cs) with param or with env var?

peter-zheng-g: Or, I can use this non-random bucket, but use a random file in dev code, would that sound good for you?

peter-zheng-g: Done.

asset/quickstart/ExportAssets/ExportAssets.cs
peter-zheng-g: I tried to use literal string for bucketName, and replace it with Environment.GetEnvironmentVariable by using runTests.ps1, somehow it complains that "ExportAssets.cs(26,66): error CS1002: ; expected", I am wondering if you know what I missed there? I guess the ps1 didn't replace it correctly.

asset/quickstart/ExportAssetsTest/runTests.ps1
```diff
@@ -0,0 +1,24 @@
...
+# WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied. See the
+# License for the specific language governing permissions and limitations under
+# the License.
+
+Import-Module -DisableNameChecking ..\..\..\BuildTools.psm1
+
+$getBucketNameClause = 'Environment.GetEnvironmentVariable("AssetBucketName")'
+BackupAndEdit-TextFile @("..\ExportAssets\ExportAssets.cs") @{
+    "YOUR-GOOGLE-PROJECT-ID" = $env:GOOGLE_PROJECT_ID;
+    '"YOUR-BUCKET-NAME"' = $getBucketNameClause
```
peter-zheng-g: Thanks!

peter-zheng-g:
@SurferJeffAtGoogle Now I see that "Cloud Asset API has not been used in project 1061780273775 before or it is disabled. Enable it by visiting https://console.developers.google.com/apis/api/cloudasset.googleapis.com/overview?project=1061780273775 then retry. If you enabled this API recently, wait a few minutes for the action to propagate to our systems and retry.". Can you please help enable Cloud Asset API for related projects? Thanks!

SurferJeffAtGoogle:
The test fails because the test user does not have the necessary permission:
https://sponge.corp.google.com/target?id=375d26ce-3c0c-469d-80a9-b816532f968a&target=github/dotnet-docs-samples/asset/quickstart/ExportAssetsTest/TestResults

Please add the necessary permission dotnet-docs-samples\.kokoro-windows\Initialize-GoogleProjectForTesting.ps1

peter-zheng-g:
> The test fails because the test user does not have the necessary permission:
> https://sponge.corp.google.com/target?id=375d26ce-3c0c-469d-80a9-b816532f968a&target=github/dotnet-docs-samples/asset/quickstart/ExportAssetsTest/TestResults
> 
> Please add the necessary permission dotnet-docs-samples.kokoro-windows\Initialize-GoogleProjectForTesting.ps1

Not sure what it means... If it means the user calling the API doesn't have permission, the what we need to do is, add the user to the targeting project as a viewer on Pantheon. But please let me know if I missed anything. Thanks.

peter-zheng-g:
> > The test fails because the test user does not have the necessary permission:
> > https://sponge.corp.google.com/target?id=375d26ce-3c0c-469d-80a9-b816532f968a&target=github/dotnet-docs-samples/asset/quickstart/ExportAssetsTest/TestResults
> > Please add the necessary permission dotnet-docs-samples.kokoro-windows\Initialize-GoogleProjectForTesting.ps1
> 
> Not sure what it means... If it means the user calling the API doesn't have permission, the what we Found the file. Added cloud asset service to the service list.

.kokoro-windows/Initialize-GoogleProjectForTesting.ps1
```diff
@@ -98,6 +98,7 @@ roles/monitoring.admin
 roles/pubsub.admin
 roles/resourcemanager.projectIamAdmin
 roles/storage.admin
+roles/viewer
```
SurferJeffAtGoogle: This role is too broad.  Please chose a role with more limited privileges.

peter-zheng-g: Replaced it with cloudasset.viewer.

peter-zheng-g:
I see that the test is still complaining "The caller does not have permission". With cloudasset.viewer role configured for the API caller, this error usually happens when the caller is trying to exportAssets for a different project for which the caller doesn't have cloudasset.viewer role configured. Jeff, can you please help check if the ps script configures the role for the caller in the caller's own project or the target project against which the API was called? If the role was only configured for the caller in the caller's own project, then it's probably not good enough.

peter-zheng-g:
Thanks very much!

SurferJeffAtGoogle:
Test is still failing.  I'm continuing to investigate...

peter-zheng-g:
I've sent you an email. The timeout issue may also be related to dev code pollUntilCompleted. I'll let you know once I got any response from that investigation. Thanks very much.

PR #678, opened 2018-11-13
Re-enable the dialogflow test.

SurferJeffAtGoogle:
Change-Id: I8480b93fa5d4c8d1195d3d475dec1205ae1d89a0

### DIFF
--- a/dialogflow/api/Test/runTests.ps1
+++ b/dialogflow/api/Test/runTests.ps1
@@ -17,5 +17,4 @@ Set-TestTimeout 1800
 
 dotnet restore
 dotnet build
-# TODO: Fix bug 110877421.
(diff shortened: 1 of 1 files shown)


beccasaurus:
@telpirion Do you have availability to look into the failing C# tests for Dialogflow

I think the steps to fix are as follows:
1. Add Dispose() cleanup to the tests which create things (Entities/Intents/...) which attempts to delete the created object (*attempts to* because it's okay if the delete fails because Create may have broken)
2. We should probably delete all existing objects for the Agent the tests are using. @SurferJeffAtGoogle may be able to help / do this, as he did some cleanup previously

SurferJeffAtGoogle:
Could we all meet at our usual 2pm Friday time slot to pair-program this?

PR #679, opened 2018-11-14
BigQuery Sample Refactor/Rewrite

alixhami:
This PR updates the BigQuery samples to follow [C# canonical example](https://github.com/GoogleCloudPlatform/dotnet-docs-samples/pull/657) as well as the [BigQuery canonical samples](https://github.com/googleapis/google-cloud-python/blob/master/bigquery/docs/snippets.py), which were written in python.


bigquery/api/SimpleApp/Program.cs
```diff
@@ -0,0 +1,54 @@
...
+// [END bigquery_simple_app_deps]
+
+namespace GoogleCloudSamples
+{
+    public class Program
+    {
+        public static void Main(string[] args)
+        {
+            string projectId = "YOUR-PROJECT-ID";
+            // [START bigquery_simple_app_client]
```
tswast: Shouldn't the `projectId` line be included here so that folks can see that it is a string when they create the client?

bigquery/api/SimpleApp/Program.cs
```diff
@@ -0,0 +1,54 @@
...
+            // [START bigquery_simple_app_query]
+            string query = @"SELECT
+                CONCAT(
+                    'https://stackoverflow.com/questions/',
+                    CAST(id as STRING)) as url, view_count
+                FROM `bigquery-public-data.stackoverflow.posts_questions`
+                WHERE tags like '%google-bigquery%'
+                ORDER BY view_count DESC
+                LIMIT 10";
+            var result = client.ExecuteQuery(query, parameters: null);
```
tswast: Why pass in null for parameters?

bigquery/api/Snippets/BrowseTable.cs
```diff
@@ -0,0 +1,49 @@
...
+            DatasetId = "samples",
+            ProjectId = "bigquery-public-data"
+        };
+        // Load all rows from a table
+        PagedEnumerable<TableDataList, BigQueryRow> result = client.ListRows(
+            tableReference: tableReference,
+            schema: null
+        );
+        // Print the first 10 rows
+        foreach (BigQueryRow row in result.Take(10))
```
tswast: Nice. I really like the LINQ methods in C#.

We could do the same in Python with [islice](https://docs.python.org/3/library/itertools.html#itertools.islice). I wonder if we should use that in Python examples instead of `max_results` since it's more standard Python? I guess `max_results` is here to stay in Python, even if it is confusing with an API parameter of the same name.

bigquery/api/Snippets/DeleteDataset.cs
```diff
@@ -0,0 +1,40 @@
...
+    )
+    {
+        BigQueryClient client = BigQueryClient.Create(projectId);
+        // Delete a dataset that does not contain any tables
+        client.DeleteDataset(datasetId: datasetId1);
+        Console.WriteLine($"Dataset {datasetId1} deleted.");
+        // Use the DeleteDatasetOptions to delete a dataset and its contents
+        client.DeleteDataset(
+            datasetId: datasetId2,
+            options: new DeleteDatasetOptions() { DeleteContents = true }
```
tswast: Would it make sense to split this into two different samples? The [docs](https://cloud.google.com/bigquery/docs/managing-tables#deleting_tables) only have one section for examples, so maybe not.

bigquery/api/Snippets/ExtractTable.cs
```diff
@@ -0,0 +1,38 @@
...
+using System;
+
+public class BigQueryExtractTable
+{
+    public void ExtractTable(
+        string projectId = "your-project-id",
+        string bucketName = "your-bucket-name")
+    {
+        BigQueryClient client = BigQueryClient.Create(projectId);
+        string destinationUri = $"gs://{bucketName}/shakespeare.csv";
```
tswast: Should we show [a wildcard in the destination URI](https://cloud.google.com/bigquery/docs/exporting-data#exporting_data_into_one_or_more_files) so that the sample works when exporting larger table? That gotcha bites me quite often.

Ditto for the other extract formats.

bigquery/api/Snippets/LoadFromFile.cs
```diff
@@ -0,0 +1,50 @@
...
+        // Create job configuration
+        var uploadCsvOptions = new UploadCsvOptions()
+        {
+            SkipLeadingRows = 1,  // Skips the file headers
+            Autodetect = true
+        };
+        using (FileStream stream = File.Open(filePath, FileMode.Open))
+        {
+            // Create and run job
+            BigQueryJob job = client.UploadCsv(
```
tswast: So there's a special "CSV" method in the C# client? This is problematic since we only have one sample for loading local files. 

Does that mean there is a corresponding method for the other file formats? Is there a more generic method we could use?

If not, let's add a comment mentioning some of the other upload options classes & methods.

bigquery/api/Snippets/TableInsertRows.cs
```diff
@@ -0,0 +1,45 @@
...
+            new BigQueryInsertRow(insertId: "row1") {
+                { "name", "Washington" },
+                { "post_abbr", "WA" }
+            },
+            new BigQueryInsertRow(insertId: "row2") {
+                { "name", "Colorado" },
+                { "post_abbr", "CO" }
+            }
+        };
+        client.InsertRows(datasetId, tableId, rows);
```
tswast: Hmm... I'd expect to see checking for partial failure to insert rows, but I can't tell from the API reference how partial failures are reported.

@jskeet What happens when there are errors inserting some rows? Does[InsertRows](https://googleapis.github.io/google-cloud-dotnet/docs/Google.Cloud.BigQuery.V2/api/Google.Cloud.BigQuery.V2.BigQueryClient.html#Google_Cloud_BigQuery_V2_BigQueryClient_InsertRows_Google_Apis_Bigquery_v2_Data_TableReference_System_Collections_Generic_IEnumerable_Google_Cloud_BigQuery_V2_BigQueryInsertRow__Google_Cloud_BigQuery_V2_InsertOptions_) raise an exception?

In the other languages, the equivalent of InsertRows returns a list of errors, which need to be explicitly checked for. The reason for this is that the API method can have partial failures, where only some of the batch of rows could not be inserted.

jskeet: Will need to check. I'm at a conference for the first half of this week, so it probably won't be for a little while, but will make sure I get to it when I can.

bigquery/api/QuickStart/QuickStartTest.cs
```diff
@@ -0,0 +1,66 @@
...
+            var output = _quickStart.Run();
+            Assert.Equal(0, output.ExitCode);
+            var outputParts = output.Stdout.Split(new[] { ' ' });
+            Assert.Equal(expectedOutputFirstWord, outputParts.First().Trim());
+            Assert.Equal(expectedOutputLastWord, outputParts.Last().Trim());
+        }
+
+        public void Dispose()
+        {
+            foreach (string datasetId in _tempDatasetIds)
```
SurferJeffAtGoogle: Imagine the machine running this test loses power right before the call to `DeleteDataset()`.  A `my_new_dataset` dataset will be left behind, and all future attempts to run the test will fail.

I'll add some powershell code that uses a random name.

bigquery/api/SimpleApp/Program.cs
SurferJeffAtGoogle: +1

bigquery/api/SimpleApp/Program.cs
SurferJeffAtGoogle: +1

bigquery/api/SimpleApp/SimpleApp.csproj
```diff
@@ -0,0 +1,21 @@
+<Project Sdk="Microsoft.NET.Sdk">
```
SurferJeffAtGoogle: My first impression is that QuickStart and SimpleApp serve the same purpose.  Why are both necessary?

bigquery/api/Snippets/CopyTable.cs
```diff
@@ -0,0 +1,43 @@
...
+            TableId = "shakespeare",
+            DatasetId = "samples",
+            ProjectId = "bigquery-public-data"
+        };
+        TableReference destinationTable = client.GetTableReference(
+            destinationDatasetId, "destination_table");
+        BigQueryJob job = client.CreateCopyJob(
+            sourceTable, destinationTable)
+            .PollUntilCompleted();  // Wait for the job to complete.
+        return client.GetTable(destinationTable);
```
SurferJeffAtGoogle: In addition to returning a value, each snippet should also be `Console.WriteLine()`ing something interesting.

bigquery/api/SimpleApp/Program.cs
alixhami: It's a required parameter for [ExecuteQuery](https://googleapis.github.io/google-cloud-dotnet/docs/Google.Cloud.BigQuery.V2/api/Google.Cloud.BigQuery.V2.BigQueryClient.html#Google_Cloud_BigQuery_V2_BigQueryClient_ExecuteQuery_System_String_System_Collections_Generic_IEnumerable_Google_Cloud_BigQuery_V2_BigQueryParameter__Google_Cloud_BigQuery_V2_QueryOptions_Google_Cloud_BigQuery_V2_GetQueryResultsOptions_). I agree that it looks weird though. I noticed it's the same for [CreateQueryJob](https://googleapis.github.io/google-cloud-dotnet/docs/Google.Cloud.BigQuery.V2/api/Google.Cloud.BigQuery.V2.BigQueryClient.html#Google_Cloud_BigQuery_V2_BigQueryClient_CreateQueryJob_System_String_System_Collections_Generic_IEnumerable_Google_Cloud_BigQuery_V2_BigQueryParameter__Google_Cloud_BigQuery_V2_QueryOptions_).

bigquery/api/Snippets/BrowseTable.cs
alixhami: I think it makes sense to use the parameter in python in our sample as long as it exists. But yes, it did send me on a wild goose chase when trying to find the equivalent parameter when looking at the different libraries 😬

bigquery/api/Snippets/DeleteDataset.cs
alixhami: I think this would make sense to split since that would make it more consistent with the other samples (only showing one thing). If you'd like to split it, I'll make separate samples for the languages that have this example and then do a different sample tab section for the second 
 force delete sample.

bigquery/api/QuickStart/QuickStartTest.cs
SurferJeffAtGoogle: Done.

bigquery/api/Snippets/ExtractTable.cs
alixhami: I'll update and make a note to update the other languages. Out of curiosity, what happens when you try to export a larger table without a wildcard?

bigquery/api/Snippets/LoadFromFile.cs
alixhami: It looks like there are methods for Avro, JSON, CSV, Parquet, and Orc starting with [UploadAvro](https://googleapis.github.io/google-cloud-dotnet/docs/Google.Cloud.BigQuery.V2/api/Google.Cloud.BigQuery.V2.BigQueryClient.html#Google_Cloud_BigQuery_V2_BigQueryClient_UploadAvro_Google_Apis_Bigquery_v2_Data_TableReference_Google_Apis_Bigquery_v2_Data_TableSchema_System_IO_Stream_Google_Cloud_BigQuery_V2_UploadAvroOptions_) until the bottom of that page in the reference docs. The generic [CreateLoadJob](https://googleapis.github.io/google-cloud-dotnet/docs/Google.Cloud.BigQuery.V2/api/Google.Cloud.BigQuery.V2.BigQueryClient.html#Google_Cloud_BigQuery_V2_BigQueryClient_CreateLoadJob_System_Collections_Generic_IEnumerable_System_String__Google_Apis_Bigquery_v2_Data_TableReference_Google_Apis_Bigquery_v2_Data_TableSchema_Google_Cloud_BigQuery_V2_CreateLoadJobOptions_) method seems to be just for files stored in GCS. I'll add a comment saying there are other formats.

bigquery/api/SimpleApp/SimpleApp.csproj
alixhami: We use them in different places in the docs. QuickStart is on language landing pages and library reference pages, and SimpleApp is used for the client library quickstart. Personally, I dislike the one called "QuickStart" and would rather fully replace it with SimpleApp. What do you think @tswast?

tswast: +1 on removing QuickStart in favor of SimpleApp.

I thought we were going to remove the code sample at https://cloud.google.com/bigquery/docs/reference/libraries (QuickStart) and point to https://cloud.google.com/bigquery/docs/quickstarts/quickstart-client-libraries (SimpleApp) with the new template, anyway?

bigquery/api/Snippets/TableInsertRows.cs
jskeet: Right - that sounds like what I'd expect, and here's the code that checks:
https://github.com/googleapis/google-cloud-dotnet/blob/f14444bbed6056b439c2961c317f27eb7dacc3aa/apis/Google.Cloud.BigQuery.V2/Google.Cloud.BigQuery.V2/BigQueryClientImpl.InsertData.cs#L151

Anything else I need to pay attention to on this PR? (There's a lot of conversation that's tricky to keep up with while also conferencing...)

bigquery/api/SimpleApp/SimpleApp.csproj
alixhami: Per in person discussion, I removed the QuickStart sample. If we need something to be explicitly labeled a "QuickStart" in the future, we will rename SimpleApp.

bigquery/api/Snippets/CopyTable.cs
alixhami: I updated to add a print statement and removed the return because none of the other samples return a value.

PR #691, opened 2018-11-28
Fix the TestBatchGetAssetsHistory().

SurferJeffAtGoogle:
The cloud storage API is *eventually* consistent.  Newly created objects may not be immediately visible.  Add retry logic in the test when the new cloud storage object is not found.


peter-zheng-g:
Cool. Thanks for fixing the issues!

PR #694, opened 2018-11-29
Bigtable: add filter region tag to hello world sample

AVaksman:
Added `creating_a_filter` region tag to follow the standard across examples in other languages.

### DIFF
--- a/bigtable/api/HelloWorld/HelloWorld.cs
+++ b/bigtable/api/HelloWorld/HelloWorld.cs
@@ -166,6 +166,10 @@ namespace GoogleCloudSamples.Bigtable
                     Mutations.SetCell(columnFamily, columnName, s_greetings[s_greetingIndex], new BigtableVersion(DateTime.UtcNow));
                 //[END writing_rows]
 
+                // [START creating_a_filter]
+                RowFilter filter = RowFilters.CellsPerRowLimit(1);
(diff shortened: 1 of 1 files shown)


AVaksman:
@billyjacobson 
Adding `[START creating_a_filter] ... [END creating_a_filter]` region tag to follow example in [Node.js](https://github.com/googleapis/nodejs-bigtable/blob/8a90c8fae94a283d2d49829f8675351cff32bb74/samples/hello-world/index.js#L93-L101)

Added same for python GoogleCloudPlatform/python-docs-samples#1878

bigtable/api/HelloWorld/HelloWorld.cs
```diff
@@ -174,7 +178,7 @@ private static void DoHelloWorld()
 
                 // Read a specific row. Apply filter to return latest only cell value accross entire row.
                 Row rowRead = bigtableClient.ReadRow(
-                    tableName, rowKey: rowKeyPrefix + rowIndex, filter: RowFilters.CellsPerRowLimit(1));
+                    tableName, rowKey: rowKeyPrefix + rowIndex, filter: filter);
```
billyjacobson: Now that this is called filter, are you able to just pass "filter" as the arg instead of "filter: filter"
For below too

PR #698, opened 2018-12-05
Compile Cloud tasks samples

averikitsch:
Is it ok to restructure with a folder for Task Sample and a folder for the App Engine App?


beccasaurus:
Oh! Here's there error...
screen shot 2018-12-05 at 2 48 03 pm

Moved from `ROOT/[1 appengine]/[2 flexible]/[3 cloud tasks]`
to `ROOT/[1 cloud tasks]/[2 appengine]`

Error...
screen shot 2018-12-05 at 2 49 56 pm


Update `runTests.ps1` to have one fewer `../` 😄

PR #703, opened 2018-12-05
Update some more dependencies.

SurferJeffAtGoogle:


pubsub/api/PubsubSample/Program.cs
```diff
@@ -236,28 +237,26 @@ public static object CreateTopic(string projectId, string topicId)
...
 
-        public static PublisherClient GetPublisher(string projectId,
+        public static async Task<PublisherClient> GetPublisherAsync(string projectId,
             string topicId)
         {
             // [START pubsub_publish]
-            PublisherServiceApiClient publisherClient = PublisherServiceApiClient.Create();
-            PublisherClient publisher = PublisherClient.Create(
-                new TopicName(projectId, topicId), new[] { publisherClient });
+            PublisherClient publisher = await PublisherClient.CreateAsync(
```
simonz130: What's the point to await creation of the client and then return it right away? You don't really need the async context to continue the execution of the method (typical reason for using await)
Why not just making the method asynchronous without async..await?

```csharp

public static Task<PublisherClient> GetPublisherAsync(
            string projectId,
            string topicId)
{
   // [START pubsub_publish]
   Task<PublisherClient> publisherTask = PublisherClient.CreateAsync(
         new TopicName(projectId, topicId));
  // [END pubsub_publish]
  return publisherTask;
}
```

pubsub/api/PubsubSample/Program.cs
```diff
@@ -236,28 +237,26 @@ public static object CreateTopic(string projectId, string topicId)
...
             // [END pubsub_publish]
             return publisher;
         }
 
         /// <summary>
         /// Create a PublisherClient with custom batch thresholds.
         /// </summary>
-        public static PublisherClient GetCustomPublisher(string projectId,
-            string topicId)
+        public static async Task<PublisherClient> GetCustomPublisherAsync(
```
simonz130: Similar here and I guess in other places

pubsub/api/PubsubSample/Program.cs
```diff
@@ -268,41 +267,33 @@ public static object CreateTopic(string projectId, string topicId)
...
-                // Record all publishing Tasks. When each Task completes
-                // without error, the message has been successfully published.
-                // In real use don't simply store all publish Tasks forever;
-                // it is usually appropriate to keep a reference to the Task
-                // only until it has completed.
-                publishTasks.Add(publisher.PublishAsync(text));
-            }
-            foreach (var task in publishTasks)
+            var publishTasks =
+                messageTexts.Select(text => publisher.PublishAsync(text));
```
simonz130: I suspect you can just do:

```csharp
var publishTasks = messageTexts.Select(publisher.PublishAsync);
```

pubsub/api/PubsubSample/Program.cs
SurferJeffAtGoogle: When the code appears in the documentation, it won't have the full context of the function call.  In that context, I think it's more useful to show the user a call that awaits the PublisherClient.

pubsub/api/PubsubSample/Program.cs
SurferJeffAtGoogle: Same story.

pubsub/api/PubsubSample/Program.cs
SurferJeffAtGoogle: I tried it:
```
Program.cs(278,30): error CS0411: The type arguments for method 'Enumerable.Select<TSource, TResult>(IEnumerable<TSource>, Func<TSource, TResult>)' cannot be inferred from the usage. Try specifying the type arguments explicitly. [/usr/local/google/home/rennie/gitrepos/dotnet-docs-samples/pubsub/api/PubsubSample/Pubsub.csproj]
```

PR #706, opened 2018-12-06
Update some more dependencies.

SurferJeffAtGoogle:


dlp/api/DlpSample/DeIdentify.cs
```diff
@@ -19,6 +19,7 @@
 using System.IO;
 using static Google.Cloud.Dlp.V2.CryptoReplaceFfxFpeConfig.Types;
 using System.Linq;
+using Google.Api.Gax.ResourceNames;
```
simonz130: sort this

dlp/api/DlpSample/RiskAnalysis.cs
```diff
@@ -74,14 +75,14 @@ class RiskAnalysis
...
-                    ParentAsProjectName = new Google.Cloud.Dlp.V2.ProjectName(callingProjectId),
+                    ParentAsProjectName = new ProjectName(callingProjectId),
                     RiskJob = config
                 });
 
             // Listen to pub/sub for the job
             var subscriptionName = new SubscriptionName(callingProjectId, subscriptionId);
-            SubscriberClient subscriber = SubscriberClient.Create(
-                subscriptionName, new[] { SubscriberServiceApiClient.Create() });
+            SubscriberClient subscriber = SubscriberClient.CreateAsync(
```
simonz130: Did they change the API? Why did you change the original code? It's better to call sync method directly than call async method and then .Result

dlp/api/DlpSample/DeIdentify.cs
SurferJeffAtGoogle: Done.

dlp/api/DlpSample/RiskAnalysis.cs
SurferJeffAtGoogle: Yes, they changed the API.  I know it's better to await the result, but that would require making all, or most of the functions in the call stack async, which is overkill for this sample that otherwise does nothing asynchronous.

PR #718, opened 2018-12-18
Product search samples for C#

telpirion:
The Python canonical sample is here:
https://github.com/GoogleCloudPlatform/python-docs-samples/tree/master/vision/cloud-client/product_search

This pull request adds the following region tags:

- vision_product_search_import_product_images
- vision_product_search_add_product_to_product_set
- vision_product_search_list_products_in_product_set
- vision_product_search_remove_product_from_product_set
- vision_product_search_create_product
- vision_product_search_list_products
- vision_product_search_get_product
- vision_product_search_update_product_labels
- vision_product_search_delete_product
- vision_product_search_get_similar_products
- vision_product_search_get_similar_products_gcs
- vision_product_search_create_product_set
- vision_product_search_list_product_sets
- vision_product_search_get_product_set
- vision_product_search_delete_product_set
- vision_product_search_create_reference_image
- vision_product_search_list_reference_images
- vision_product_search_get_reference_image
- vision_product_search_delete_reference_image


vision/api/ProductSearch/ImportProductSets.cs
```diff
@@ -0,0 +1,71 @@
...
+using Google.Cloud.Vision.V1;
+using Google.LongRunning;
+using System;
+
+namespace GoogleCloudSamples
+{
+    [Verb("import_product_set", HelpText = "Import a product set from GCS")]
+    class ImportProductSetOptions : BaseOptions
+    {
+        [Value(2, HelpText = "Google Cloud Storage URI. Target files must be in Product Search CSV format.")]
```
beccasaurus: I don't remember what the integer here is used for. Just calling this value out, incase it shouldn't be 2 👍

telpirion: The integer value is the index of the argument passed into the command. BaseOptions provides indexes 0 (project ID) and 1 (region).

Do you think it would be better to use flags instead?

vision/api/ProductSearch/ImportProductSets.cs
```diff
@@ -0,0 +1,71 @@
...
+                // Set the input configuration along with Google Cloud Storage URI
+                InputConfig = new ImportProductSetsInputConfig
+                {
+                    GcsSource = new ImportProductSetsGcsSource
+                    {
+                        CsvFileUri = opts.GcsUri
+                    }
+                }
+            };
+            Operation<ImportProductSetsResponse, BatchOperationMetadata> response = client.ImportProductSets(request);
```
beccasaurus: Do the other existing LRO operation samples explicitly show the type?

I think we normally do this, which would be more idiomatic IMO (unless you're following an existing standard)

You can browse through some other examples in the repo:
https://github.com/GoogleCloudPlatform/dotnet-docs-samples/search?q=PollUntilCompleted&unscoped_q=PollUntilCompleted

```cs
var operation = client.ImportProductSets(request);

// Wait for operation to complete
var response = operation.PollUntilCompleted();

foreach (var status in response.Result.Statuses)
{
  // ...
}
```

telpirion: Here's the thing ... when I didn't explicitly set the type, the operations returned before they completed.

When I did explicitly set the type, they completed and then returned.

vision/api/VisionTest/ProductSearchTest.cs
```diff
@@ -0,0 +1,260 @@
...
+        public void TestProductSearchGcs()
+        {
+            _productSearch.Run("import_product_set", _projectId, REGION_NAME, CSV_GCS_URI);
+
+            var output = _productSearch.Run("get_similar_products_gcs", _projectId, REGION_NAME, PRODUCT_SET_ID, PRODUCT_CATEGORY, REF_IMAGE_GCS_URI, SEARCH_FILTER);
+            Assert.Equal(0, output.ExitCode);
+            Assert.Contains(PRODUCT_ID, output.Stdout);
+        }
+
+        public void CleanUp()
```
beccasaurus: Rename this `Dispose()` unless you're calling it from elsewhere

It's idiomatic for the setup code to go into the constructor and the teardown code to go into dispose

telpirion: Done.

vision/api/ProductSearch/ImportProductSets.cs
telpirion: ... and of course, now that I change back to var it works again.

Thank you!

PR #720, opened 2018-12-22
Convert Cloud KMS samples to GAPIC clients

bdhess:
cc @DanSanche

PR #726, opened 2019-01-11
Slightly simplify KMS request creation

chrisdunelm:
This is generally the recommended way to create request objects.


chrisdunelm:
Closing, as I've just seen #720, which is a better way to go.

PR #729, opened 2019-01-23
Vision no longer sees a 'mammal.' It sees a cat.

SurferJeffAtGoogle:
Change-Id: I915b2d260d2e551ad9d974739feccf6151757074

### DIFF
--- a/vision/api/VisionTest/VisionTest.cs
+++ b/vision/api/VisionTest/VisionTest.cs
@@ -110,7 +110,7 @@ namespace GoogleCloudSamples
         {
             var output = Run("labels", Path.Combine("data", "cat.jpg"));
             Assert.Equal(0, output.ExitCode);
-            Assert.Contains("mammal", output.Stdout);
+            Assert.Contains("cat", output.Stdout);
         }
 
         [Fact]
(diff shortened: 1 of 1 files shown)


vision/api/VisionTest/VisionTest.cs
```diff
@@ -258,31 +258,31 @@ protected override ConsoleOutput Run(params string[] args)
...
-
-            using (var collector = new BucketCollector(_bucketName))
-            {
-                collector.CopyToBucket(localPath, _pdfFileName);
+        {
+            var _pdfFileName = "HodgeConj.pdf";
+            var outputPrefix = "";
+            var localPath = Path.Combine("data", _pdfFileName);
+            var gcsSourceURI = $"gs://{_bucketName}/{_pdfFileName}";
+            ConsoleOutput output;
```
beccasaurus: Why is this section of the diff here?

_squinting_ my eyes to see a different 👀

SurferJeffAtGoogle: Just whitespace diff.  Not sure why.  Maybe the source code had \ns instead of \r\ns there.  Append ?w=1 to the github url to hide white space differences.  For example, https://github.com/GoogleCloudPlatform/dotnet-docs-samples/pull/729/files?w=1

PR #741, opened 2019-02-13
added gateway mgmt and its unit tests

munkhuushmgl:


iot/api/CloudIotTest/CloudIotTest.cs
```diff
@@ -235,46 +262,118 @@ public void TestListGateways()
...
+            var sendDataBoundDeviceOut = Run("sendDataFromBoundDevice", _fixture.ProjectId,
+                _fixture.RegionId, _fixture.RegistryId,
+            deviceId, gatewayId, _fixture.PrivateKeyPath, "RS256", _fixture.CertPath,
+            "state", "test-message");
+            Assert.Contains("Data sent", sendDataBoundDeviceOut.Stdout);
+            Assert.Contains("On Publish", sendDataBoundDeviceOut.Stdout);
+
+            Assert.DoesNotContain("An error occured", sendDataBoundDeviceOut.Stdout);
+
+            //Clean up
```
SurferJeffAtGoogle: Use try {} finally {} blocks to make sure that if an assertion fails, the clean up code still runs.

iot/api/CloudIotTest/CloudIotTest.cs
```diff
@@ -235,46 +262,118 @@ public void TestListGateways()
...
+                _fixture.RegionId, _fixture.RegistryId, gatewayId, deviceId, _fixture.CertPath,
+                _fixture.PrivateKeyPath, "RS256", "--listentime", "10");
+
+            //Assertions
+            Assert.DoesNotContain("error occurred", listenConfigMsgOut.Stdout);
+            Assert.Contains("On Subscribe", listenConfigMsgOut.Stdout);
+            Assert.Contains("On Publish", listenConfigMsgOut.Stdout);
+
+            //Clean up
+            Run("unbindDeviceFromGateway", _fixture.ProjectId, _fixture.RegionId,
```
SurferJeffAtGoogle: Use try {} finally {}.

iot/api/CloudIotSample/Program.cs
```diff
@@ -1376,6 +1491,209 @@ public static object ClearRegistry(string projectId, string cloudRegion, string
...
+        public static object AttachDevice(MqttClient client, string deviceId)
+        {
+            var attachTopic = $"/devices/{deviceId}/attach";
+            Console.WriteLine("Attaching: {0}", attachTopic);
+            var attachPayLoad = "{}";
+            var BinaryData = Encoding.UTF8.GetBytes(attachPayLoad);
+            client.Publish(attachTopic, BinaryData, MqttMsgBase.QOS_LEVEL_AT_LEAST_ONCE, true);
+
+            Console.WriteLine("Waiting for device to attach.");
+            System.Threading.Thread.Sleep(2000);
```
SurferJeffAtGoogle: Given the call to Sleep(), I would expect a later statement to confirm the the device attached.  Why sleep here?

iot/api/CloudIotSample/Program.cs
```diff
@@ -1376,6 +1491,209 @@ public static object ClearRegistry(string projectId, string cloudRegion, string
         }
         //[END iot_clear_registry]
 
+
+        //[START iot_attach_device]
+        public static object AttachDevice(MqttClient client, string deviceId)
```
SurferJeffAtGoogle: When a user sees MqttClient passed in as an argument, they get frustrated and we'll see bug reports like "How do I create a MqttClient object?!?"  Every sample should include the code to create the client.

iot/api/CloudIotSample/Program.cs
```diff
@@ -1376,6 +1491,209 @@ public static object ClearRegistry(string projectId, string cloudRegion, string
...
+            }
+            CloudIotMqttExample.SetupMqttTopics(mqttClient, gatewayId);
+            return mqttClient;
+        }
+
+        //[START iot_send_data_from_device]
+        public static object SendDataFromDevice(MqttClient client, string deviceId,
+            string messageType, string data)
+        {
+            if (!messageType.Equals("events") && !messageType.Equals("state"))
```
SurferJeffAtGoogle: always prefer == over .Equals.

iot/api/CloudIotTest/CloudIotTest.cs
munkhuushmgl: Ok will do

iot/api/CloudIotSample/Program.cs
munkhuushmgl: https://github.com/GoogleCloudPlatform/python-docs-samples/blob/1980f31828104d9c39ed4ee11239b1a95ce1f00c/iot/api-client/beta-features/gateway/gateway.py#L463-L465

munkhuushmgl: https://github.com/GoogleCloudPlatform/python-docs-samples/blob/1980f31828104d9c39ed4ee11239b1a95ce1f00c/iot/api-client/beta-features/gateway/gateway.py#L432-L442

this is original code which I tried to convert.

Do u think placing Sleep() right after AttachDevice() method will be better?

iot/api/CloudIotSample/Program.cs
munkhuushmgl: Will do for every method

PR #754, opened 2019-03-09
Video Intelligence text detection, object tracking

telpirion:
This pull request adds the following code snippets / region tags:
  * video_object_tracking_gcs
  * video_object_tracking
  * video_detect_text
  * video_detect_text_gcs

Python canonicals are here:
  * [Text detection](https://github.com/GoogleCloudPlatform/python-docs-samples/blob/master/video/cloud-client/analyze/beta_snippets.py#L83-L172)
  * [Object tracking](https://github.com/GoogleCloudPlatform/python-docs-samples/blob/master/video/cloud-client/analyze/beta_snippets.py#L176-L273)

### DIFF
--- a/video/api/Analyze/Program.cs
+++ b/video/api/Analyze/Program.cs
@@ -245,7 +245,7 @@ namespace GoogleCloudSamples.VideoIntelligence
             var startTime = textSegment.Segment.StartTimeOffset;
             var endTime = textSegment.Segment.EndTimeOffset;
             Console.Write(
-                $"Start time: {startTime.Seconds + startTime.Nanos * 1e9 },");
+                $"Start time: {startTime.Seconds + startTime.Nanos * 1e9 }, ");
             Console.WriteLine(
                 $"End time: {endTime.Seconds + endTime.Nanos * 1e9 }");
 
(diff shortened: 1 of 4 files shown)


video/api/Analyze/Program.cs
```diff
@@ -216,13 +224,225 @@ public static object TranscribeVideo(string uri)
...
+
+        // [START video_object_tracking_gcs]
+        public static object TrackObjectGcs(string gcsUri)
+        {
+            var client = VideoIntelligenceServiceClient.Create();
+            var request = new AnnotateVideoRequest
+            {
+                InputUri = gcsUri,
+                Features = { Feature.ObjectTracking },
+                LocationId = "us-east1"
```
beccasaurus: Add comment:
> It is recommended to use location_id as 'us-east1' for the best latency due to different types of processors used in this region and others.

From canonical. I never would've known this!

Unless the docs mention this above the sample. In which case it's not needed in the code (as this truth may change over time)

I do like having it in the code like the Python though

beccasaurus:
Can't find any test failures, it's all `Failures: 0` ~ log ends with `TIMED OUT`

Trying to run again...

Issue #748, opened 2019-03-03
The IOT tests are failing.


SurferJeffAtGoogle:
I disabled them so that the nightly build succeeds.

Please submit a fix for the tests before making any other changes to the IOT code.

munkhuushmgl:
I will fix it.

munkhuushmgl:
IoT Core registries are limited 100 registries. It has reached 101 registries that was why it was failing.

SurferJeffAtGoogle:
The tests are failing again:
![image](https://user-images.githubusercontent.com/11778530/54294686-e3ad1100-456e-11e9-9310-6a1ac7ae56c4.png)

munkhuushmgl:
I have added limit checker method for reaching 100 registries.
I tested it, and it should not fail due to overloaded registries.

https://github.com/GoogleCloudPlatform/dotnet-docs-samples/pull/760

Best
Mike

SurferJeffAtGoogle:
The last 8 test runs have failed:
![image](https://user-images.githubusercontent.com/11778530/54837876-ab02eb00-4c84-11e9-8e9e-daff73e86878.png)
![image](https://user-images.githubusercontent.com/11778530/54837948-cff75e00-4c84-11e9-8ddd-d01e69453c23.png)

munkhuushmgl:
Hey Jefft,

I found the bug, and fixed it. Looks like there was an update on an error
message which I expected in Assert.Contains().
Since they changed the error message, it was not passing that specific test.

It tested it, and tests should be good now
https://github.com/GoogleCloudPlatform/dotnet-docs-samples/pull/769

Mike

PR #759, opened 2019-03-13
Skip iot tests

SurferJeffAtGoogle:
They're broken.


dzlier-gcp:
Is there a tracking bug open to return to these at some point though? Cause otherwise they'll be lost in neverland.

SurferJeffAtGoogle:
The tracking bug is here: https://github.com/GoogleCloudPlatform/dotnet-docs-samples/issues/748

PR #763, opened 2019-03-16
storage: add Bucket Policy Only samples

frankyn:
Add Bucket Policy Only samples.
https://cloud.google.com/storage/docs/using-bucket-policy-only


frankyn:
Thanks @jsimonweb

PR #764, opened 2019-03-18
[Cloud Tasks] Add HTTP Push Queue Sample

averikitsch:
Test projects will have to be granted permission for HTTP targets or wait until whitelisting is off.


cloudtasks/api/HttpTasksSample/Samples.cs
```diff
@@ -0,0 +1,65 @@
...
+using Google.Protobuf;
+using Google.Protobuf.WellKnownTypes;
+using System;
+
+namespace GoogleCloudSamples
+{
+    public partial class Samples
+    {
+        // [START cloud_tasks_create_http_task]
+        public static object CreateTask(
```
SurferJeffAtGoogle: According to the new rubric, the START and END tags should enclose the using statements at the top and the complete class.

cloudtasks/api/HttpTasksSample/Samples.cs
```diff
@@ -0,0 +1,65 @@
...
+            string projectId,
+            string location,
+            string queue,
```
SurferJeffAtGoogle: According to the new rubric, each param should have a default value that looks like a real value the caller might pass in.  i.e.:
```
 string projectId = "YOUR-PROJECT-ID",
 string location = "global",
 string queue = "YOUR-QUEUE-ID",
```

cloudtasks/api/HttpTasksSample/Program.cs
```diff
@@ -0,0 +1,66 @@
...
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+using CommandLine;
+using System;
+
+namespace GoogleCloudSamples
```
SurferJeffAtGoogle: According to the new rubric, a command-line application is no longer necessary.  You can call the function directly from the test.

SurferJeffAtGoogle:
The tests failed.

averikitsch:
Looks like I rebased incorrectly. Will close and open cleaner PR.

PR #769, opened 2019-03-22
Error message got updated

munkhuushmgl:


iot/api/CloudIotTest/CloudIotTest.cs
```diff
@@ -235,7 +235,7 @@ public void TestSendCommand()
             }
             catch (Google.GoogleApiException e)
             {
-                Assert.Contains("not subscribed to the commands topic", e.Message);
+                Assert.Contains("not connected", e.Message);
```
SurferJeffAtGoogle: Is there a way to make the test less fragile, so if they change the text again, the test will keep passing?

munkhuushmgl: I'm not sure, I will look you into it and let you know

munkhuushmgl:
Looks like it is failing outside of IoT

PR #781, opened 2019-04-05
Combine all the IoT solutions and add all dependency projects

jskeet:
This still gives a lot of warnings due to dependency *packages* that only target .NET Framework instead of .NET Core, but the solution does open and build.


SurferJeffAtGoogle:
Code changes look fine.

Can you add a note to the PR description about what motivated the change?

jskeet:
Sure, will do tomorrow. I hadn't expected the PR to be merged - was more just to check we're on the same page, but happy to make it more useful :)

jskeet:
Whoops - was expecting this to be a different PR. Yup, will edit the commit message.

jskeet:
Updated PR and commit.

PR #783, opened 2019-04-08
Make video test less fragile.

SurferJeffAtGoogle:
Verify the output of the code in the samples, not the output of the API.

And extend the time out a little more.

Change-Id: I26da4db3af10ef2473dd97b30d82958d0cec8e5a


beccasaurus:
Ready to merge!

PR #785, opened 2019-04-09
[Cloud Tasks] Add HTTP Target Sample

averikitsch:
Test projects will have to be granted permission for HTTP targets or wait until whitelisting is off.


cloudtasks/api/TasksSample/CreateHttpTask.cs
```diff
@@ -0,0 +1,62 @@
...
+using Google.Cloud.Tasks.V2Beta3;
+using Google.Protobuf;
+using Google.Protobuf.WellKnownTypes;
+using System;
+
+class CreateHttpTask
+{
+    public string CreateTask(
+        string projectId = "YOUR-PROJECT-ID",
+        string location = "YOUR-LOCATION-ID",
```
SurferJeffAtGoogle: Is the location id usually something like "us-west1"?  In that case, I think we'll provide a better clue to the reader what the value should be if we show the value "us-west1" or the like instead of the more abstract "YOUR-LOCATION-ID".

cloudtasks/api/TasksSample/CreateHttpTask.cs
```diff
@@ -0,0 +1,62 @@
...
+        string queue = "YOUR-QUEUE-ID",
+        string url = "http://example.com",
+        string payload = "Hello World!",
+        int inSeconds = 0)
+    {
+        CloudTasksClient client = CloudTasksClient.Create();
+
+        QueueName parent = new QueueName(projectId, location, queue);
+
```
SurferJeffAtGoogle: This code has more blank lines than I'm used to seeing in C# code.

cloudtasks/api/TasksSample/CreateHttpTask.cs
```diff
@@ -0,0 +1,62 @@
...
+            Parent = parent.ToString(),
+            Task = new Task
+            {
+                HttpRequest = new HttpRequest
+                {
+                    HttpMethod = HttpMethod.Post,
+                    Url = url,
+                    Body = ByteString.CopyFromUtf8(payload)
+                },
+                ScheduleTime = new Timestamp
```
SurferJeffAtGoogle: ScheduleTime can be calculated with less code, in a way that's tough to explain with words, so I made the changes and committed them to this branch.

PR #790, opened 2019-04-10
added a check that allows only certain private key

munkhuushmgl:
Added a filter that will fix invalidcast exception which was caused by unsupported private keys,


munkhuushmgl:
Thanks @SurferJeffAtGoogle

PR #791, opened 2019-04-10
Product search test

SurferJeffAtGoogle:


SurferJeffAtGoogle:
telpirion, I tried updating the tests to make all the ids unique so that tests running simultaneously don't clobber each other.  I got it all working except for the Search tests, which fail and find no matching products.  Do you see why?

beccasaurus:
Going thru old PRs ~ @SurferJeffAtGoogle does this PR need to still exist? 🧹

PR #793, opened 2019-04-12
Deleted random mtd

munkhuushmgl:
In current existing code, register id was generated using random string generator class.
To make clean-up method work, all new registries need have unix timestamp post-fix.


SurferJeffAtGoogle:
Could you expand the PR description to define the term 'mtd' and explain _why_ you are making these changes?

SurferJeffAtGoogle:
Also, a tests is failing.

munkhuushmgl:
Could you expand the PR description to define the term 'mtd' and explain why you are making these changes? 

In current existing code, register id was generated using random string generator class. 
To make clean-up method work, all new registries need have unix timestamp post-fix.

PR #803, opened 2019-04-29
Added sample for using WebSocket with App Engine

michaelawyu:


SurferJeffAtGoogle:
I'll update the docker image so the test failure error message about the SDK version goes away.

[SignalR](https://docs.microsoft.com/en-us/aspnet/core/signalr/introduction?view=aspnetcore-2.2) is a [very popular](https://www.nuget.org/packages/Microsoft.AspNet.SignalR/) library for doing WebSockets with .NET core.  Could we also include a sample that uses SignalR?

appengine/flexible/WebSocket/app.yaml
```diff
@@ -0,0 +1,13 @@
+runtime: aspnetcore
```
andrewsg: Can we additionally include the commented-out session affinity line and associated comment documentation present in the Python sample? https://github.com/GoogleCloudPlatform/python-docs-samples/blob/db9218fa0a186db254cedc1fa5eb0b4980d69737/appengine/flexible/websockets/app.yaml#L17

appengine/flexible/WebSocket/README.md
```diff
@@ -0,0 +1,40 @@
...
+PS > dotnet publish
+PS > gcloud beta app deploy .\bin\Debug\netcoreapp2.1\publish\app.yaml
+```
+
+## Using the Sample
+
+The sample app listens for WebSocket connections at http://localhost:5000/ws
+(when running locally) or https://YOUR-PROJECT-ID.appspot.com/ws
+(when deployed to App Engine). It echos every incoming WebSocket message.
+You may [send a WebSocket message](https://developer.mozilla.org/en-US/docs/Web/API/WebSocket)
```
andrewsg: I'm confused about this; the canonical sample serves an HTTP page that allows the user to interact with the websockets server directly, but is that not possible in this case? Also, if the user is supposed to interact with websockets via the API, why is a .js file included in this app? Apologies for the lack of context around C# websockets

michaelawyu:
@andrewsg Hi Andrew! Sorry for the confusion. I have updated the sample. As for the `.js` file, if I understand it correctly it is for testing purposes, as seen in the [HelloWorld sample](https://github.com/GoogleCloudPlatform/dotnet-docs-samples/blob/master/appengine/flexible/HelloWorld/test.js).

@SurferJeffAtGoogle I will send a separate PR for using SignaIR :)

appengine/flexible/WebSocket/README.md
michaelawyu: Updated. Please see the comment below.

appengine/flexible/WebSocket/app.yaml
michaelawyu: Fixed.

andrewsg:
Thank you, Chen!

PR #808, opened 2019-05-03
Social auth again

SurferJeffAtGoogle:
Uses new kms libraries for dataprotection.


appengine/flexible/SocialAuth/Set-Up.ps1
```diff
@@ -0,0 +1,146 @@
...
+    }
+    return $keyName
+}
+
+<#
+.SYNOPSIS
+Grants a role to the App Engine default service account for a given key ring.
+
+.PARAMETER keyRingId
+The key ring to grant the role.
```
jskeet: This sentence doesn't work for me at the moment. How about:

> The key ring that the new role should grant access to.

appengine/flexible/SocialAuth/Startup.cs
```diff
@@ -92,8 +92,17 @@ public void ConfigureServices(IServiceCollection services)
...
-            services.AddSingleton<IDataProtectionProvider>(serviceProvider =>
-                new KmsDataProtectionProvider(ProjectId, "global", "dataprotectionprovider"));
+            // Add Google libraries for data protection.
+            services.AddDataProtection()
+                // Store keys in Cloud Storage so that multiple instances
+                // of the web application see the same keys.
+                .PersistKeysToGoogleCloudStorage(
+                    Configuration["DataProtection:Bucket"],
+                    Configuration["DataProtection:Object"])
+                // Protect the keys with Google KMS for encryption and fine-
```
jskeet: Move fine- onto the next line, to avoid the pseudo-break?

SurferJeffAtGoogle: Done.

appengine/flexible/SocialAuth/Set-Up.ps1
SurferJeffAtGoogle: Done.

PR #809, opened 2019-05-03
Some video files on GCS have been removed.  Use new ones.

SurferJeffAtGoogle:


video/api/Test/Tests.cs
```diff
@@ -96,7 +96,7 @@ void TestShotsGcs()
         void TestLabels()
         {
             ConsoleOutput output = _analyze.Run("labels",
-                DownloadGcsObject(@"gs://demomaker/cat.mp4"));
+                DownloadGcsObject(@"gs://cloud-samples-data/video/cat.mp4"));
```
simonz130: Can we make this a const so we update it only once in all tests in case it changes in the future?

PR #811, opened 2019-05-06
Added sample for using WebSocket (SignaIR) with App Engine Flexible E…

michaelawyu:
…nvironment.


appengine/flexible/WebSocketSignalR/wwwroot/js/chat.js
```diff
@@ -0,0 +1,38 @@
+"use strict";
```
SurferJeffAtGoogle: I think this file needs a copyright statement too.

michaelawyu: Added. PTAL :)

PR #813, opened 2019-05-09
Add V4 signing samples

JesseLovelace:
V4 put and get samples

### DIFF
--- a/storage/api/Storage/Storage.cs
+++ b/storage/api/Storage/Storage.cs
@@ -27,7 +27,7 @@ namespace GoogleCloudSamples
 {
     public class Storage
     {
-        private static readonly string s_projectId = "YOUR-PROJECT-ID";
+        private static readonly string s_projectId = "YOUR-PROJECT";
 
         private static readonly string s_usage =
             "Usage: \n" +
@@ -44,6 +44,8 @@ namespace GoogleCloudSamples
             "  Storage download [-key encryption-key] bucket-name object-name [local-file-path]\n" +
             "  Storage download-byte-range bucket-name object-name range-begin range-end [local-file-path]\n" +
             "  Storage generate-signed-url bucket-name object-name\n" +
+            "  Storage generate-signed-get-url-v4 bucket-name object-name\n" +
+            "  Storage generate-signed-put-url-v4 bucket-name object-name\n" +
             "  Storage view-bucket-iam-members bucket-name\n" +
             "  Storage add-bucket-iam-member bucket-name member\n" +
             "  Storage remove-bucket-iam-member bucket-name role member\n" +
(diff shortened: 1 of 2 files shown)


storage/api/Storage/Storage.cs
```diff
@@ -705,6 +707,43 @@ private void GenerateSignedUrl(string bucketName, string objectName)
         }
         // [END storage_generate_signed_url]
 
+        // [START storage_generate_signed_url_v4]
+        private void GenerateV4SignedGetUrl(string bucketName, string objectName)
+        {
+            UrlSigner urlSigner = UrlSigner
+            .FromServiceAccountPath(Environment.GetEnvironmentVariable("GOOGLE_APPLICATION_CREDENTIALS"))
```
jskeet: I'd indent these lines by an extra four spaces if possible. If that makes the line too long, you could fetch the credentials path into a separate variable first.

storage/api/Storage/Storage.cs
```diff
@@ -705,6 +707,43 @@ private void GenerateSignedUrl(string bucketName, string objectName)
...
+        {
+            UrlSigner urlSigner = UrlSigner
+            .FromServiceAccountPath(Environment.GetEnvironmentVariable("GOOGLE_APPLICATION_CREDENTIALS"))
+            .WithSigningVersion(SigningVersion.V4);
+            string url =
+                urlSigner.Sign(bucketName, objectName, TimeSpan.FromHours(1), HttpMethod.Get);
+            Console.WriteLine("Generated GET signed URL:");
+            Console.WriteLine(url);
+            Console.WriteLine("You can use this URL with any user agent, for example:");
+            Console.WriteLine("curl '" + url + "'");
```
jskeet: I'd personally use an interpolated string literal here: `Console.WriteLine($"curl '{url}'");` - and likewise at line 742. But it's not a big deal.

storage/api/Storage/Storage.cs
```diff
@@ -705,6 +707,43 @@ private void GenerateSignedUrl(string bucketName, string objectName)
...
+        // [END storage_generate_signed_url_v4]
+
+        // [START storage_generate_upload_signed_url_v4]
+        private void GenerateV4SignedPutUrl(string bucketName, string objectName)
+        {
+            UrlSigner urlSigner = UrlSigner
+            .FromServiceAccountPath(Environment.GetEnvironmentVariable("GOOGLE_APPLICATION_CREDENTIALS"))
+            .WithSigningVersion(SigningVersion.V4);
+
+            Dictionary<string, IEnumerable<string>> contentHeaders =
```
jskeet: Unless we're particular about not using `var`, I think it would make this code simpler to read:

```csharp
var contentHeaders = new Dictionary<string, IEnumerable<string>>
{
    { "Content-Type", new[] { "text/plain" } }
};
```

storage/api/Storage/Storage.cs
```diff
@@ -705,6 +707,43 @@ private void GenerateSignedUrl(string bucketName, string objectName)
         }
         // [END storage_generate_signed_url]
 
+        // [START storage_generate_signed_url_v4]
+        private void GenerateV4SignedGetUrl(string bucketName, string objectName)
+        {
+            UrlSigner urlSigner = UrlSigner
+            .FromServiceAccountPath(Environment.GetEnvironmentVariable("GOOGLE_APPLICATION_CREDENTIALS"))
+            .WithSigningVersion(SigningVersion.V4);
+            string url =
```
jskeet: I'm assuming the line break is to keep the sample line length short?

SurferJeffAtGoogle:
Please run `dotnet format` on the new code to fix Lint errors.

storage/api/Storage/Storage.cs
SurferJeffAtGoogle: +1

storage/api/Storage/Storage.cs
SurferJeffAtGoogle: +1

PR #822, opened 2019-05-21
Spanner: Use correct/consistent logic to transfer budget.

hegemonic:
The `WriteWithTransactionUsingDmlCoreAsync` snippet moves $200,000 from Album1 to Album2. However, Album1 is initialized with a budget of $100,000. As a result, you can't work through the [getting started doc](https://cloud.google.com/spanner/docs/getting-started/csharp/) from start to finish. This snippet also defined and checked a `minimumAmountToTransfer`, which isn't consistent with the equivalent snippets for other languages.

In addition, the `ReadWriteWithTransactionCoreAsync` snippet moved $200,000, but in the opposite direction, which was somewhat confusing.

Finally, neither snippet confirmed that the source album's budget was at least as large as the transfer amount.

I made both snippets transfer $100,000 from Album1 to Album2. I also fixed the logic errors and updated the tests as needed.

### DIFF
--- a/spanner/api/Spanner/Program.cs
+++ b/spanner/api/Spanner/Program.cs
@@ -1026,15 +1026,15 @@ namespace GoogleCloudSamples.Spanner
             string instanceId,
             string databaseId)
         {
-            // This sample transfers 200,000 from the MarketingBudget
-            // field of the second Album to the first Album. Make sure to run
+            // This sample transfers 100,000 from the MarketingBudget
+            // field of the first Album to the second Album. Make sure to run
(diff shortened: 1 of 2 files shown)


hegemonic:
Looks like the Cloud Firestore tests are timing out. Any suggestions for getting a clean CI build?

SurferJeffAtGoogle:
The test failures are unrelated to your code changes.  We'll be happy to merge with firestore tests failing, after the code is reviewed.

hegemonic:
@jsimonweb is sending a new PR to update this code, with some changes from my version, so I'm closing this PR.

Issue #425, opened 2018-01-15
NAudio has conflict with speech api samples 


heathyates:
# Steps to Reproduce 

1.  Clone the repo and download 
2. Find the speech `sln` and load it 
3. When VS rebuilds the solution, you should the following error: 

`Package 'NAudio 1.8.4' was restored using '.NETFramework,Version=v4.6.1' instead of the project target framework '.NETCoreApp,Version=v2.0'. This package may not be fully compatible with your project.`

# NAudio and .NET 2.0 seems to be the issue 

Please see this issue on NAudio as follows. Please see [here](https://github.com/naudio/NAudio/issues/285).

SurferJeffAtGoogle:
Thank you for reporting this issue.

Were you able to successfully run the sample?  On Windows, Mac, or Linux?

mofolom:
Bump.

Did the same. Cloned the repo and tried to rebuild the solution.

SurferJeffAtGoogle:
mofolom, could you reply with your build output?

shaikhnizam:
I'm also facing same error while trying to resolve Nuget issue. Any solutions??

warning NU1701: Package 'NAudio 1.8.4' was restored using '.NETFramework,Version=v4.6.1' instead of the project target framework '.NETCoreApp,Version=v2.0'. This package may not be fully compatible with your project.

dfdumaresq:
I'm having the same issue.

helloneoberg:
You may wish to refer to this link: https://markheath.net/post/naudio-netstandard
Look under "Try it out!" section as it guide you to use the preview package.
The sample i downloaded is working now with the preview package from NuGet.
Hope it help..

SurferJeffAtGoogle:
I see the same warning:

```
$ dotnet test
/usr/local/google/home/rennie/gitrepos/dotnet-docs-samples/speech/api/Recognize/Recognize.csproj : warning NU1701: Package 'NAudio 1.8.5' was restored using '.NETFramework,Version=v4.6.1' instead of the project target framework '.NETCoreApp,Version=v2.0'. This package may not be fully compatible with your project. [/usr/local/google/home/rennie/gitrepos/dotnet-docs-samples/speech/api/SpeechTest/SpeechTest.csproj]
[...]

PR #831, opened 2019-06-06
Upgrade NAudio to 1.9.

SurferJeffAtGoogle:
Fixes https://github.com/GoogleCloudPlatform/dotnet-docs-samples/issues/425

### DIFF
--- a/speech/api/Recognize/Recognize.cs
+++ b/speech/api/Recognize/Recognize.cs
@@ -511,11 +511,6 @@ namespace GoogleCloudSamples
         // [START speech_transcribe_streaming_mic]
         static async Task<object> StreamingMicRecognizeAsync(int seconds)
         {
-            if (NAudio.Wave.WaveIn.DeviceCount < 1)
-            {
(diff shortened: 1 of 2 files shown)


speech/api/Recognize/Recognize.cs
```diff
@@ -511,11 +511,6 @@ static async Task<object> StreamingRecognizeAsync(string filePath)
         // [START speech_transcribe_streaming_mic]
         static async Task<object> StreamingMicRecognizeAsync(int seconds)
         {
-            if (NAudio.Wave.WaveIn.DeviceCount < 1)
```
gguuss: LGTM, on first read I assumed something changed in 1.9.0 that broke this logic. Did we want to explicitly remove the check for an available Mic / does this affect automation?

SurferJeffAtGoogle: The DeviceCount property no longer exists in 1.9.  I didn't want to change the logic, but I had to.

PR #847, opened 2019-06-25
Improvements for samples of job polling

jskeet:
- Always assign back to a variable
- Always throw on any error

The return value of PollUntilCompleted() is important; ignoring it leads to issues such as https://github.com/googleapis/google-cloud-dotnet/issues/3169#issuecomment-505347038

The samples now always use the return value; I'm going to update the XML documentation as well. Any suggestions for further action would be welcome too.


SurferJeffAtGoogle:
The linter choked on some white space formatting.  Running [`dotnet format`](https://github.com/dotnet/format) should fix it.

jskeet:
Great, thanks - fixed now, PTAL.

PR #856, opened 2019-07-27
[Storage] Support UniformBucketLevelAccess

frankyn:
Update name of BucketPolicyOnly to use UniformBucketLevelAccess.
- [x] Fix pending for backend bug: 134695162. Remove workaround.
- [x] Verify integration tests pass.
- [x] C# client updates dep on Apiary client that supports UBLA field.


storage/api/Storage/Storage.cs
```diff
@@ -999,60 +999,66 @@ private void DisableBucketDefaultEventBasedHold(string bucketName)
...
 
-        // [START storage_enable_bucket_policy_only]
-        private void EnableBucketPolicyOnly(string bucketName)
+        // [START storage_enable_uniform_bucket_level_access]
+        private void EnableUniformBucketLevelAccess(string bucketName)
         {
             var storage = StorageClient.Create();
             var bucket = storage.GetBucket(bucketName);
+            bucket.IamConfiguration.UniformBucketLevelAccess.Enabled = true;
+            /** THIS IS A WORKAROUND */
```
SurferJeffAtGoogle: Include a link to the issue it's working around.

SurferJeffAtGoogle: When the issue becomes public.

PR #868, opened 2019-08-15
Add `fs_update_document_increment` snippet

BenWhitehead:


SurferJeffAtGoogle:
The test failed.  Not clear why.  Does it succeed when you run it locally?

https://source.cloud.google.com/results/invocations/999d20cc-02fe-4e74-a34b-6d880608d019/targets/github%2Fdotnet-docs-samples%2Ffirestore%2Fapi%2FFirestoreTest/tests

BenWhitehead:
I believe I've found the issue. I was incorrectly asserting against the output of a different test instead of the one I added. I've fixed the typo and pushed a new commit.

BenWhitehead:
@SurferJeffAtGoogle It looks like the Cloud KMS tests timed out while running. Is that test suite know to be flaky?

PR #869, opened 2019-08-16
Remove unnecessary `Exists` checks.

mikelehen:
A QuerySnapshot only consists of existing documents, so there's no need to check DocumentSnapshot.Exists.

### DIFF
--- a/firestore/api/GetData/Program.cs
+++ b/firestore/api/GetData/Program.cs
@@ -154,20 +154,13 @@ Where command is one of
             QuerySnapshot capitalQuerySnapshot = await capitalQuery.GetSnapshotAsync();
             foreach (DocumentSnapshot documentSnapshot in capitalQuerySnapshot.Documents)
             {
-                if (documentSnapshot.Exists)
+                Console.WriteLine("Document data for {0} document:", documentSnapshot.Id);
+                Dictionary<string, object> city = documentSnapshot.ToDictionary();
+                foreach (KeyValuePair<string, object> pair in city)
                 {
-                    Console.WriteLine("Document data for {0} document:", documentSnapshot.Id);
-                    Dictionary<string, object> city = documentSnapshot.ToDictionary();
-                    foreach (KeyValuePair<string, object> pair in city)
-                    {
-                        Console.WriteLine("{0}: {1}", pair.Key, pair.Value);
-                    }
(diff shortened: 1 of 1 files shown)


mikelehen:
Note I am unable to (conveniently) compile / run this change because I don't have a Windows machine, but I'll be extremely disappointed in myself if this is not a correct change. 😄

firestore/api/GetData/Program.cs
```diff
@@ -154,20 +154,13 @@ private static async Task GetMultipleDocs(string project)
...
-                    Console.WriteLine("Document data for {0} document:", documentSnapshot.Id);
-                    Dictionary<string, object> city = documentSnapshot.ToDictionary();
-                    foreach (KeyValuePair<string, object> pair in city)
-                    {
-                        Console.WriteLine("{0}: {1}", pair.Key, pair.Value);
-                    }
-                    Console.WriteLine("");
-                }
-                else
+                Console.WriteLine("Document data for {0} document:", documentSnapshot.Id);
```
SurferJeffAtGoogle: I don't know this API well.  Can you explain why it's unnecessary to check Exists in this scenario?

jskeet: These are document snapshots that have been returned from a query. Non-existent snapshots are only returned if you call GetDocumentSnapshot on a DocumentReference where the document doesn't exist.

mikelehen: Sure.  

DocumentSnapshot is used to represent a document read from Firestore, but if you are reading a document directly, it may not exist, e.g.:
```
var docSnap = Tasks.Await(db.Document("nonexisting/document").GetAsync());
Console.Out.WriteLine(docSnap.Exists); // False
```

But when you execute any sort of _query_ against Firestore, you'll get a `QuerySnapshot` containing all of the `DocumentSnapshot`s of the matching documents.  Since a document must exist in order to match the query, it is unnecessary to check `DocumentSnapshot.Exists` on the matching documents.

And if you look at the docs, all of the other languages (other than C#) don't have an exists check for this sample: https://firebase.google.com/docs/firestore/query-data/get-data

SurferJeffAtGoogle: Thank you for clarifying, and thank you for improving the sample.

PR #873, opened 2019-08-20
Add RemoveMember.cs, make AddMember a static method

mdejong3:
I'm updating this doc (https://cloud.google.com/iam/docs/granting-changing-revoking-access) and thought it could benefit from code samples for removing members using the client libraries, so I wrote a RemoveMember.cs code sample.

I also made AddMember a static method to match all of the other methods in this sample.


mdejong3:
@googlebot I signed it!

iam/api/Access/RemoveMember.cs
```diff
@@ -0,0 +1,32 @@
...
+// limitations under the License.
+
+// [START iam_modify_policy_remove_member]
+
+using System.Linq;
+using Google.Apis.CloudResourceManager.v1.Data;
+
+public partial class AccessManager
+{
+    public static Policy RemoveMember(Policy policy, string role, string member)
```
SurferJeffAtGoogle: Please add a test that exercises the new code.

mdejong3: Added tests

mdejong3:
I haven't really pushed to GitHub much--could you help me understand why these checks are failing and what I need to do to fix it?

Thanks!

mdejong3:
Ready to be checked again!

mdejong3:
I think the windows tests are failing due to linting, but the linux test is still failing on AccessTest--which is unfortunate, because it's still working on my machine. I put in some additional conditions that should give more informative errors, but I can't find them on the invocation log. Would you be able to run the test on your machine and tell me what error it gives you?

SurferJeffAtGoogle:
The test is failing with a permissions issue: 
Failed   GoogleCloudSamples.AccessTest.TestAccess
Error Message:
 Google.GoogleApiException : Google.Apis.Requests.RequestError
The caller does not have permission [403]
Errors [
        Message[The caller does not have permission] Location[ - ] Reason[forbidden] Domain[global]
]

mdejong3:
...oops. I left my test project ID in the _project variable. I removed it--should (hopefully) work now, or at least give a more informative error.

mdejong3:
This is still passing for me, locally. Any idea what's going wrong?

PR #875, opened 2019-08-22
Remove direct dependencies on gRPC because

SurferJeffAtGoogle:
they're redundant and make the build more fragile.

Change-Id: I4c4b8d05f66b3cc0961d6587585f62ec0fd03dff


jskeet:
Not sure about this. The redundant Grpc.Auth dependency should be fine to remove, but specifying Grpc.Core in a way that *looks* redundant can often be important. If an app depends directly on a project that depends on Grpc.Core, all the native libraries are handled appropriately. If it's via a longer transitive dependency chain, that doesn't work in all cases. This is why our client libraries all directly depend on Grpc.Core.

That *may* not be a problem in these cases, but it's worth making absolutely sure.

PR #883, opened 2019-09-09
Infinite streaming sample for Speech

jskeet:


speech/api/Recognize/InfiniteStreaming.cs
```diff
@@ -0,0 +1,248 @@
...
+        /// </summary>
+        private SpeechClient.StreamingRecognizeStream _rpcStream;
+
+        /// <summary>
+        /// The deadline for when we should stop the current stream.
+        /// </summary>
+        private DateTime _rpcStreamDeadline;
+
+        /// <summary>
+        /// The task indicating when the next response is ready.
```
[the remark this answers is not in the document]
jskeet: That's fair. Will need to update the comment to make it clearer too. Will try to do that some time next week - thanks for catching it.

speech/api/Recognize/InfiniteStreaming.cs
```diff
@@ -0,0 +1,248 @@
...
+            _moveNextTask = _rpcStream.ResponseStream.MoveNext();
+            await _rpcStream.WriteAsync(new StreamingRecognizeRequest
+            {
+                StreamingConfig = new StreamingRecognitionConfig
+                {
+                    Config = new RecognitionConfig
+                    {
+                        Encoding = RecognitionConfig.Types.AudioEncoding.Linear16,
+                        SampleRateHertz = SampleRate,
+                        LanguageCode = "en-US",
```
munkhuushmgl: Could we take language code from user input [ as parameter ]? my suggestion

jskeet: While that would be entirely reasonable, I think I'd prefer to do that for all the Speech samples rather than make this one different from everything else. I'd like to get this one in first, then we can consider updating everything to allow a language code to be specified.

speech/api/Recognize/InfiniteStreaming.cs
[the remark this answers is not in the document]
jskeet: Done - PTAL.

speech/api/Recognize/InfiniteStreaming.cs
munkhuushmgl: Okay sounds good to me

PR #889, opened 2019-10-09
Fix vision test

SurferJeffAtGoogle:


vision/api/VisionTest/ProductSearchTest.cs
```diff
@@ -1,318 +1,365 @@
...
-        private const string PRODUCT_ID_2 = "fake_product_id_for_testing_2";
-        private const string PRODUCT_SET_ID = "fake_product_set_id_for_testing";
-        private const string PRODUCT_SET_DISPLAY_NAME = "fake_product_set_display_name_for_testing";
-        private const string REF_IMAGE_ID = "fake_ref_image_id";
-        private const string REF_IMAGE_GCS_URI = "gs://cloud-samples-data/vision/product_search/shoes_1.jpg";
-        private const string CSV_GCS_URI = "gs://cloud-samples-data/vision/product_search/product_sets.csv";
-        private const string IMAGE_URI_1 = "shoes_1.jpg";
-        private const string IMAGE_URI_2 = "shoes_2.jpg";
-        private const string SEARCH_FILTER = "style=womens";
+        public string REGION_NAME { get; private set; } = "us-west1";
```
jskeet: Any reason these are in SHOUTY_CASE? Would prefer RegionName etc.

It feels like not all of these need to be instance properties, and not all need to be settable. Would be good to clarify that.

vision/api/VisionTest/ProductSearchTest.cs
```diff
@@ -1,318 +1,365 @@
...
+        /// </summary>
+        protected void DeleteCreations()
+        {
+            // Clean up everything the test created.
+            _createCommands.Reverse();
+            var exceptions = new List<Exception>();
+            foreach (string[] arguments in _createCommands)
             {
-                _productSearch.Run("create_product", _projectId, REGION_NAME, INDEXED_PRODUCT_1, PRODUCT_DISPLAY_NAME, PRODUCT_CATEGORY);
+                var deleteCommand = arguments.ToList();
```
amanda-tarafa: Why ToList() here and then back ToArray() again?

vision/api/VisionTest/ProductSearchTest.cs
```diff
@@ -1,318 +1,365 @@
...
 
-            output = _productSearch.Run("list_products", _projectId, REGION_NAME);
-            if (!output.Stdout.Contains(INDEXED_PRODUCT_1))
+        /// <summary>
+        /// Delete all the things created in Run() commands.
+        /// </summary>
+        protected void DeleteCreations()
+        {
+            // Clean up everything the test created.
+            _createCommands.Reverse();
```
amanda-tarafa: I know this is a test class not to be used from outside, but if this method were to be called multiple times, the list would be reversed and back multiple times. Also, the same deletions would be tried several times. Consider copying _createCommands at the start of this method and clearing _createCommands right after copying, and work with this copy in this method? (Suppose someone wants to use this method from within a test several times)

vision/api/VisionTest/ProductSearchTest.cs
```diff
@@ -1,318 +1,365 @@
...
+            Run("purge_orphan_products", PROJECT_ID, REGION_NAME);
 
-            output = _productSearch.Run("list_products", _projectId, REGION_NAME);
+            output = Run("list_products", PROJECT_ID, REGION_NAME);
             Assert.DoesNotContain(String.Format("Product id: {0}", PRODUCT_ID), output.Stdout);
         }
+    }
 
-        public void Dispose()
+    // These tests all require products and indexes that live longer than the
```
amanda-tarafa: If I understand correctly what you are doing here, with XUnit this is usually done using [Class Fixtures](https://xunit.net/docs/shared-context#class-fixture).

vision/api/VisionTest/ProductSearchTest.cs
SurferJeffAtGoogle: I converted from SHOUTY_CASE to TitleCase.

Figuring out the exact minimal visibility required for each variable would be ideal, but just doesn't seem like a valuable use of my time.

vision/api/VisionTest/ProductSearchTest.cs
SurferJeffAtGoogle: Fixed.

vision/api/VisionTest/ProductSearchTest.cs
SurferJeffAtGoogle: Fixed.

vision/api/VisionTest/ProductSearchTest.cs
SurferJeffAtGoogle: Actually, the resources created here need live beyond the whole test program.  It's a quirk of the API.  I think the indexes take a while to build.  So a class fixture would actually tear everything down too soon.

vision/api/VisionTest/ProductSearchTest.cs
```diff
@@ -1,318 +1,367 @@
...
-            }
+        /// <summary>
+        /// Delete all the things created in Run() commands.
+        /// </summary>
+        protected void DeleteCreations()
+        {
+            // Clean up everything the test created.
+            List<string[]> commands = new List<string[]>(_createCommands);
+            _createCommands.Clear();
+            commands.Reverse();
```
simonz130: nit: you could just use Stack instead of list and reverse.

vision/api/VisionTest/VisionTest.cs
```diff
@@ -118,7 +118,7 @@ public void DetectLandmark()
         {
             var output = Run("landmarks", Path.Combine("data", "tower.jpg"));
             Assert.Equal(0, output.ExitCode);
-            Assert.Contains("Eiffel", output.Stdout);
+            Assert.Matches(@".*(Eiffel|Mars).*", output.Stdout);
```
gguuss: /FYI I have found some of these label results to change, sometimes even to empty, and it still being valid.

PR #893, opened 2019-10-22
Add 3rd shard for Windows builds too.

SurferJeffAtGoogle:
Change-Id: I50d6c9fe2b3eb95ca0b6463fef1c84052d317755


.kokoro-windows/trampoline3.bat
```diff
@@ -0,0 +1,14 @@
+REM Copyright 2017 Google Inc.
```
amanda-tarafa: super nit: 2019

PR #895, opened 2019-10-25
Update Program.cs

therronjordan:
Default=null on automatically loaded credential information? Just doesn't seem right

### DIFF
--- a/auth/AuthSample/Program.cs
+++ b/auth/AuthSample/Program.cs
@@ -31,7 +31,7 @@ namespace GoogleCloudSamples
     class BaseOptions
     {
         string _projectId;
-        [Option('p', Default = null, HelpText = "Your Google project id.")]
+        [Option('p', HelpText = "Your Google project id.")]
         public string ProjectId
         {
             get


SurferJeffAtGoogle:
Without a signed CLA, I can't merge.

therronjordan:
Sorry, didn't realize. Done.

amanda-tarafa:
@therronjordan It still seems like you haven't signed the CLA, I just rechecked.
Anyway, I'll just be closing this PR now, see my comment on code.

auth/AuthSample/Program.cs
```diff
@@ -31,7 +31,7 @@ namespace GoogleCloudSamples
     class BaseOptions
     {
         string _projectId;
-        [Option('p', Default = null, HelpText = "Your Google project id.")]
+        [Option('p', HelpText = "Your Google project id.")]
```
amanda-tarafa: All of the other options in this file have mostly null default values. This is mostly right since we need the projectId to be provided by the caller. Setting this default at null explicitly is mostly to make clear to the caller that they need to specify their own projectId.

PR #907, opened 2019-11-11
Storage quickstart now disposes of StorageClient

linjmeyer:
`StorageClient` is disposable, this PR adds a `using` statement to ensure it is properly disposed of.    In this particular app it isn't an issue but if, like me, you copy this code into a high throughput application you will eventually get exceptions due to `HttpClient` running out of available TCP ports.


linjmeyer:
@googlebot I signed it!

PR #912, opened 2019-12-03
Fix lint error in Storage Quickstart.

amanda-tarafa:


amanda-tarafa:
Will wait on green and merge myself, given that the change is extremely trivial.

PR #915, opened 2019-12-04
docs(firestore): add samples for distributed counters

denis-ovsyannikov:
Previously made samples for [Go](https://github.com/GoogleCloudPlatform/golang-samples/pull/959) and [Python](https://github.com/GoogleCloudPlatform/python-docs-samples/pull/2412) are already merged and posted to Distributed Counters page, so, as promised, I'm pushing similar samples for C#.

### DIFF
--- a/firestore/api/FirestoreTest/FirestoreTest.cs
+++ b/firestore/api/FirestoreTest/FirestoreTest.cs
@@ -230,8 +230,8 @@ namespace GoogleCloudSamples
         {
             return _manageIndexes.Run(args);
         }
-
-		readonly CommandLineRunner _distrubutedCounter = new CommandLineRunner()
+        
+        readonly CommandLineRunner _distrubutedCounter = new CommandLineRunner()
         {
             VoidMain = DistributedCounter.Main,
             Command = "dotnet run"
@@ -241,7 +241,7 @@ namespace GoogleCloudSamples
         {
             return _distrubutedCounter.Run(args);
         }
-		
+
         // QUICKSTART TESTS
         [Fact]
         public void InitializeProjectIdTest()
@@ -840,8 +840,8 @@ namespace GoogleCloudSamples
             RunQueryData("query-create-examples", Environment.GetEnvironmentVariable("FIRESTORE_PROJECT_ID"));
             RunPaginateData("multiple-cursor-conditions", Environment.GetEnvironmentVariable("FIRESTORE_PROJECT_ID"));
         }
-		
-		[Fact]
+        
+        [Fact]
         public void RunDistributedCounterTest()
         {
             var output = RunDistributedCounter("run-distributed-counter", Environment.GetEnvironmentVariable("FIRESTORE_PROJECT_ID"));
--- a/firestore/api/FirestoreTest/FirestoreTest.cs
+++ b/firestore/api/FirestoreTest/FirestoreTest.cs
@@ -231,6 +231,17 @@ namespace GoogleCloudSamples
             return _manageIndexes.Run(args);
         }
 
+		readonly CommandLineRunner _distrubutedCounter = new CommandLineRunner()
+        {
+            VoidMain = DistributedCounter.Main,
+            Command = "dotnet run"
+        };
+
+        protected ConsoleOutput RunDistributedCounter(params string[] args)
+        {
+            return _distrubutedCounter.Run(args);
+        }
+		
         // QUICKSTART TESTS
         [Fact]
         public void InitializeProjectIdTest()
@@ -828,6 +839,16 @@ namespace GoogleCloudSamples
             Assert.Contains("Index creation completed!", manageIndexesOutput.Stdout);
             RunQueryData("query-create-examples", Environment.GetEnvironmentVariable("FIRESTORE_PROJECT_ID"));
             RunPaginateData("multiple-cursor-conditions", Environment.GetEnvironmentVariable("FIRESTORE_PROJECT_ID"));
+        }
+		
+		[Fact]
+        public void RunDistributedCounterTest()
+        {
+            var output = RunDistributedCounter("run-distributed-counter", Environment.GetEnvironmentVariable("FIRESTORE_PROJECT_ID"));
+            Assert.Contains("Created Cloud Firestore client with project ID", output.Stdout);
+            Assert.Contains("Distributed counter initialized.", output.Stdout);
+            Assert.Contains("Distributed counter incremented.", output.Stdout);
+            Assert.DoesNotContain("Total count: 0", output.Stdout);
         }
     }
 }
--- a/firestore/api/FirestoreTest/FirestoreTest.cs
+++ b/firestore/api/FirestoreTest/FirestoreTest.cs
@@ -230,7 +230,7 @@ namespace GoogleCloudSamples
         {
             return _manageIndexes.Run(args);
         }
-        
+
         readonly CommandLineRunner _distrubutedCounter = new CommandLineRunner()
         {
             VoidMain = DistributedCounter.Main,
@@ -840,7 +840,7 @@ namespace GoogleCloudSamples
             RunQueryData("query-create-examples", Environment.GetEnvironmentVariable("FIRESTORE_PROJECT_ID"));
             RunPaginateData("multiple-cursor-conditions", Environment.GetEnvironmentVariable("FIRESTORE_PROJECT_ID"));
         }
-        
+
         [Fact]
         public void RunDistributedCounterTest()
         {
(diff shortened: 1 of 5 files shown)


laljikanjareeya:
@googlebot I consent

firestore/api/SolutionCounter/Program.cs
```diff
@@ -0,0 +1,209 @@
...
+    }
+
+    // [END fs_counter_classes]
+
+    /// <summary>
+    /// Defines the <see cref="DistributedCounter" />
+    /// </summary>
+    public class DistributedCounter
+    { 
+        public static string Usage = @"Usage:
```
amanda-tarafa: This should probably be a const.

firestore/api/SolutionCounter/Program.cs
```diff
@@ -0,0 +1,209 @@
...
+C:\> dotnet run command YOUR_PROJECT_ID
+
+Where command is one of
+    run-distributed-counter
+";
+        /// <summary>
+        /// Gets or sets the FireStoreDb
+        /// </summary>
+        public static FirestoreDb FirestoreDb { get; set; }
```
amanda-tarafa: Why are all of these static? They probably shouldn't be.

firestore/api/SolutionCounter/Program.cs
```diff
@@ -0,0 +1,209 @@
...
+        /// </summary>
+        public static Counter ShardsCounter { get; set; }
+
+        /// <summary>
+        /// The InitFireStoreDb
+        /// </summary>
+        public static void InitFirestoreDb(string project)
+        {
+            var projectId = project;
+            if (string.IsNullOrEmpty(projectId))
```
amanda-tarafa: The check should be done before the assignment.

firestore/api/SolutionCounter/Program.cs
```diff
@@ -0,0 +1,209 @@
...
+                throw new ArgumentException("ArgumentException: failed read Firestore Project ID");
```
amanda-tarafa: No need to add the text "ArgumentException", that's already clear thanks to the exception type.
Also please include the parameter name.
The message "failed to read..." is misleading, nothing has been read, the method just received a null or empty string as a parameter. You should use the ArgumentNullException which is more suitable and then you can skip the message entirely.

firestore/api/SolutionCounter/Program.cs
```diff
@@ -0,0 +1,209 @@
...
+        /// Gets or sets the ShardsCounter
+        /// </summary>
+        public static Counter ShardsCounter { get; set; }
+
+        /// <summary>
+        /// The InitFireStoreDb
+        /// </summary>
+        public static void InitFirestoreDb(string project)
+        {
+            var projectId = project;
```
amanda-tarafa: I think you don't need this. You can work with project directly although I would rename the parameter to projectId, it is clearer.

firestore/api/SolutionCounter/Program.cs
```diff
@@ -0,0 +1,209 @@
...
+        // [START fs_increment_counter]
+
+        /// <summary>
+        /// The IncrementCounterAsync increments a randomly picked shard.
+        /// </summary>
+        /// <param name="docRef">The docRef<see cref="DocumentReference"/></param>
+        /// <returns>The <see cref="Task"/></returns>
+        public static async Task IncrementCounterAsync(DocumentReference docRef)
+        {
+            var rand = new Random();
```
amanda-tarafa: Even for a sample this should probabl be static and used within a lock.
See https://stackoverflow.com/questions/36376888.

firestore/api/SolutionCounter/Program.cs
```diff
@@ -0,0 +1,209 @@
...
+            var docId = rand.Next(0, ShardsCounter.NumShards);
+            var shardRef = docRef.Collection("shards").Document(docId.ToString());
+            await shardRef.UpdateAsync("Count", FieldValue.Increment(1));
```
amanda-tarafa: I'm not that familiar with the Firestore API, but how does UpdateAsync manages concurrent modifications? If it doesn't fail on concurrent modifcations, then in my opinion this whole example is not a good example.

firestore/api/SolutionCounter/Program.cs
```diff
@@ -0,0 +1,209 @@
...
+        // [START fs_get_count]
+
+        /// <summary>
+        /// The GetCount returns a total count across all shards.
+        /// </summary>
+        /// <param name="docRef">The docRef<see cref="DocumentReference"/></param>
+        /// <returns>The <see cref="int"/></returns>
+        public static int GetCount(DocumentReference docRef)
+        {
+            var snapshotList = docRef.Collection("shards").GetSnapshotAsync().Result;
```
amanda-tarafa: Don't use Result, use await async.

firestore/api/SolutionCounter/Program.cs
```diff
@@ -0,0 +1,209 @@
...
+            return snapshotList.Sum(snap => snap.GetValue<int>("Count"));
```
amanda-tarafa: The name snap conveys nothing. Maybe use shard.

firestore/api/SolutionCounter/Program.cs
```diff
@@ -0,0 +1,209 @@
...
+    // to realize counter with high frequency.
+
+    // [START counter_classes]
+
+    // counters/${ID}
+    /// <summary>
+    /// Defines the <see cref="Counter" />
+    /// </summary>
+    [FirestoreData]
+    public class Counter
```
amanda-tarafa: I believe this is never store in Firestore, in that case it shouldn't be FirestoreData.
But actually, for the example to make a little more sense, maybe the Counter should be stored in Firestore?

laljikanjareeya:
@amanda-tarafa Simplified and refactor the sample, Please take a look.

firestore/api/SolutionCounter/Program.cs
```diff
@@ -0,0 +1,137 @@
...
+        /// <returns>The <see cref="Task"/></returns>
+        private static async Task IncrementCounterAsync(DocumentReference docRef, int numOfShards)
+        {
+            int documentId;
+            lock (s_randLock)
+            {
+                documentId = s_rand.Next(numOfShards);
+            }
+            var shardRef = docRef.Collection("shards").Document(documentId.ToString());
+            await shardRef.UpdateAsync("count", FieldValue.Increment(1));
```
amanda-tarafa: This is just because I'm not that familiar with Firestore, but I'm a little worried still about how this handles concurrent modifications and whether the example makes sense or not then.
Imagine three different threads grab shard number 5 at the same time, that is when the counter is say, 3. They all increment the value by one, but I wonder, aftert the 3 increments, is the shard value 6 or 4? This depends on how UpdateAsync works but I'm not really that familiar with that. Maybe a test with only one shard and hundreds of simultaneous increments can shed some light on this.
My point is that, if in my example, the final result of the shard is 4, then the whole distributed counter example doesn't make a lot of sense as it stands.

firestore/api/SolutionCounter/Program.cs
```diff
@@ -0,0 +1,137 @@
...
+            {
+                case "distributed-counter":
+                    FirestoreDb db = FirestoreDb.Create(projectId);
+                    const int numberOfShards = 5;
+                    var docRef = db.Collection("counter_samples").Document("DCounter");
+
+                    CreateCounterAsync(docRef, numberOfShards).Wait();
+                    Console.WriteLine("Distributed counter created.");
+
+                    IncrementCounterAsync(docRef, numberOfShards).Wait();
```
amanda-tarafa: nit: In this wait and the above, consider using Task.Run().WaitWithUnwrappedExceptions similar to what you are doing a couple of lines down.

firestore/api/FirestoreTest/FirestoreTest.cs
```diff
@@ -829,5 +833,14 @@ public void MultipleCursorConditionsTest()
...
             RunPaginateData("multiple-cursor-conditions", Environment.GetEnvironmentVariable("FIRESTORE_PROJECT_ID"));
         }
+
+        [Fact]
+        public void RunDistributedCounterTest()
+        {
+            var output = RunDistributedCounter("distributed-counter", Environment.GetEnvironmentVariable("FIRESTORE_PROJECT_ID"));
+            Assert.Contains("Distributed counter created.", output.Stdout);
+            Assert.Contains("Distributed counter incremented.", output.Stdout);
+            Assert.DoesNotContain("Total count: 0", output.Stdout);
```
amanda-tarafa: Should be 1, right?
Maybe better `Assert.Contains("Total count: 1", output.Stdout);`

firestore/api/SolutionCounter/Program.cs
laljikanjareeya: I run with hundreds of simultaneous increments and it's working fine.
I don't see any problem here because we increment counter by 1 with existing value.

amanda-tarafa:
Just FYI: I won't get to this until next week.

firestore/api/SolutionCounter/Program.cs
amanda-tarafa: Perfect then.

PR #916, opened 2019-12-05
Add samples for asymmetric crypto operations

bdhess:


kms/api/CloudKmsSample/CloudKmsSample.csproj
```diff
@@ -7,6 +7,7 @@
   <ItemGroup>
     <PackageReference Include="CommandLineParser" Version="2.3.0" />
     <PackageReference Include="Google.Cloud.Kms.V1" Version="1.0.0" />
+    <PackageReference Include="Portable.BouncyCastle" Version="1.8.5.2" />
```
bdhess: I'm not crazy about the BC dependency, but System.Security.Cryptography doesn't get the crypto we need in .NET Core until 3.0 :(

bdhess:
@sethvargo fyi

bdhess:
I think I've gotten this down to the last of the integration test failures, and it's something I can't fix: the user running the tests needs the "Cloud KMS CryptoKey Signer/Verifier" IAM role added.

SurferJeffAtGoogle:
Please add the needed role to the list of roles here:
https://github.com/GoogleCloudPlatform/dotnet-docs-samples/blob/2f03a8105e5f3c85daf8c52e4bf42047650fabee/.kokoro-windows/Initialize-GoogleProjectForTesting.ps1#L74

And then Amanda can re-run that script which will grant the test account the necessary permissions.

kms/api/CloudKmsTest/CloudKmsTest.cs
```diff
@@ -24,7 +29,17 @@ namespace GoogleCloudSamples
     // </summary>
     public class CommonTests
     {
-        private static readonly string s_projectId = Environment.GetEnvironmentVariable("GOOGLE_PROJECT_ID");
+        private static readonly string s_projectId = LoadProjectId();
```
amanda-tarafa: Aint there a fixture on these project that already loads the project ID? If there is, you should use it. If not, this is OK.

kms/api/CloudKmsTest/CloudKmsTest.cs
```diff
@@ -24,7 +29,17 @@ namespace GoogleCloudSamples
     // </summary>
     public class CommonTests
     {
-        private static readonly string s_projectId = Environment.GetEnvironmentVariable("GOOGLE_PROJECT_ID");
+        private static readonly string s_projectId = LoadProjectId();
+
+        private static string LoadProjectId()
+        {
+            string projectId = Environment.GetEnvironmentVariable("GOOGLE_PROJECT_ID");
```
amanda-tarafa: Create a constant for the name of the env variable, since you are using it more than once.

kms/api/CloudKmsTest/CloudKmsTest.cs
```diff
@@ -279,6 +294,124 @@ private bool AssertRoleAndMemberPresent(string response, string role, string mem
...
+                Purpose = purpose,
+                VersionTemplate = new CryptoKeyVersionTemplate
+                {
+                    Algorithm = algorithm,
+                    ProtectionLevel = ProtectionLevel.Software,
+                },
+            };
+            CryptoKey key = client.CreateCryptoKey(keyRingName, cryptoKey, cryptoKeyToCreate);
+
+            CryptoKeyVersionName ckvName = CryptoKeyVersionName.Parse(key.Name + "/cryptoKeyVersions/1");
```
amanda-tarafa: Instead of doing this, use new CryptoKeyVersionname(key....., 1);

kms/api/CloudKmsTest/CloudKmsTest.cs
```diff
@@ -279,6 +294,124 @@ private bool AssertRoleAndMemberPresent(string response, string role, string mem
...
+            CryptoKeyVersion ckv = client.GetCryptoKeyVersion(ckvName);
+
+            // The version is generated asynchronously. Wait for it to be in state Enabled before
```
amanda-tarafa: When is the version created, when you create the key or when you first ask for it?
I wonder if any of these operations should be long running operations.

kms/api/CloudKmsTest/CloudKmsTest.cs
```diff
@@ -279,6 +294,124 @@ private bool AssertRoleAndMemberPresent(string response, string role, string mem
...
+            string signatureFile = @"../../../../test/data/test_file_signature.bin";
+            string timeStamp = $"-{DateTime.Now.ToString("yyyyMMddHHmmssfff")}";
+            string keyRing = $"testKeyRing{timeStamp}";
+            string cryptoKey = $"testCryptoKey{timeStamp}";
+            KeyRingName keyRingName = new KeyRingName(s_projectId, "global", keyRing);
+
+            Run("createKeyRing", keyRingName.ProjectId, keyRingName.LocationId, keyRingName.KeyRingId);
+
+            CryptoKeyVersionName version = BuildAsymmetricKey(
+                keyRingName, cryptoKey, CryptoKey.Types.CryptoKeyPurpose.AsymmetricSign,
```
amanda-tarafa: nit: Same when passing parameters.

kms/api/CloudKmsTest/CloudKmsTest.cs
```diff
@@ -279,6 +294,124 @@ private bool AssertRoleAndMemberPresent(string response, string role, string mem
...
+                    version.CryptoKeyId, version.CryptoKeyVersionId, contentFile, signatureFile);
+            Assert.Equal(0, signOutput.ExitCode);
+
+            var verifyOutput = Run("asymmetricVerify", version.ProjectId, version.LocationId, version.KeyRingId,
+                    version.CryptoKeyId, version.CryptoKeyVersionId, contentFile, signatureFile);
+            Assert.Equal(0, verifyOutput.ExitCode);
+            Assert.Contains("verified: True", verifyOutput.Stdout);
+
+            // Delete signature file to clean up
+            File.Delete(signatureFile);
```
amanda-tarafa: Shouldn't you delete the key you created server side as well?

kms/api/CloudKmsTest/CloudKmsTest.cs
```diff
@@ -279,6 +294,124 @@ private bool AssertRoleAndMemberPresent(string response, string role, string mem
...
+
+            CryptoKeyVersionName version = BuildAsymmetricKey(
+                keyRingName, cryptoKey, CryptoKey.Types.CryptoKeyPurpose.AsymmetricSign,
+                CryptoKeyVersion.Types.CryptoKeyVersionAlgorithm.EcSignP256Sha256);
+
+            var signOutput = Run("asymmetricSign", version.ProjectId, version.LocationId, version.KeyRingId,
+                    version.CryptoKeyId, version.CryptoKeyVersionId, contentFile, signatureFile);
+            Assert.Equal(0, signOutput.ExitCode);
+
+            var verifyOutput = Run("asymmetricVerify", version.ProjectId, version.LocationId, version.KeyRingId,
```
amanda-tarafa: I would rather have a test for each of these, I mean one for asymmetricSign and another one for asymmetricVerify.

kms/api/CloudKmsTest/CloudKmsTest.cs
```diff
@@ -279,6 +294,124 @@ private bool AssertRoleAndMemberPresent(string response, string role, string mem
...
+            CryptoKeyVersionName version = BuildAsymmetricKey(
+                keyRingName, cryptoKey, CryptoKey.Types.CryptoKeyPurpose.AsymmetricDecrypt,
+                CryptoKeyVersion.Types.CryptoKeyVersionAlgorithm.RsaDecryptOaep2048Sha256);
+
+            var signOutput = Run("asymmetricEncrypt", version.ProjectId, version.LocationId, version.KeyRingId,
+                    version.CryptoKeyId, version.CryptoKeyVersionId, plaintextFile, ciphertextFile);
+            Assert.Equal(0, signOutput.ExitCode);
+            Assert.NotEqual(File.ReadAllBytes(plaintextFile), File.ReadAllBytes(ciphertextFile));
+
+            var verifyOutput = Run("asymmetricDecrypt", version.ProjectId, version.LocationId, version.KeyRingId,
```
amanda-tarafa: Same with having a test for each.

kms/api/CloudKmsTest/CloudKmsTest.cs
```diff
@@ -279,6 +294,124 @@ private bool AssertRoleAndMemberPresent(string response, string role, string mem
...
+                    version.CryptoKeyId, version.CryptoKeyVersionId, ciphertextFile, recoveredPlaintextFile);
+            Assert.Equal(0, verifyOutput.ExitCode);
+            Assert.Equal(File.ReadAllBytes(plaintextFile), File.ReadAllBytes(recoveredPlaintextFile));
+
+            // Delete files to clean up
+            File.Delete(ciphertextFile);
```
amanda-tarafa: And also, delete what you created server side.

kms/api/CloudKmsTest/CloudKmsTest.cs
bdhess: There isn't a fixture in this file, and there's no dependency that pulls in a fixture either.

kms/api/CloudKmsTest/CloudKmsTest.cs
bdhess: Done

kms/api/CloudKmsTest/CloudKmsTest.cs
bdhess: Done

kms/api/CloudKmsTest/CloudKmsTest.cs
bdhess: They aren't modeled as LROs in the API, so this would require handwritten code

kms/api/CloudKmsTest/CloudKmsTest.cs
bdhess: The output from asymmetric sign is what's verified when running asymmetricVerify.
We can have a separate test, but it's just going to duplicate code.

kms/api/CloudKmsTest/CloudKmsTest.cs
bdhess: As above, the output from encrypt is what's decrypted.
Again, trying to split these into separate tests will just end up duplicating a bunch of code.

kms/api/CloudKmsTest/CloudKmsTest.cs
bdhess: As above, this is consistent with the code that already exists in this file.

kms/api/CloudKmsSample/Program.cs
```diff
@@ -551,6 +634,123 @@ public static void GetKeyRingIamPolicy(string projectId, string locationId, stri
         }
         // [END kms_decrypt]
 
+        // [START kms_asymmetric_sign]
+        public static void AsymmetricSign(string projectId, string locationId, string keyRingId,
```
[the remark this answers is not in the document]
bdhess: Is there a style guide you can refer me to?  What I've added is consistent with what's already in this file.

kms/api/CloudKmsTest/CloudKmsTest.cs
bdhess: What is the concern? Billing?

This is consistent with the tests that are already here.

kms/api/CloudKmsTest/CloudKmsTest.cs
bdhess: As above

kms/api/CloudKmsSample/CloudKmsSample.csproj
bdhess: As I understand it, it's not just a library change.  It means upgrading the project's `targetFramework` from `netcoreapp2.0` to something later, since ultimately the runtime needs to be able to invoke the right OpenSSL APIs.

kms/api/CloudKmsTest/CloudKmsTest.cs
amanda-tarafa: I'll ask around to see if they should be long running in the API but agree that if not, this is fine as it is.

kms/api/CloudKmsTest/CloudKmsTest.cs
amanda-tarafa: As above.

kms/api/CloudKmsTest/CloudKmsTest.cs
amanda-tarafa: The problem with these is that it's hard to know which of the two is failing if they fail. Kind of thing a fixture might help here with setting up and cleanup to avoid duplicating code (except for the actual testing of each sample). Testing only one sample per test is one of the things that will be in the guidelines, but I'm ok if you want to leave as is and change later.

kms/api/CloudKmsTest/CloudKmsTest.cs
amanda-tarafa: Quota, which I'm not sure if it applies for this API, but we have other samples failing because they hit quota all the time because there's no cleanup. Leave as is if the rest is like that but this will change as well in the near future.

kms/api/CloudKmsTest/CloudKmsTest.cs
amanda-tarafa: As above.

kms/api/CloudKmsTest/CloudKmsTest.cs
amanda-tarafa: As above.

amanda-tarafa:
There's a build error though, there seems to be a semicolon missing.

amanda-tarafa:
@sethvargo @bdhess Is this one superseeded by #1020 and can it be closed? Else would this one be adressed after #1020 I guess?

sethvargo:
@amanda-tarafa merging #1020 will auto-close this one (there's a note in the original description)

amanda-tarafa:
Oh that's perfect, didn't see the note.

PR #922, opened 2019-12-17
Fixes dependecy updates breakages.

amanda-tarafa:
We recently released almost all .NET client libraries, some with known breaking changes.
This PR fixes such breaking changes.

### DIFF
--- a/appengine/flexible/Stackdriver/Startup.cs
+++ b/appengine/flexible/Stackdriver/Startup.cs
@@ -79,7 +79,7 @@ namespace Stackdriver
         public void Configure(IApplicationBuilder app, IHostingEnvironment env, ILoggerFactory loggerFactory)
         {
             // Configure logging service.
-            loggerFactory.AddGoogle(Configuration["Stackdriver:ProjectId"]);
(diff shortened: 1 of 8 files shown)


amanda-tarafa:
@jskeet If you could take a look, mainly at the second commit, to confirm that the breaking change was expected. I had to add the using because the `BigTable.Admin.V2.LocationName` class was conflicting with one in `Gax.ResourceNames`. I'm not sure if we should have one in `BigTable.Admin.V2`, but we do.

jskeet:
@amanda-tarafa: I suspect it only worked before because it was using an old version of GAX that didn't include LocationName. [Browsing at the 1.0 release](https://github.com/googleapis/google-cloud-dotnet/tree/6b803b00d37842b76011760e9a89bc31adada5c0), LocationName was present then as well, so I don't think this is really a breaking change, just a new naming conflict.

In the next major version, `LocationName` will be dropped from all of these separate APIs in favour of the common one.

amanda-tarafa:
CI is now green. There's a lot of flakiness still but I'll handle that later. Merging

PR #932, opened 2020-01-14
Adds explicit authentication with JSON credentials to BQ snippets

telpirion:
This pull request adds the following region tag:
- bigquery_client_json_credentials

The Python canonical is here:
https://github.com/GoogleCloudPlatform/python-docs-samples/blob/1a87fb3e076d60ec6bde274c9aecf7dc5296a43b/bigquery/cloud-client/authenticate_service_account.py

### DIFF
--- a/bigquery/api/Snippets/AuthenticateServiceAccount.cs
+++ b/bigquery/api/Snippets/AuthenticateServiceAccount.cs
@@ -1,0 +1,32 @@
+﻿using Google.Apis.Auth.OAuth2;
+using Google.Cloud.BigQuery.V2;
+using System;
+
+public class BigQueryAuthenticateServiceAccount
+{
+    public void AuthenticateWithServiceAccount(string projectId,
+                                               string jsonPath)
+    {
+        // [START bigquery_client_json_credentials]
+        var credentials = GoogleCredential.FromFile(jsonPath);
+        var client = BigQueryClient.Create(projectId, credentials);
+        // [END bigquery_client_json_credentials]
+
+        // Use the client
+        string query = @"
+            SELECT name FROM `bigquery-public-data.usa_names.usa_1910_2013`
+            WHERE state = 'TX'
+            LIMIT 100";
+        BigQueryJob job = client.CreateQueryJob(
+            sql: query,
+            parameters: null,
+            options: new QueryOptions { UseQueryCache = false });
+        // Wait for the job to complete.
+        job = job.PollUntilCompleted().ThrowOnAnyError();
+        // Display the results
+        foreach (BigQueryRow row in client.GetQueryResults(job.Reference))
+        {
+            Console.WriteLine($"{row["name"]}");
+        }
+    }
+}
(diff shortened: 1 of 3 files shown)


bigquery/api/Snippets/AuthenticateServiceAccount.cs
```diff
@@ -0,0 +1,32 @@
+﻿using Google.Apis.Auth.OAuth2;
```
amanda-tarafa: You are missing the license here. Remember to change the year to 2020 if you copy it from another sample.

bigquery/api/Snippets/AuthenticateServiceAccount.cs
```diff
@@ -0,0 +1,32 @@
...
+                                               string jsonPath)
+    {
+        // [START bigquery_client_json_credentials]
+        var credentials = GoogleCredential.FromFile(jsonPath);
+        var client = BigQueryClient.Create(projectId, credentials);
+        // [END bigquery_client_json_credentials]
+
+        // Use the client
+        string query = @"
+            SELECT name FROM `bigquery-public-data.usa_names.usa_1910_2013`
```
amanda-tarafa: nit: move the FROM to its own line.

bigquery/api/Snippets/Test.cs
```diff
@@ -37,13 +38,23 @@ public class BigQueryTest : IDisposable, IClassFixture<RandomBucketFixture>
...
         _stringOut = new StringWriter();
         Console.SetOut(_stringOut);
     }
 
+    [Fact]
+    public void TestAuthenticateServiceAccount()
+    {
+        var snippet = new BigQueryAuthenticateServiceAccount();
+        snippet.AuthenticateWithServiceAccount(_projectId, _jsonPath);
+        var outputLines = _stringOut.ToString().Trim().Split(new[] { '\n' });
```
amanda-tarafa: The params keyword on the `Split (params char[] separator)` declaration means that you don't need to create the char array, you can simply do `.Split('\n')` or even (if you wanted to split on several characters) `.Split(',', ';', ':')`. 
Also, isntead of using '\n' consider using [`Environment.NewLine`](https://docs.microsoft.com/en-us/dotnet/api/system.environment.newline?view=netframework-4.8) so as to have into account the possibility of different new line characters. So it should look like this:
```suggestion
        var outputLines = _stringOut.ToString().Trim().Split(Environment.NewLine);
```

bigquery/api/Snippets/AuthenticateServiceAccount.cs
telpirion: Done

bigquery/api/Snippets/AuthenticateServiceAccount.cs
telpirion: Done

bigquery/api/Snippets/Test.cs
telpirion: Done

PR #937, opened 2020-01-21
Specifies UTF-8 encoding for output of Translation response 

telpirion:
This updates two code samples:

- [START translate_translate_text]
- [START translate_quickstart]

This change allows the console to output non Latin characters (e.g. Cyrillic, Chinese, Greek).


telpirion:
Thank you for the review!

PR #939, opened 2020-01-22
Changes references to Features type to be more qualified

telpirion:
These changes removes a static using statement in the Natural Language samples and replaces instances of the Features class with the more qualified AnnotateTextRequest.Types.Features.

This affects the following regions tags / code snippets:

- language_syntax_gcs
- language_syntax_text


telpirion:
Thank you for the review!

PR #942, opened 2020-01-28
Add Secret Manager IAM samples

sethvargo:


sethvargo:
Failed linux tests are for Dialogflow

sethvargo:
Hey @amanda-tarafa do you have any feedback on this?

secretmanager/api/SecretManagerSample/Program.cs
```diff
@@ -362,6 +380,88 @@ public static void GetSecret(string projectId, string secretId)
...
+        /// <code>IAMGrantAccess("my-project", "my-secret", "user:foo@example.com")</code>
+        /// </example>
+        public static void IAMGrantAccess(string projectId, string secretId, string member)
+        {
+            SecretManagerServiceClient client = SecretManagerServiceClient.Create();
+
+            // Create the request to get the current IAM policy.
+            var getRequest = new GetIamPolicyRequest
+            {
+                ResourceAsResourceName = new SecretName(projectId, secretId),
```
amanda-tarafa: This should probably be called something like SecretAsSecretName, so there's probably something missing in the proto annotations. Nothing to do with you though. Let me check internally.

secretmanager/api/SecretManagerSample/Program.cs
```diff
@@ -362,6 +380,88 @@ public static void GetSecret(string projectId, string secretId)
...
+            // Get the current IAM policy.
+            var policy = client.GetIamPolicy(getRequest);
+
+            // Add the user to the list of bindings.
+            policy.AddRoleMember("roles/secretmanager.secretAccessor", member);
+
+            // Create the request to update the IAM policy.
+            var setRequest = new SetIamPolicyRequest
+            {
+                ResourceAsResourceName = new SecretName(projectId, secretId),
```
amanda-tarafa: And same here. And the others.

secretmanager/api/SecretManagerTest/SamplesTest.cs
```diff
@@ -30,6 +30,7 @@ namespace GoogleCloudSamples
     public class SampleTests : IClassFixture<SecretsFixture>
     {
         private static readonly string s_projectId = Environment.GetEnvironmentVariable("GOOGLE_PROJECT_ID");
+        private static readonly string s_iamUser = $"user:sethvargo@google.com";
```
amanda-tarafa: Hmmm, don't think this should be public like this. Use an env variable that we can set on CI

secretmanager/api/SecretManagerSample/Program.cs
amanda-tarafa: Nope, scratch that, this is alright since the GetIamPolicyRequest message is from the IAM API and it's used to get policies for all kinds of resources.

secretmanager/api/SecretManagerSample/Program.cs
amanda-tarafa: See my comment above.

secretmanager/api/SecretManagerTest/SamplesTest.cs
sethvargo: Can you please create a new service account in the project for me? Then we can update this to use that service account and pull its value from an environment variable.

sethvargo: Please set the value of `DOTNET_SECRETMANAGER_SAMPLES_IAM_USER` to `serviceAccount:<email>`.

secretmanager/api/SecretManagerTest/SamplesTest.cs
```diff
@@ -30,6 +30,7 @@ namespace GoogleCloudSamples
     public class SampleTests : IClassFixture<SecretsFixture>
     {
         private static readonly string s_projectId = Environment.GetEnvironmentVariable("GOOGLE_PROJECT_ID");
+        private static readonly string s_iamUser = Environment.GetEnvironmentVariable("DOTNET_SECRETMANAGER_SAMPLES_IAM_USER");
```
amanda-tarafa: Actually, so we don't have to set the environment variable, consider something like this:

```
var credential = ((ServiceAccountCredential)GoogleCredential.GetApplicationDefault().UnderlyingCredential);
string email = credential.Id;
```
This will give you the email address of the service account that executes the tests. Would that work for you?

sethvargo: That will break for people who run tests locally though:

```text
Unable to cast object of type 'Google.Apis.Auth.OAuth2.UserCredential' to type 'Google.Apis.Auth.OAuth2.ServiceAccountCredential'.
```

IAM also requires that the entity is prefixed with the type (e.g. `serviceAccount:foo@bar.com`, `user:foo@bar.com`, `group:foo@bar.com`). I suppose we could add a big switch statement, but that feels like a lot of investment vs just making a static service account for these tets.

amanda-tarafa: Hmmm, true (I run locally with a service account but still might be someone who doesn't).
I'll set up the env variable then.

sethvargo: Done

PR #951, opened 2020-02-08
add synthesize-file-line

newhillchan:
synthesize file line by line with multi threads (for simplicity, text file only)

synthesize with guid instead of file content

report performance of synthesizing


newhillchan:
@googlebot I signed it!

PR #953, opened 2020-02-10
Cloud Talent solution v4 PART 1

munkhuushmgl:


amanda-tarafa:
@munkhuushmgl I probably won't get to this until next week. Let's chat later if you need me to take a look before that.

munkhuushmgl:
> @munkhuushmgl I probably won't get to this until next week. Let's chat later if you need me to take a look before that.

Thats okay, this is not urgent.

jobs/api/v4/JobsSample/CreateCompanySample.cs
```diff
@@ -0,0 +1,49 @@
...
+            string parent = TenantName.Format(projectId, tenantId);
+            Company company = new Company
+            {
+                DisplayName = displayName,
+                ExternalId = externalId
+            };
+
+            CreateCompanyRequest request = new CreateCompanyRequest
+            {
+                Parent = parent,
```
amanda-tarafa: nit: Is there something like `ParentAsParentName`? Then you can use:
(on line 26) `ParentName parent = TenantName.From(projectId, tenantId);`

(on line 35) `ParentAsParentName = parent;`

munkhuushmgl: I have changed all string values into Specific Objects [TenantName, ProjectName, ParentAsTenantOrProjectNameOneof ]

Issue #930, opened 2020-01-13
storage: separate each sample into a separate file


crwilcox:
Each sample should be in a standalone file. This workitem is to track that work.

PR #954, opened 2020-02-12
refactor(storage): separate sample for create bucket

laljikanjareeya:
towards #930

### DIFF
--- a/storage/api/Storage.sln
+++ b/storage/api/Storage.sln
@@ -8,6 +8,8 @@ Project("{9A19103F-16F7-4668-BE54-9A1E7A4F7556}") = "Storage", "Storage\Storage.
 EndProject
 Project("{9A19103F-16F7-4668-BE54-9A1E7A4F7556}") = "StorageTest", "StorageTest\StorageTest.csproj", "{22A2AE8E-6635-496E-A52A-EAA63361EEAA}"
 EndProject
+Project("{9A19103F-16F7-4668-BE54-9A1E7A4F7556}") = "testutil", "..\..\testutil\testutil.csproj", "{C0736C40-22CC-467A-8020-A965AD16AF9B}"
+EndProject
 Global
 	GlobalSection(SolutionConfigurationPlatforms) = preSolution
 		Debug|Any CPU = Debug|Any CPU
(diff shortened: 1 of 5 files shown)


storage/api/Storage.sln
```diff
@@ -54,6 +56,18 @@ Global
...
+		{C0736C40-22CC-467A-8020-A965AD16AF9B}.Debug|x64.ActiveCfg = Debug|Any CPU
+		{C0736C40-22CC-467A-8020-A965AD16AF9B}.Debug|x64.Build.0 = Debug|Any CPU
+		{C0736C40-22CC-467A-8020-A965AD16AF9B}.Debug|x86.ActiveCfg = Debug|Any CPU
+		{C0736C40-22CC-467A-8020-A965AD16AF9B}.Debug|x86.Build.0 = Debug|Any CPU
+		{C0736C40-22CC-467A-8020-A965AD16AF9B}.Release|Any CPU.ActiveCfg = Release|Any CPU
+		{C0736C40-22CC-467A-8020-A965AD16AF9B}.Release|Any CPU.Build.0 = Release|Any CPU
+		{C0736C40-22CC-467A-8020-A965AD16AF9B}.Release|x64.ActiveCfg = Release|Any CPU
+		{C0736C40-22CC-467A-8020-A965AD16AF9B}.Release|x64.Build.0 = Release|Any CPU
+		{C0736C40-22CC-467A-8020-A965AD16AF9B}.Release|x86.ActiveCfg = Release|Any CPU
+		{C0736C40-22CC-467A-8020-A965AD16AF9B}.Release|x86.Build.0 = Release|Any CPU
```
laljikanjareeya: This is about to add `testutil` project in current solution as Visual Studio was complaining about `testutil` project.

laljikanjareeya:
@amanda-tarafa please provide feedback if any.
For other samples I will create one PR based on your feedback.

amanda-tarafa:
FYI: I won't get to this until next week.

storage/api/Storage/CreateNewBucket.cs
```diff
@@ -0,0 +1,31 @@
...
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+using Google.Cloud.Storage.V1;
+using System;
+
+namespace Storage
+{
+    public class CreateNewBucket
```
amanda-tarafa: I would name it CreateBucket as well as the file.

storage/api/Storage/CreateNewBucket.cs
```diff
@@ -0,0 +1,31 @@
...
+    {
+        // [START storage_create_bucket]
+        public static void CreateBucket(string projectId, string bucketName)
+        {
+            var storage = StorageClient.Create();
+            storage.CreateBucket(projectId, bucketName);
```
amanda-tarafa: I don't remember if CreateBucket returns something? If it does I believe I would make the samples more interesting if the returned value is used. I understand this might fall out of the scope of this PR though.

laljikanjareeya: Added Return Type.

storage/api/Storage/CreateNewBucket.cs
laljikanjareeya: I was also confused as it will give error while giving both file and method name same.
just rename the file name as per your suggestion and method name accordingly.

PR #959, opened 2020-02-20
[Storage] feat(#955): add support for Iam conditions

frankyn:
Fix: #955

### DIFF
--- a/storage/api/Storage/Storage.cs
+++ b/storage/api/Storage/Storage.cs
@@ -716,17 +716,17 @@ namespace GoogleCloudSamples
                 RequestedPolicyVersion = 3
             });
             policy.Version = 3;
-            policy.Bindings.ToList().ForEach(response =>
+            policy.Bindings.ToList().ForEach(binding =>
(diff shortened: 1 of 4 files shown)


storage/api/Storage/Storage.cs
```diff
@@ -647,27 +659,66 @@ private void ViewBucketIamMembers(string bucketName)
...
             string role, string member)
         {
             var storage = StorageClient.Create();
-            var policy = storage.GetBucketIamPolicy(bucketName);
+            var policy = storage.GetBucketIamPolicy(bucketName, new GetBucketIamPolicyOptions()
+            {
+                RequestedPolicyVersion = 3
+            });
+            policy.Version = 3;
             policy.Bindings.ToList().ForEach(response =>
```
jkwlui: let's rename `response` to `binding` so it's clear these are bindings.

frankyn:
Thanks @jkwlui! Addressed feedback. Pending review from @jsimonweb.

PR #961, opened 2020-02-21
fix: removed not about supported enhanced models for Speech-to-Text

telpirion:
Speech-to-Text now supports more enhanced models than just phone_call. The comment in the samle needs to be updated to reflect this.

This pull request affects the following region tags:

- speech_transcribe_enhanced_model


telpirion:
Thank you!

PR #967, opened 2020-02-27
[Secret Manager] Updated version v1beta1->v1

gwhitehawk:
Client library has been published: https://www.nuget.org/packages/Google.Cloud.SecretManager.V1

### DIFF
--- a/secretmanager/api/QuickStart/QuickStart.csproj
+++ b/secretmanager/api/QuickStart/QuickStart.csproj
@@ -5,6 +5,6 @@
     <GenerateAssemblyInfo>false</GenerateAssemblyInfo>
   </PropertyGroup>
   <ItemGroup>
-    <PackageReference Include="Google.Cloud.SecretManager.V1Beta1" Version="1.0.0-beta01" />
+    <PackageReference Include="Google.Cloud.SecretManager.V1" Version="1.0.0-beta01" />
   </ItemGroup>
 </Project>
--- a/secretmanager/api/SecretManagerSample/SecretManagerSample.csproj
+++ b/secretmanager/api/SecretManagerSample/SecretManagerSample.csproj
@@ -6,7 +6,7 @@
   </PropertyGroup>
   <ItemGroup>
     <PackageReference Include="CommandLineParser" Version="2.3.0" />
-    <PackageReference Include="Google.Cloud.SecretManager.V1Beta1" Version="1.0.0-beta01" />
+    <PackageReference Include="Google.Cloud.SecretManager.V1" Version="1.0.0-beta01" />
     <PackageReference Include="xunit" Version="2.4.1" />
   </ItemGroup>
   <ItemGroup>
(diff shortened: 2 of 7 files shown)


secretmanager/api/QuickStart/QuickStart.csproj
```diff
@@ -5,6 +5,6 @@
     <GenerateAssemblyInfo>false</GenerateAssemblyInfo>
   </PropertyGroup>
   <ItemGroup>
-    <PackageReference Include="Google.Cloud.SecretManager.V1Beta1" Version="1.0.0-beta01" />
+    <PackageReference Include="Google.Cloud.SecretManager.V1" Version="1.0.0-beta01" />
```
sethvargo: I think the `version` may also need to change? I'm not sure what the new package version is though.

secretmanager/api/SecretManagerSample/SecretManagerSample.csproj
```diff
@@ -6,7 +6,7 @@
   </PropertyGroup>
   <ItemGroup>
     <PackageReference Include="CommandLineParser" Version="2.3.0" />
-    <PackageReference Include="Google.Cloud.SecretManager.V1Beta1" Version="1.0.0-beta01" />
+    <PackageReference Include="Google.Cloud.SecretManager.V1" Version="1.0.0-beta01" />
```
sethvargo: Same note on version

gwhitehawk: I was actually confused by this beta01 as well, hence I verified in Nuget. The version (for V1) really is 1.0.0-beta01: https://www.nuget.org/packages/Google.Cloud.SecretManager.V1. This seems deliberate and is consistent with other products where they moved to 1.0.0 only after a couple of months.

secretmanager/api/QuickStart/QuickStart.csproj
gwhitehawk: See below.

secretmanager/api/SecretManagerSample/SecretManagerSample.csproj
sethvargo: Wow. That's definitely surprising, but thanks for checking!

amanda-tarafa: So, in .NET at least, at the library level we deal with the API version and with the client library version, and we follow semver for client library versioning. For a beta API the library will never be GA. For a GA API the library can be beta or GA (we rarely publish alphas although it may happen).
When an API goes GA, even if the client library is fully autogenerated, as with this case, we still release only a beta and wait for 2 or 3 weeks before doing a GA release. It's not unusual to discover breaking issues with proto annotations in that period, for instance, and having released only a beta library there's no problem with taking a breaking change that comes from fixing the protos and regenaring the library. If that happens we reset the beta waiting period.
Even if a GA version of a client library has been released for a given API and even if the API introduces no new versions, we sometimes need to bump up library versions because of breaking changes introduced by other dependencies, for instance gRPC (as we had to do just a couple of weeks ago for all libraries).
If no problems are found with Secret Manager V1 then we'll be releasing a GA client library in a few weeks. You won't have to do anything because there's a bot in this repo that updates the current libraries to it's latest version.
Happy to go into all of this further.

PR #971, opened 2020-03-06
Fix invalid permission name

bdhess:
This is what I get for going from memory rather than consulting the
docs:
https://cloud.google.com/kms/docs/reference/permissions-and-roles#predefined_roles


bdhess:
@amanda-tarafa I don't have write access; can you please merge?

amanda-tarafa:
Done!

PR #972, opened 2020-03-10
Dependency cleanup on several projects.

amanda-tarafa:
Some of the tests projects had dependencies on explicit versions of the Apiary libraries instead of on the projects containing the samples for the Apiary libraries. With the latest Apiary releases, version conflicts appeared.
Similar issues with external libraries like Newtonsoft.Json.
And, cleared some unused dependencies as well.

A similar thing will probably happen when we release gRPC libraries in the comming weeks.

### DIFF
--- a/dlp/api/DlpSample/RedactSamples.cs
+++ b/dlp/api/DlpSample/RedactSamples.cs
@@ -1,4 +1,18 @@
-﻿using Google.Api.Gax.ResourceNames;
+﻿// Copyright 2018 Google Inc.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+using Google.Api.Gax.ResourceNames;
 using Google.Cloud.Dlp.V2;
 using Google.Protobuf;
 using System;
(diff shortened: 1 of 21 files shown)


dlp/api/DlpSample/RedactSamples.cs
```diff
@@ -1,4 +1,18 @@
-﻿using Google.Api.Gax.ResourceNames;
+﻿// Copyright 2018 Google Inc.
```
jskeet: Was this file actually created in 2018? (Fine if so, obviously.)

iot/api/CloudIotMqttExample/CloudIotMqttExample.csproj
```diff
@@ -1,19 +1,16 @@
 ﻿<Project Sdk="Microsoft.NET.Sdk">
   <PropertyGroup>
     <OutputType>Exe</OutputType>
-    <TargetFramework>netcoreapp2.0</TargetFramework>
+    <TargetFramework>net461</TargetFramework>
```
jskeet: I'd normally consider going from Core to Framework a retrograde step. Is this because one of the dependencies doesn't support Core?

jskeet: (Feel free to update the NodaTime dependency to 2.4.7 at the same time...)

dlp/api/DlpSample/RedactSamples.cs
amanda-tarafa: Yep, March 2018 :).

iot/api/CloudIotMqttExample/CloudIotMqttExample.csproj
amanda-tarafa: It's Bouncy Castle and one other. I did tried to update them but still, they don't support Core. Bouncy Castle will starting from 1.9.* bcgit/bc-csharp/pull/68.

I will update NodaTime and push again after Kokoro has finished.

PR #981, opened 2020-03-19
[Job V2] Deletes unused and obsolote Jobs V2 API Quickstart.

amanda-tarafa:


amanda-tarafa:
Will self review and merge given that this is just deleting known obsolote and unused code.

PR #984, opened 2020-03-24
video:logo detect GA[URGENT]

munkhuushmgl:
canonical sample: https://github.com/GoogleCloudPlatform/python-docs-samples/pull/3098

### DIFF
--- a/video/api/Analyze/Program.cs
+++ b/video/api/Analyze/Program.cs
@@ -63,6 +63,10 @@ namespace GoogleCloudSamples.VideoIntelligence
     class TrackObjectOptions : VideoOptions
     { }
 
+    [Verb("logo-detect", HelpText = "Detect logos in a video")]
+    class DetectLogoOptions : VideoOptions
+    { }
+
     public class Analyzer
     {
         // [START video_analyze_shots]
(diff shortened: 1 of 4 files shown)


video/api/Analyze/Program.cs
```diff
@@ -224,6 +228,166 @@ public static object TranscribeVideo(string uri)
...
+                Console.WriteLine($"Description :{entity.Description}");
+
+                // All logo tracks where the recognized logo appears. Each track corresponds to one logo
+                // instance appearing in consecutive frames.
+                foreach (var track in logoRecognitionAnnotation.Tracks)
+                {
+                    // Video segment of a track.
+                    var startTimeOffset = track.Segment.StartTimeOffset;
+                    Console.WriteLine(
+                        $"\n\tStart Time Offset: {startTimeOffset.Seconds}.{startTimeOffset.Nanos}\n");
```
amanda-tarafa: Nit: remove the line break and tabs (\n\t). Here and everywhere.

munkhuushmgl:
> LGTM, just remove the line breaks and tabs.

I think the intention of tabs was to show that output comes from nested loop.

Did you mean by deleting all tabs and breaks or only ones that are in the beginning of `Video segment of track`?

amanda-tarafa:
> > LGTM, just remove the line breaks and tabs.
> 
> I think the intention of tabs was to show that output comes from nested loop.
> 
> Did you mean by deleting all tabs and breaks or only ones that are in the beginning of `Video segment of track`?

Your are right abouth tabs, but still is slightly weird. If you have the time substitute the  tabs for - and spaces where necessary. But only if you have the time.
Line breaks you don't need, because Console.WriteLine will write a new line each time. If you leave the line breaks, then you'll get an empty line in between each line.

Issue #988, opened 2020-03-25
Check that all sample library dependencies have been updated and fix breakages.


amanda-tarafa:
After merging #952 which updated library dependencies for most sample projects, CI has been knowingly broken. The libraries have taken a major version bump, breaking changes included, and that is what has broken samples.
I'm working through this right now.

amanda-tarafa:
Note to self: If by the time this is fixed, Dialogflow hasn't generated properly (googleapis/google-cloud-dotnet#4646) create a new issue to track that and update Dialoglfow dependencies when it has generated properly.

amanda-tarafa:
Note to self: IRM samples. (have checked internally)

amanda-tarafa:
> Note to self: IRM samples. (have checked internally)

#1000 Created.

amanda-tarafa:
> Note to self: If by the time this is fixed, Dialogflow hasn't generated properly ([googleapis/google-cloud-dotnet#4646](https://github.com/googleapis/google-cloud-dotnet/issues/4646)) create a new issue to track that and update Dialoglfow dependencies when it has generated properly.

Created #1003

PR #991, opened 2020-03-26
spanner: Backup and Restore samples

### LINKED ISSUE #985: Spanner: Samples for Backup and Restore functionaity
- CreateBackup - create a new backup 
- DeleteBackup - Deletes a pending or completed backup.
- GetBackup - Gets metadata on a pending or completed backup.
- GetBackupOperations - get backup operations in the instance.
- GetBackups - Get completed and pending backups.
- RestoreDatabase - Creates a new database by restoring from a completed backup.
- UpdateBackup - Updates the expiration time of a pending or completed backup.
- CancelBackupOperation- cancel backup create operation.
- GetDatabaseOperations Get Database operations

laljikanjareeya:
Fixes #985


spanner/api/Spanner/CreateBackup.cs
```diff
@@ -0,0 +1,72 @@
...
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+using Google.Cloud.Spanner.Admin.Database.V1;
+using Google.Cloud.Spanner.Common.V1;
+using Google.LongRunning;
+using Google.Protobuf.WellKnownTypes;
+using log4net;
+using System;
+using static GoogleCloudSamples.Spanner.Program;
```
skuruppu, line 21: Would it be possible to include the imports within the tags so that it's a complete example? How is this done for separated samples in other languages?

skuruppu:
@amanda-tarafa, would it be ok to ask you for feedback on the spanner/api/Spanner/CreateBackup.cs sample? Before putting more effort into tests, just wanted get your thoughts on if you're happy with this new sample setup.

amanda-tarafa:
@skuruppu Yes, I'll take a look as soon as I can. I'm still trying to get #988 done, almost there now so it shouldn't be long.

spanner/api/Spanner/CreateBackup.cs
```diff
@@ -0,0 +1,65 @@
...
+                Database = DatabaseName.Format(projectId, instanceId, databaseId),
+                ExpireTime = DateTime.UtcNow.AddDays(14).ToTimestamp()
+            };
+            string parent = InstanceName.Format(projectId, instanceId);
+
+            // Make the CreateBackup request.
+            Operation<Backup, CreateBackupMetadata> response =
+                databaseAdminClient.CreateBackup(parent, backup, backupId);
+
+            s_logger.Info("Waiting for the operation to finish.");
```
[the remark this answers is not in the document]
laljikanjareeya: For consistency, I have seen in `Program.cs` where used `s_logger.Info` to print the information.

spanner/api/Spanner/CreateBackup.cs
```diff
@@ -0,0 +1,65 @@
...
+                response.PollUntilCompleted();
+
+            if (completedResponse.IsFaulted)
+            {
+                s_logger.Error($"Error while creating backup: {completedResponse.Exception}");
+                return ExitCode.InvalidParameter;
+            }
+
+            s_logger.Info($"Backup created successfully.");
+            return ExitCode.Success;
```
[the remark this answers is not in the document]
laljikanjareeya: Ideally I should return the Information from the sample  e.g `CreateBackup` return the newly created `Backup` Object but currently in `Program.cs` it has configure as return `int` on below line so I Just return the `ExitCode`.
https://github.com/GoogleCloudPlatform/dotnet-docs-samples/blob/42b67b226d1350a475bbf4fa2080786c350b1159/spanner/api/Spanner/Program.cs#L3775

spanner/api/Spanner/CreateBackup.cs
laljikanjareeya: We need refactor all the samples with tags as it's done in [storage](https://github.com/GoogleCloudPlatform/dotnet-docs-samples/pull/973#pullrequestreview-380738531).

spanner/api/Spanner/CancelBackupOperation.cs
```diff
@@ -0,0 +1,70 @@
...
+
+            s_logger.Info("Waiting for the operation to finish.");
+
+            // Poll until the long-running operation is complete. It will
+            // either complete or be cancelled.
+            Operation<Backup, CreateBackupMetadata> completedResponse =
+                response.PollUntilCompleted();
+
+            if (!completedResponse.IsFaulted)
+            {
```
amanda-tarafa, line 60: Hmm I think it would be a little more idiomatic to poll inside a try catch:
... = response.PollUntilCompleted().ThrowOnAnyError();

And actually, is CancelOperation long running? Would it failed if the backup was already created? Because if so, I think that the clearest thing would be to poll and throw CancelOperation inside the try catch instead.

laljikanjareeya: Yes `CancelOperation` is long running but it never throws any exception except `google.rpc.Code.UNIMPLEMENTED` when API is not implemented.
 
It has to check by `GetOperation` and if the value of status code is `Code.CANCELLED` then CancelOperation has been successfully cancel, but we never know when this CancelOperation will Finish and we don't have any Poll option on same.

spanner/api/Spanner/CancelBackupOperation.cs
```diff
@@ -0,0 +1,70 @@
...
+using System;
+using static GoogleCloudSamples.Spanner.Program;
+
+namespace GoogleCloudSamples.Spanner
+{
+    public class CancelBackupOperation
+    {
+        static readonly ILog s_logger = LogManager.GetLogger(typeof(CancelBackupOperation));
+
+        // [START spanner_cancel_backup_create]
```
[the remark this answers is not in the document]
laljikanjareeya: Fixed in All Files.

spanner/api/Spanner/CancelBackupOperation.cs
```diff
@@ -0,0 +1,70 @@
...
+        public static object SpannerCancelBackupOperation(
+            string projectId, string instanceId, string databaseId, string backupId)
+        {
+            // Create the DatabaseAdminClient instance.
+            DatabaseAdminClient databaseAdminClient = DatabaseAdminClient.Create();
+
+            // Initialize request parameters.
+            Backup backup = new Backup
+            {
+                Database = DatabaseName.Format(projectId, instanceId, databaseId),
```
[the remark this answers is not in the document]
laljikanjareeya: `DatabaseAsDatabaseName` is not available here.

spanner/api/Spanner/CancelBackupOperation.cs
```diff
@@ -0,0 +1,70 @@
...
+using Google.Protobuf.WellKnownTypes;
+using log4net;
+using System;
+using static GoogleCloudSamples.Spanner.Program;
+
+namespace GoogleCloudSamples.Spanner
+{
+    public class CancelBackupOperation
+    {
+        static readonly ILog s_logger = LogManager.GetLogger(typeof(CancelBackupOperation));
```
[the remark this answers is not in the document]
laljikanjareeya: Fixed in All Files.

spanner/api/Spanner/CreateBackup.cs
laljikanjareeya: Done!

spanner/api/Spanner/CreateBackup.cs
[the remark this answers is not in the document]
laljikanjareeya: Done!!

spanner/api/Spanner/CreateBackup.cs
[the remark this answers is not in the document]
laljikanjareeya: Just return `0` or `1` in all the sample removed dependency of `ExitCode`.

spanner/api/Spanner/CancelBackupOperation.cs
skuruppu: Just to clarify, I don't think the `CancelOperation` is long-running. It sends the cancel request to the server which registers the intent and returns immediately. The backend then goes about cancelling the `CreateBackup` operation in its own time. It's possible for the backup to be created before the backend tries to cancel it. Hence the `PollUntilCompleted()` above.

@laljikanjareeya is right about checking for the `Code.CANCELLED` in the original `CreateBackup` operation to check if it was actually cancelled before the backup could be completed.

spanner/api/Spanner/CancelBackupOperation.cs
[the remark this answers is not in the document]
skuruppu: Agree, we seem to be missing an annotation. I will try to fix it.

spanner/api/Spanner/CancelBackupOperation.cs
```diff
@@ -0,0 +1,70 @@
...
+            // Create the DatabaseAdminClient instance.
+            DatabaseAdminClient databaseAdminClient = DatabaseAdminClient.Create();
+
+            // Initialize request parameters.
+            Backup backup = new Backup
+            {
+                Database = DatabaseName.Format(projectId, instanceId, databaseId),
+                ExpireTime = DateTime.UtcNow.AddDays(14).ToTimestamp()
+            };
+            string parent = InstanceName.Format(projectId, instanceId);
```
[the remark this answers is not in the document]
skuruppu: Now fixed for all samples except in the `Backup` class.

spanner/api/Spanner/CancelBackupOperation.cs
amanda-tarafa: OK, I understand this is the APIs design, it's just I find it weird to have to poll the operation being cancelled and check if it failed to actually know that it was cancelled.
Ths intuitive thing for me would have been to poll the cancel operation itself and if that fails then it's because the backup was created before you could cancel it.
I suppose it is too late to try to influence that?

spanner/api/Spanner/CreateBackup.cs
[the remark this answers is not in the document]
amanda-tarafa: > Ideally It should return the Information from the sample e.g CreateBackup return the newly created Backup Object

Absolutely right, but as you say, it's not currently possible with the current sample/testing setup. As an FYI we'll be moving towards this in the near future, no more command line runners, just method that return relevant results and are tested as such.

spanner/api/Spanner/CancelBackupOperation.cs
skuruppu: Yeah, I agree that would've been more intuitive. Unfortunately it's too late to change that behavior in the backend since the APIs were finalized sometime earlier last year.

spanner/api/Spanner/CancelBackupOperation.cs
```diff
@@ -0,0 +1,68 @@
...
+            };
+            InstanceName parentAsInstanceName = InstanceName.FromProjectInstance(projectId, instanceId);
+
+            // Make the CreateBackup request.
+            Operation<Backup, CreateBackupMetadata> response =
+                databaseAdminClient.CreateBackup(parentAsInstanceName, backup, backupId);
+
+            // Create the OperationsClient instance and execute CancelOperation.
+            OperationsClient operationsClient = OperationsClient.Create();
+            operationsClient.CancelOperation(response.Name);
```
skuruppu, line 46: @laljikanjareeya I added a test for this sample but I'm having trouble getting it to work. This call returns a "CANCELLED" error and the sample exits because we don't have the call around a try/catch. But it doesn't make sense to me why the cancel operation is cancelled. If you could run the test and take a look, that would be great.

laljikanjareeya: @skuruppu I am also getting below exception:
`Status(StatusCode=Cancelled, Detail="Received http2 header with status: 404")`.
Also tried with `CancelOperationRequest` but same error, digging more into it will let you know further findings.

skuruppu: I managed to find a fix. I don't know why but the error is 404 so for some reason the backend is not able to find the operation by name. But you can also cancel an operation by just doing Operation.Cancel() so I changed it to that and everything works. It's also much simpler.

skuruppu:
@amanda-tarafa this PR is now ready to be reviewed. I will work on pushing a proto update to get the annotations in the meantime.

spanner/api/Spanner/CancelBackupOperation.cs
amanda-tarafa: Ok so the real problem here is that you shouldn't be using the OperationsClient directly. I'm not on a computer but I'll look later and try to see exactly how this should be done. For the time being see a response to a similar issue here. Sorry for not realizing earlier.
 https://github.com/googleapis/google-cloud-dotnet/issues/4843#issuecomment-611731182

amanda-tarafa: Oh, I think Shanika's comment and mine got crossed. Shanika, the solution that you found is the right one.
Using the OperationsClient directly would give a 404 because that's an OperationsClient that we use for all long running operations from all APIs, so as you were using it it doesn't know it needs to talk to Spanner.
What you did is the cleanest solution, but if you want to use an OperationsClient for dealing with Backup creation operations the you need to use the one in `databaseAdminClient.CreateBackupOperationsClient`.

skuruppu: Ah ok that makes a lot of sense now. Thanks for the explanation. I don't think we every want our users to use the OperationsClient directly. So having the samples doing the simplest thing, even for creating a backup, is the way to go.

skuruppu:
I don't understand the [Windows test failure](https://source.cloud.google.com/results/invocations/d48d4e44-668c-49e6-919f-15133df493c4/targets) because it got a permission denied error while the other tests didn't.

The error is also weird because the error is in `SpannerTests.GoogleCloudSamples.Spanner.SpannerTests.TestGetBackupOperations`.

The error itself though is for CreateBackup:
```
Error: Grpc.Core.RpcException : Status(StatusCode=PermissionDenied, Detail="Caller is missing IAM permissions.\n Resource: projects/dotnet-docs-samples-tests/instances/my-instance/databases/my-db-jldukupjnsvrbwaenusb Permission: spanner.databases.createBackup\nResource: projects/dotnet-docs-samples-tests/instances/my-instance Permission: spanner.backups.create.")
```

But the output seems to indicate that it was able to complete the CreateBackup call.
```
Spanner createSampleDatabase dotnet-docs-samples-tests my-instance my-db-etkeupujpxwnplxtjhxj
QuickStart
Spanner insertSampleData dotnet-docs-samples-tests my-instance my-db-etkeupujpxwnplxtjhxj
Spanner insertStructSampleData dotnet-docs-samples-tests my-instance my-db-etkeupujpxwnplxtjhxj
Spanner addColumn dotnet-docs-samples-tests my-instance my-db-etkeupujpxwnplxtjhxj
Passed   GoogleCloudSamples.Spanner.QuickStartTests.TestQuickStart
Spanner addCommitTimestamp dotnet-docs-samples-tests my-instance my-db-etkeupujpxwnplxtjhxj
Spanner addIndex dotnet-docs-samples-tests my-instance my-db-etkeupujpxwnplxtjhxj
Spanner createTableWithDatatypes dotnet-docs-samples-tests my-instance my-db-etkeupujpxwnplxtjhxj
Spanner writeDatatypesData dotnet-docs-samples-tests my-instance my-db-etkeupujpxwnplxtjhxj
Spanner createBackup dotnet-docs-samples-tests my-instance my-db-etkeupujpxwnplxtjhxj my-backup-vmmucajeqdpjmfbkecvs
Spanner queryWithBytes dotnet-docs-samples-tests my-instance my-db-etkeupujpxwnplxtjhxj
[xUnit.net 00:02:06.59]     GoogleCloudSamples.Spanner.SpannerTests.TestGetBackupOperations [FAIL]
BUILD FAILED
```

Looking more into the passed tests, it looks like we don't run SpannerTest in all environments, only in "Kokoro CI Windows". So I think we actually have a permissions issue. I don't have access to the cloud project. @amanda-tarafa do you think you or someone in your team will access can give the relevant permissions to the project? The permissions we need are:

* Cloud Spanner Backup Admin
* Cloud Spanner Backup Writer
* Cloud Spanner Restore Admin

amanda-tarafa:
I'll take a look.

amanda-tarafa:
Permissions added, rerunning Kokoro

spanner/api/Spanner/CreateBackup.cs
```diff
@@ -0,0 +1,71 @@
...
+        public static object SpannerCreateBackup(
+            string projectId, string instanceId, string databaseId, string backupId)
+        {
+            // Create the DatabaseAdminClient instance.
+            DatabaseAdminClient databaseAdminClient = DatabaseAdminClient.Create();
+
+            // Initialize request parameters.
+            Backup backup = new Backup
+            {
+                Database = DatabaseName.Format(projectId, instanceId, databaseId),
```
amanda-tarafa, line 36: Just flagging for when the annotations are in. DatabaseAsDatabaseName

spanner/api/Spanner/CancelBackupOperation.cs
```diff
@@ -0,0 +1,67 @@
...
+                ExpireTime = DateTime.UtcNow.AddDays(14).ToTimestamp()
+            };
+            InstanceName parentAsInstanceName = InstanceName.FromProjectInstance(projectId, instanceId);
+
+            // Make the CreateBackup request.
+            Operation<Backup, CreateBackupMetadata> operation =
+                databaseAdminClient.CreateBackup(parentAsInstanceName, backup, backupId);
+
+            // Cancel the operation.
+            operation.Cancel();
```
amanda-tarafa, line 46: What's the difference between canceling a backup and deleting one?
The docs for deleting say "Deletes a pending or completed backup", so I wonder if maybe canceling here is the same as deleting, and it would be best to just delete since we might need to do so anyway if the cancel didn't came in time?
Just asking, I'm fime with the sample as is.

spanner/api/Spanner/GetBackups.cs
```diff
@@ -0,0 +1,107 @@
...
+    {
+        public static object SpannerGetBackups(
+            string projectId, string instanceId, string databaseId, string backupId)
+        {
+            // Create the DatabaseAdminClient instance.
+            DatabaseAdminClient databaseAdminClient = DatabaseAdminClient.Create();
+
+            InstanceName parentAsInstanceName = InstanceName.FromProjectInstance(projectId, instanceId);
+
+            Action<List<Backup>> printBackups = backups =>
```
amanda-tarafa, line 36: Consider defining a local method instead, at the end of this one:

void PrintBackups(List<Backup> backups) {...}

spanner/api/Spanner/GetBackups.cs
```diff
@@ -0,0 +1,107 @@
...
+            int pageSize = 5;
+            string nextPageToken = string.Empty;
+            do
+            {
+                var response = databaseAdminClient.ListBackups(parentAsInstanceName, nextPageToken);
+
+                Page<Backup> currentPage = response.ReadPage(pageSize);
+                foreach (Backup backup in currentPage)
+                {
+                    Console.WriteLine($"Backup Name : {backup.Name}");
```
amanda-tarafa, line 97: If instead of a list you define the method as PrinBackups(IEnumerable<Backup> backups)... you can call it here as well.

spanner/api/Spanner/GetBackups.cs
```diff
@@ -0,0 +1,107 @@
...
+            Console.WriteLine("Backups with size > 100 bytes:");
+            request.Filter = "size_bytes > 100";
+            var backupsWithSize = databaseAdminClient.ListBackups(request).ToList();
+            printBackups(backupsWithSize);
+
+            // List backups created in the last day that are ready.
+            Console.WriteLine("Backups created within last day that are ready:");
+            string createTime = DateTime.UtcNow.AddDays(-1).ToString("O");
+            request.Filter = $"create_time >= \"{createTime}\" AND state:READY";
+            var recentReadyBackups = databaseAdminClient.ListBackups(request).ToList();
```
amanda-tarafa, line 83: If instead of a list you define the method as PrinBackups(IEnumerable<Backup> backups)... you can skip the ToList() call.

spanner/api/Spanner/GetDatabaseOperations.cs
```diff
@@ -0,0 +1,55 @@
...
+            var filter = "(metadata.@type:type.googleapis.com/google.spanner.admin.database.v1.OptimizeRestoredDatabaseMetadata)";
+
+            ListDatabaseOperationsRequest request = new ListDatabaseOperationsRequest
+            {
+                ParentAsInstanceName = InstanceName.FromProjectInstance(projectId, instanceId),
+                Filter = filter
+            };
+
+            // List the optimize restored databases operations on the instance.
+            var operations = databaseAdminClient.ListDatabaseOperations(request).ToList();
```
amanda-tarafa, line 40: In general skip the ToList() calls. What that does is to execute get all the pages one after the other, while if you don't, the pages will be obtained when necessary, i.e., when you get to it on an iteration.

spanner/api/Spanner/GetDatabaseOperations.cs
```diff
@@ -0,0 +1,55 @@
...
+
+            operations.ForEach(operation =>
```
amanda-tarafa, line 42: Better use foreach(var operation in operations) {...} it's way more idiomatic. Here and everywhere.

spanner/api/Spanner/RestoreDatabase.cs
```diff
@@ -0,0 +1,62 @@
...
+
+            // Make the RestoreDatabase request.
+            Operation<Database, RestoreDatabaseMetadata> response =
+                databaseAdminClient.RestoreDatabase(
+                    parentAsInstanceName, databaseId, backupAsBackupName);
+
+            Console.WriteLine("Waiting for the operation to finish");
+
+            // Poll until the returned long-running operation is complete.
+            var completedResponse = response.PollUntilCompleted();
```
amanda-tarafa, line 45: It's probably more idiomatic to call response.PollUntilCompleted().ThrowOnAnyError(); inside a try catch block and handle exceptions as usual. Here and everywhere.

spanner/api/Spanner/GetBackups.cs
laljikanjareeya: Done!

spanner/api/Spanner/GetDatabaseOperations.cs
laljikanjareeya: Removed ToList() from everywhere.

spanner/api/Spanner/RestoreDatabase.cs
laljikanjareeya: I see `ThrowOnAnyError` is available in `BigQuery` but, it's not available in Spanner.

spanner/api/Spanner/CancelBackupOperation.cs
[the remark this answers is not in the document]
skuruppu: This is available in v3.0.0 so I've updated the deps and fixed this.

spanner/api/Spanner/CancelBackupOperation.cs
skuruppu: I guess the only difference is that delete is guaranteed to delete your backup while cancel is more of a best-effort operation and only applies if the backup is in the process of being created. If it's already created then it's a no-op.

Since the backup creation is a LRO, the backend team asked us to show how to cancel that LRO with this sample so for consistency with what's in the Cloud docs, we should leave this here as is.

skuruppu:
Looks like the Kokoro builds run tests with a 10 min timeout but creating and restoring a backup can take way longer than that. @amanda-tarafa would it be ok for us to increase the timeout?

amanda-tarafa:
@skuruppu Yes please do so. You can do that in the runTests.ps1 script for Spanner tests.

skuruppu:
@amanda-tarafa, this PR is now ready for another round of reviews.

amanda-tarafa:
I'll get to it early Monday, unless it's urgent. Let me know.

amanda-tarafa:
Let me know if you want me to force merge, the IAP failures are unrelated.
@skuruppu you should Approve as well, we now require 2 approvals in much cases and you are Spanner code owner ;).

skuruppu:
Thanks so much @amanda-tarafa for the review and @laljikanjareeya for the work. I did another review and approved it. Another CI build is running. If you can do a force-merge, that would be great @amanda-tarafa.

skuruppu:
@laljikanjareeya would you please be able to take a look at the failing spanner test? I took a quick look at the logs but couldn't figure out what exactly the failure was. It only started happening once I merged in the latest master.

amanda-tarafa:
It seems to be GoogleCloudSamples.Spanner.SpannerTests.TestGetBackupOperations timing out on Windows only. I've seen that you increased the timeout already.
If you keep increasing the timeout, then tests are going to potentially take a long time to be done, and we run these on presubmit for every PR (they already take longer than what I'd like tbh).

That the backup was actually created is testing that the service works properly, and that's good to do, if it doesn't costs us too much. I wonder if there's a way in which we could test that the sample is working as expected without having to wait for the actual backup to be created? Isn't the fact that we get an operation back enough proof that the sample is OK? @skuruppu I leave that to your consideration, if you feel that we should wait for the backup some more, then feel free to increase the timeout and let's see how long it takes. Else, if we can find an alternative way to test the sample, that'd be great.

skuruppu:
Thanks for the suggestion @amanda-tarafa. It is reasonable although it would mean that we will have to remove some of the other samples tests that we added that depends on it. E.g. to restore a database from a backup, we will need a completed backup. Also probably for some of the list backup calls.

On the other hand, I'm also reluctant to increase the timeout especially since the tests are run when other folks create PRs. Would there be a way in the future where only the tests for the changed API can be run?

Ideally, the backend would complete backups within a reasonable time for an essentially empty database but I don't think that will change any time soon.

amanda-tarafa:
For other samples tests, what about if we keep a known backup live. On test startup you could check if the known backup exists and if it doesn't, just create it (we would only need to wait for that one backup to be created in the unlikely case that someone has deleted it). Does this sound OK?

> Would there be a way in the future where only the tests for the changed API can be run?

This is an ongoing discussion, the problem is that in very rare cases, apparently, samples for one API are used in samples for another API. I would also favor only executing the affected tests but we might want to be consistent across languages on this one. I wouldn't count on this being decided soon, thouhg.

skuruppu:
That's a great idea.

@laljikanjareeya would you please be able to update the tests to create a rely on a known backup? You can do one run where you create a backup that you never delete, then you can add the name of it to the test.

skuruppu:
Thanks @laljikanjareeya. I made a few changes to the setup to not update the samples to do the already exists check. It's possible that tests will still timeout due to the restore sample taking too long to run. If I find that, I'll remove that test. But otherwise, with the new setup, the tests are super fast.

skuruppu:
For reasons I don't understand yet, the tests are still timing out. Based on the output, I suspect that it's actually creating the sample database that's timing out.

```
Spanner createSampleDatabase dotnet-docs-samples-tests my-instance my-db-vhixdjawmmjdrrzkmcke
QuickStart
[xUnit.net 00:00:01.84]     GoogleCloudSamples.Spanner.SpannerTests.TestGetBackupOperations [FAIL]
BUILD FAILED
```

amanda-tarafa:
OK, I think I may have found what's happening. I looked at the nightly run and Spanner tests have been failing since 28 April almost consistently, almost always with a timeout, but the first time they failed was because database quota had been reached. I went into the instance that we use for testing and we do have 100 databases there, which is quota. I think they were all created by this PR, because I tried to delete them from the UI and got a message saying that they have backups and cannot be deleted until all backups are deleted.
I can't go into each of the database to delete the backups manually and then delete the database, but you can probably write some code that does that. That would be a one off, to get rid of all those databases, and then make sure that this PR cleansup after itself.
You can delete all databases starting with "my-db-".

laljikanjareeya:
@skuruppu  I have created repository to delete all the backups and databases [here](https://github.com/laljikanjareeya/spanner-db-cleanup) can you run it on server.

skuruppu:
Thanks so much for the script @laljikanjareeya. I ran it and the old backups and databases are cleared. Now going to attempt to run the tests again.

skuruppu:
The tests still failed with the output:
```
Spanner createSampleDatabase dotnet-docs-samples-tests my-instance my-db-vezrjylexevmtplvfppk
QuickStart
Spanner insertSampleData dotnet-docs-samples-tests my-instance my-db-vezrjylexevmtplvfppk
[xUnit.net 00:00:44.83]     GoogleCloudSamples.Spanner.QuickStartTests.TestQuickStart [FAIL]
BUILD FAILED
```

It's still not clear to me in what part of the code the tests are timing out. So to rule out that it's in the backup creation, I'm now manually creating the test database and backup through the UI. Then I'll do another run.

Based on the output though, I doubt that it's in the backup code at all because there's a whole lot of other setup steps between `createSampleDatabase` and `createBackup` that doesn't seem to get run.

laljikanjareeya:
@skuruppu I think we need to change the `QuickStart` Program as it Query on `my-instance` instance and `my-database` database but it does not check for the instance and database is exists or not.
I will change it to Check first and if not than create It.

skuruppu:
> @skuruppu I think we need to change the `QuickStart` Program as it Query on `my-instance` instance and `my-database` database but it does not check for the instance and database is exists or not.
> I will change it to Check first and if not than create It.

Thanks @laljikanjareeya. Unfortunately we can't really change the samples themselves, since they're meant to be consistent across languages. I suppose you can try to modify the test to make sure it's created. That should be fine.

skuruppu:
I manually created `my-database` for now so hopefully it'll work.

laljikanjareeya:
@skuruppu Now All the test cases for Spanner is Passing successfully.

skuruppu:
Thanks so much @laljikanjareeya :)

skuruppu:
@amanda-tarafa, I think we'll need a force-merge from you. The two test failures are related to language and monitoring api test failures.

PR #1001, opened 2020-04-01
Dependency update and cleanup.

amanda-tarafa:
Fixes #988

PR #1005, opened 2020-04-02
Deletes IRM samples, no IRM client library has been released.

### LINKED ISSUE #1000: IRM samples: Remove samples
We have never released a package and only generated an alpha library.

Leftover from #988

amanda-tarafa:
Fixes #1000


amanda-tarafa:
Thanks. Will merge, CI is failing because of #988 and #1001 is on the way.

Issue #1013, opened 2020-04-06
Move samples to .NET Core 3.1 and C# 8


amanda-tarafa:

amanda-tarafa:
We are currently supporting .NET Core App 2.1, 2.2 and 3.1. Moving forward we'll only support 2.1 andd 3.1 which are the ones wich MS will keep support for in the long term.
For removing 2.2, we'll need to update some samples.

PR #1020, opened 2020-04-10
KMS: Add new samples and adopt new format

sethvargo:
This updates the Cloud KMS samples to use the new sample format, layout, and testing. Please review, but do not merge, because we have to coordinate a merge across the other languages.

This PR removes all the old samples and creates two new projects under a single solution: `Kms.Samples` and `Kms.Samples.Tests`. Each file is dedicated to a single file, and the snippet includes any import statements. There are no external dependencies besides KMS, and everything has been updated to .NET Core 3.1.

- Closes #916 
- References #1013


kms/api/Kms.Samples/VerifyAsymmetricEc.cs
```diff
@@ -0,0 +1,39 @@
...
+
+public class VerifyAsymmetricSignatureEcSample
+{
+
+    public bool VerifyAsymmetricSignatureEc(
+      string projectId = "my-project", string locationId = "us-east1", string keyRingId = "my-key-ring", string keyId = "my-key", string keyVersionId = "123",
+      string message = "my message",
+      byte[] signature = null)
+    {
+        // Cloud KMS returns signatures in a DER-encoded format. .NET requires
```
sethvargo, line 28: I'll pre-empt a question here. I'd really really really prefer to include this. We get a lot of feedback and bug reports on "please include a sample for X". Often times, we haven't _forgotten_ to include a sample, but it's actually not possible with the current language primitives (or would require hundreds of lines of code, 800 in this case). I'd rather users see this warning than think we forgot.

amanda-tarafa: This is totally fine by me. I wouldn't usually mess with the content of a sample unless I find it too weird.
Only thing I would do is remove the method signature, the class etc. Basically just have a text file with the tags and the comment. If it need to be a .cs file then leave the comments as //, else just a .txt file with this text in between tags. Less code to maitan or that people can add their own things inadvertently.

kms/api/Kms.Samples.Tests/CreateKeyAsymmetricDecryptTests.cs
```diff
@@ -0,0 +1,52 @@
...
+        _fixture = fixture;
+        _sample = new CreateKeyAsymmetricDecryptSample();
+    }
+
+    [Fact]
+    public void CreatesKey()
+    {
+        // Run the sample code.
+        var result = _sample.CreateKeyAsymmetricDecrypt(
+          projectId: _fixture.ProjectId, locationId: _fixture.LocationId, keyRingId: _fixture.KeyRingId,
```
amanda-tarafa, line 37: nit: I'm pretty sure this should be 4 spaces.

kms/api/Kms.Samples.Tests/CreateKeyAsymmetricDecryptTests.cs
```diff
@@ -0,0 +1,52 @@
...
+          id: _fixture.RandomId());
+
+        // Get the key.
+        var client = KeyManagementServiceClient.Create();
+        var key = client.GetCryptoKey(new GetCryptoKeyRequest
```
amanda-tarafa, line 42: You can be more relaxed about var in tests, although the rule of thumb for C# development is to only use var if it is very clear what the type is.

kms/api/Kms.Samples/CreateKeyAsymmetricDecrypt.cs
```diff
@@ -0,0 +1,52 @@
...
+                {
+                    Algorithm = CryptoKeyVersion.Types.CryptoKeyVersionAlgorithm.RsaDecryptOaep2048Sha256,
+                }
+            }
+        };
+
+        // Call the API.
+        var result = client.CreateCryptoKey(request);
+
+        // Return the result.
```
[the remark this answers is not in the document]
sethvargo: The feedback we've gotten on samples for KMS is that customers aren't running these samples, so they really don't care about output.

kms/api/Kms.Samples.Tests/KmsFixture.cs
```diff
@@ -0,0 +1,265 @@
...
+        var client = KeyManagementServiceClient.Create();
+
+        var result = client.CreateCryptoKeyVersion(new CreateCryptoKeyVersionRequest
+        {
+            ParentAsCryptoKeyName = new CryptoKeyName(this.ProjectId, this.LocationId, this.KeyRingId, keyId),
+        });
+
+        for (var i = 1; i <= 5; i++)
+        {
+            var version = client.GetCryptoKeyVersion(new GetCryptoKeyVersionRequest
```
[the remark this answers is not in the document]
sethvargo: Yea, sadly we're stuck with the loop for now

kms/api/Kms.Samples.Tests/CreateKeyAsymmetricDecryptTests.cs
sethvargo: Done

kms/api/Kms.Samples.Tests/KmsFixture.cs
```diff
@@ -0,0 +1,265 @@
...
+                {
+                    CryptoKeyVersionName = keyVersion.CryptoKeyVersionName,
+                });
+            }
+        }
+    }
+
+    public string RandomId()
+    {
+        return $"csharp-{System.Guid.NewGuid()}";
```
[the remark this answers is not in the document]
sethvargo: Fortunately KMS supports 64 characters for resource names

kms/api/Kms.Samples.Tests/KmsFixture.cs
```diff
@@ -0,0 +1,265 @@
...
+
+    public readonly string AsymmetricSignRsaKeyId;
+    public readonly CryptoKeyName AsymmetricSignRsaKeyName;
+
+    public readonly string SymmetricKeyId;
+    public readonly CryptoKeyName SymmetricKeyName;
+
+    public KmsFixture()
+    {
+        this.ProjectId = Environment.GetEnvironmentVariable("GOOGLE_PROJECT_ID");
```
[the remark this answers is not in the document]
sethvargo: Done

kms/api/Kms.Samples.Tests/KmsFixture.cs
```diff
@@ -0,0 +1,265 @@
...
+using Xunit;
+
+using Google.Api.Gax.ResourceNames;
+using Google.Cloud.Kms.V1;
+using Google.Protobuf.WellKnownTypes;
+
+[CollectionDefinition(nameof(KmsFixture))]
+public class KmsFixture : IDisposable, ICollectionFixture<KmsFixture>
+{
+    public readonly string ProjectId;
```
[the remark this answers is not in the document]
sethvargo: Done

kms/api/Kms.Samples/README.md
```diff
@@ -0,0 +1,32 @@
+# .NET Cloud Key Management Service (KMS) Samples
+
+These samples demonstrate how to interact with the [Cloud KMS][kms] using the
+Cloud KMS API using .NET in C#.
```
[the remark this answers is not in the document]
sethvargo: Done

kms/api/Kms.Samples/GetPublicKey.cs
```diff
@@ -0,0 +1,46 @@
...
+        var request = new GetPublicKeyRequest
+        {
+            CryptoKeyVersionName = new CryptoKeyVersionName(projectId, locationId, keyRingId, keyId, keyVersionId),
+        };
+
+        // Call the API.
+        var result = client.GetPublicKey(request);
+
+        // Extract and print the public key in PEM format.
+        Console.WriteLine(result.Pem);
```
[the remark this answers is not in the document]
sethvargo: Yea, I actually removed it 😄

kms/api/Kms.Samples/GetKeyLabels.cs
```diff
@@ -0,0 +1,50 @@
...
+        {
+            CryptoKeyName = new CryptoKeyName(projectId, locationId, keyRingId, keyId),
+        };
+
+        // Call the API.
+        var result = client.GetCryptoKey(request);
+
+        // Extract and print labels.
+        var labels = result.Labels;
+        foreach (var item in labels)
```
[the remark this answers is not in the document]
sethvargo: Done

kms/api/Kms.Samples/EncryptAsymmetric.cs
```diff
@@ -0,0 +1,52 @@
...
+        // Split the key into blocks and base64-decode the PEM parts.
+        var blocks = publicKey.Pem.Split("-", StringSplitOptions.RemoveEmptyEntries);
+        var pem = Convert.FromBase64String(blocks[1]);
+
+        // Create a new RSA key.
+        var rsa = RSA.Create();
+        rsa.ImportSubjectPublicKeyInfo(pem, out _);
+
+        // Encrypt the data.
+        var ciphertext = rsa.Encrypt(Encoding.UTF8.GetBytes(plaintext), RSAEncryptionPadding.OaepSHA256);
```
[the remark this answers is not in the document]
sethvargo: Done

kms/api/Kms.Samples/EncryptAsymmetric.cs
```diff
@@ -0,0 +1,52 @@
...
+        var client = KeyManagementServiceClient.Create();
+
+        // Get the public key.
+        var publicKey = client.GetPublicKey(new GetPublicKeyRequest
+        {
+            CryptoKeyVersionName = new CryptoKeyVersionName(projectId, locationId, keyRingId, keyId, keyVersionId),
+        });
+
+        // Split the key into blocks and base64-decode the PEM parts.
+        var blocks = publicKey.Pem.Split("-", StringSplitOptions.RemoveEmptyEntries);
```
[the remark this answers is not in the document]
sethvargo: Done

kms/api/Kms.Samples/EncryptAsymmetric.cs
```diff
@@ -0,0 +1,52 @@
...
+
+    public byte[] EncryptAsymmetric(
+      string projectId = "my-project", string locationId = "us-east1", string keyRingId = "my-key-ring", string keyId = "my-key", string keyVersionId = "123",
+      string plaintext = "Sample message")
+    {
+        // Create the client.
+        var client = KeyManagementServiceClient.Create();
+
+        // Get the public key.
+        var publicKey = client.GetPublicKey(new GetPublicKeyRequest
```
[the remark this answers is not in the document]
sethvargo: Done

kms/api/Kms.Samples/DestroyKeyVersion.cs
```diff
@@ -0,0 +1,41 @@
...
+ * limitations under the License.
+ */
+
+// [START kms_destroy_key_version]
+using Google.Cloud.Kms.V1;
+
+public class DestroyKeyVersionSample
+{
+
+    public CryptoKeyVersion DestroyKeyVersion(string projectId = "my-project", string locationId = "us-east1", string keyRingId = "my-key-ring", string keyId = "my-key", string keyVersionId = "123")
```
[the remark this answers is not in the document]
sethvargo: Done

kms/api/Kms.Samples/CreateKeyAsymmetricDecrypt.cs
```diff
@@ -0,0 +1,52 @@
...
+                Purpose = CryptoKey.Types.CryptoKeyPurpose.AsymmetricDecrypt,
+                VersionTemplate = new CryptoKeyVersionTemplate
+                {
+                    Algorithm = CryptoKeyVersion.Types.CryptoKeyVersionAlgorithm.RsaDecryptOaep2048Sha256,
+                }
+            }
+        };
+
+        // Call the API.
+        var result = client.CreateCryptoKey(request);
```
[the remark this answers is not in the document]
sethvargo: Done

kms/api/Kms.Samples/VerifyAsymmetricEc.cs
sethvargo: I have to keep the class (it gets sad otherwise), but I reduced it just to a comment.

kms/api/Kms.Samples/DecryptAsymmetric.cs
```diff
@@ -0,0 +1,45 @@
...
+// [START kms_decrypt_asymmetric]
+using Google.Protobuf;
+using Google.Cloud.Kms.V1;
+
+public class DecryptAsymmetricSample
+{
+
+    public string DecryptAsymmetric(
+      string projectId = "my-project", string locationId = "us-east1", string keyRingId = "my-key-ring", string keyId = "my-key", string keyVersionId = "123",
+      byte[] ciphertext = null)
```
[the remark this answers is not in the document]
sethvargo: So it seems to work. I mean... it compiles. It obviously fails because the ciphertext is null, but the API call succeeds.

kms/api/Kms.Samples.Tests/KmsFixture.cs
```diff
@@ -0,0 +1,265 @@
...
+            {
+                CryptoKeyVersionName = result.CryptoKeyVersionName,
+            });
+
+            if (version.State == CryptoKeyVersion.Types.CryptoKeyVersionState.Enabled)
+            {
+                return version;
+            }
+
+            Thread.Sleep(500 * i);
```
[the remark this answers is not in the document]
sethvargo: This is called relatively infrequently, so I'd prefer to keep it like this for now. If it becomes a performance bottleneck, we can always revisit, but the new KMS tests all run in parallel (because of the splitting into different files) and only take ~10s :smile.

kms/api/Kms.Samples/SignAsymmetric.cs
```diff
@@ -0,0 +1,56 @@
...
+        // Create the client.
+        KeyManagementServiceClient client = KeyManagementServiceClient.Create();
+
+        // Calculate the message digest.
+        SHA256 sha256 = SHA256.Create();
+        byte[] digest = sha256.ComputeHash(Encoding.UTF8.GetBytes(message));
+
+        // Build the request.
+        //
+        // Note: Key algorithms will require a varying hash function. For
```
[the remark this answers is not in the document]
sethvargo: Yea, that's correct. The type of key determines the type of hash it expects.

kms/api/Kms.Samples.Tests/IamAddMemberTests.cs
```diff
@@ -0,0 +1,54 @@
...
+        {
+            if (b.Role == "roles/cloudkms.cryptoKeyEncrypterDecrypter")
+            {
+                role = b;
+                break;
+            }
+        }
+
+        Assert.NotNull(role);
+        Assert.Contains("group:test@google.com", role.Members);
```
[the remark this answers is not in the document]
sethvargo: one

kms/api/Kms.Samples/CreateKeyAsymmetricDecrypt.cs
```diff
@@ -0,0 +1,51 @@
...
+ * limitations under the License.
+ */
+
+// [START kms_create_key_asymmetric_decrypt]
+using Google.Cloud.Kms.V1;
+
+public class CreateKeyAsymmetricDecryptSample
+{
+    public CryptoKey CreateKeyAsymmetricDecrypt(
+      string projectId = "my-project", string locationId = "us-east1", string keyRingId = "my-key-ring",
```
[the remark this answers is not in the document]
sethvargo: The linter doesn't seem to complain, and neither does the linter on CI. I'm happy to change it, but it seems like it won't be enforced since neither linter actually cares.

sethvargo:
Hey @amanda-tarafa CI is complaining about the .NET version. Any ideas?

```
The current .NET SDK does not support targeting .NET Core 3.1.
```

sethvargo:
Hmm @amanda-tarafa any ideas why it's still not picking up the new version on the builder from https://github.com/GoogleCloudPlatform/dotnet-docs-samples/pull/1022? Do we need to do something special?

amanda-tarafa:
Did you rebased after I merged the #1022?

sethvargo:
@amanda-tarafa yea I did. I just double checked and I'm off master.

```
$ git push -f sethvargo sethvargo/kms_samples
Everything up-to-date
```

amanda-tarafa:
@sethvargo  I've sent an internal email trying to figure this out. You are copied.

kms/api/Kms.Samples.Tests/CreateKeyLabelsTests.cs
```diff
@@ -0,0 +1,50 @@
...
+    {
+        _fixture = fixture;
+        _sample = new CreateKeyLabelsSample();
+    }
+
+    [Fact]
+    public void CreatesKey()
+    {
+        // Run the sample code.
+        var result = _sample.CreateKeyLabels(
```
amanda-tarafa, line 36: I just realized, you are not deleting what gets created on the samples that create something.
You can add methods in the fixture "MarkForDeletion" to which you pass the ID of what you created. Those are held in lists by the fixture and deleted on dispose.

sethvargo: I am, but it's a bit indirect. I can document it if that'd be helpful. Everything gets created in the single KeyRing. The KeyRing has a dispose method that deletes all the things inside of it at the end.

I've run the samples a bunch of times now against my project and there's no left-over resources.

amanda-tarafa: Ah that's perfect, and true, I remember seeing that you were iterating over everything in the key ring. Yes, not a problem then.
Maybe add a note at the beginning of the test class, just to make it easy.

sethvargo:
@amanda-tarafa looks like we need to grant the builder service account more permissions?

kms/api/Kms.Samples.Tests/CreateKeyAsymmetricDecryptTests.cs
```diff
@@ -0,0 +1,50 @@
...
+    public void CreatesKey()
+    {
+        // Run the sample code.
+        var result = _sample.CreateKeyAsymmetricDecrypt(
+            projectId: _fixture.ProjectId, locationId: _fixture.LocationId, keyRingId: _fixture.KeyRingId,
+            id: _fixture.RandomId());
+
+        // Get the key.
+        KeyManagementServiceClient client = KeyManagementServiceClient.Create();
+        var key = client.GetCryptoKey(new GetCryptoKeyRequest
```
bdhess, line 42: Why are we avoiding the convenience overloads? Is this now a deprecated feature of GAPIC?

I would write this as
`var key = client.GetCryptoKey(result.CryptoKeyName);`

(Question applies to all of these tests and samples)

kms/api/Kms.Samples/EncryptAsymmetric.cs
```diff
@@ -0,0 +1,52 @@
...
+using Google.Cloud.Kms.V1;
+using System;
+using System.Security.Cryptography;
+using System.Text;
+
+public class EncryptAsymmetricSample
+{
+    public byte[] EncryptAsymmetric(
+      string projectId = "my-project", string locationId = "us-east1", string keyRingId = "my-key-ring", string keyId = "my-key", string keyVersionId = "123",
+      string plaintext = "Sample message")
```
bdhess, line 28: The idea that cryptographic plaintexts are bytes, as opposed to text, is often confusing to to cryptography novices.  It pushes people down all kinds of strange roads (e.g. "I have bytes, but they need text, I guess I will base64-encode...")

I'd reserve the word 'plaintext' to refer to a byte array, and call out the transformation you're doing, like:

```
public byte[] EncryptAsymmetric(..., string message = "Sample message")
{
    // Cryptographic plaintexts are always byte arrays. Convert our message into UTF-8 bytes.
    byte[] plaintext = Encoding.UTF8.GetBytes(message);
    ...
```

kms/api/Kms.Samples/EncryptSymmetric.cs
```diff
@@ -0,0 +1,45 @@
...
+// [START kms_encrypt_symmetric]
+
+using Google.Cloud.Kms.V1;
+using Google.Protobuf;
+
+public class EncryptSymmetricSample
+{
+    public byte[] EncryptSymmetric(
+      string projectId = "my-project", string locationId = "us-east1", string keyRingId = "my-key-ring", string keyId = "my-key",
+      string plaintext = "Sample message")
```
bdhess, line 26: Same as before re: the term plaintext

kms/api/Kms.Samples/DecryptSymmetric.cs
```diff
@@ -0,0 +1,45 @@
...
+        {
+            CryptoKeyName = new CryptoKeyName(projectId, locationId, keyRingId, keyId),
+            Ciphertext = ByteString.CopyFrom(ciphertext),
+        };
+
+        // Call the API.
+        DecryptResponse result = client.Decrypt(request);
+
+        // Return the result.
+        return result.Plaintext.ToStringUtf8();
```
bdhess, line 42: Same comment as above

kms/api/Kms.Samples/DecryptAsymmetric.cs
```diff
@@ -0,0 +1,45 @@
...
+        {
+            CryptoKeyVersionName = new CryptoKeyVersionName(projectId, locationId, keyRingId, keyId, keyVersionId),
+            Ciphertext = ByteString.CopyFrom(ciphertext),
+        };
+
+        // Call the API.
+        AsymmetricDecryptResponse result = client.AsymmetricDecrypt(request);
+
+        // Return the result.
+        return result.Plaintext.ToStringUtf8();
```
bdhess, line 42: I would show this in two steps:
```
byte[] plaintext = result.Plaintext.ToByteArray();
return Encoding.UTF8.GetString(plaintext);
```

That way it's obvious to a developer what they should be doing if their plaintext is actually binary data (which is probably the majority of cases).

kms/api/Kms.Samples/SignAsymmetric.cs
```diff
@@ -0,0 +1,57 @@
...
+    public byte[] SignAsymmetric(
+      string projectId = "my-project", string locationId = "us-east1", string keyRingId = "my-key-ring", string keyId = "my-key", string keyVersionId = "123",
+      string message = "Sample message")
+    {
+        // Create the client.
+        KeyManagementServiceClient client = KeyManagementServiceClient.Create();
+
+        // Calculate the message digest.
+        SHA256 sha256 = SHA256.Create();
+        byte[] digest = sha256.ComputeHash(Encoding.UTF8.GetBytes(message));
```
bdhess, line 35: Let's convert the text to UTF-8 bytes earlier and separately.  Same family of concerns as previous comments about making it very clear that our interfaces are designed to accept binary data.

kms/api/Kms.Samples/VerifyAsymmetricEc.cs
```diff
@@ -0,0 +1,29 @@
...
+ * limitations under the License.
+ */
+
+// [START kms_verify_asymmetric_signature_ec]
+
+public class VerifyAsymmetricSignatureEcSample
+{
+    // Cloud KMS returns signatures in a DER-encoded format. .NET requires
+    // signatures to be in IEEE 1363 format, and converting between these formats
+    // is a few hundred lines of code.
```
bdhess, line 23: I guess the idea of doing this in BC is totally out?
I think it would be nice to show it, even if it adds a dependency.  We can always move to System.Security.Cryptography when the functionality is available there.

We can include the `using` statement within the scope of the START/END blocks and add comments to make it clear that a third party library is required.

kms/api/Kms.Samples.Tests/CreateKeyAsymmetricDecryptTests.cs
amanda-tarafa: If the protos are correctly annotated then we should be generating these overloads. I agree that if they are there, it's better to always use the simplest one.

kms/api/Kms.Samples/DecryptSymmetric.cs
sethvargo: Done

kms/api/Kms.Samples/EncryptSymmetric.cs
sethvargo: Done

kms/api/Kms.Samples/EncryptAsymmetric.cs
sethvargo: Done

kms/api/Kms.Samples/DecryptAsymmetric.cs
sethvargo: Done

kms/api/Kms.Samples/SignAsymmetric.cs
sethvargo: Done

kms/api/Kms.Samples/VerifyAsymmetricEc.cs
sethvargo: While the refactored samples would make it very clear that we're depending on BC, I'm against encouraging users to take a third-party dependency on software we do not control or audit.

kms/api/Kms.Samples.Tests/CreateKeyAsymmetricDecryptTests.cs
sethvargo: They exist, but I intentionally avoided them. The other statically-typed languages (Go and Java) both prefer the "construct and accept the entire request object" pattern, and I tried to be consistent here. I've also heard feedback that the magical methods that accept arguments often aren't flexible enough for advanced use cases, and customers have to drop down to this way anyway.

It's not a hill upon which I stake my claim, but it feels good to be consistent?

amanda-tarafa: I do like the consistency aspect of using the request object. But I do think that when the overloads are enough they should be used for simplicity. If there's a complex use case that is not covered by the overloads then I would suggest writing a sample that covers that complex use case and shows how to create and populate the request object accordingly.

sethvargo: Ack, updated to use overloads.

sethvargo:
@amanda-tarafa good news - looks like all the KMS tests are passing! IAM is still upset, but progress!

We still need to hold off on merging this until the docs changes are ready though.

amanda-tarafa:
Perfect! IAM is not a worry, I've checked the failures and it's just weekly quota. I probably blew it off over the weekend running tests to get the CI environment right.
Fine with keeping this around until docs ready.
Thanks!!

amanda-tarafa:
Force merging, CI failures unrelated.

PR #1026, opened 2020-04-20
Makes the Docker file used for testing support .NET Core 3.1

amanda-tarafa:
Towards #1013 .
Superseeds #1022 which was very "optimistic" in its approach.


amanda-tarafa:
Force merging. The CI errors are just because IAM is reaching role quota, nothing to do with updating the supported .NET versions.

PR #1031, opened 2020-04-21
Update Cloud SQL MySQL sample for Cloud Run

jsimonweb:


cloud-sql/mysql/Program.cs
```diff
@@ -14,21 +14,56 @@
...
+            BuildWebHost(args).Build().Run();
+        }
+        
+        public static IWebHostBuilder BuildWebHost(string[] args)
+        {
+            ReadAppSettings();
+
+            string port = Environment.GetEnvironmentVariable("PORT") ?? "8080";
+            string hostingUrl = $"http://0.0.0.0:{port}";
+            string localhostUrl = "http://localhost:5567";
```
SurferJeffAtGoogle, line 44: Should this be`$"http://localhost:port"`?

jsimonweb: Reusing the same port causes a "System.Net.Sockets.SocketException: Address already in use" error in Cloud Run and App Engine. By using "localhost:5567" the same URL can be used for both running locally and for Run-KestralTest.

SurferJeffAtGoogle: Ok.  Please add a comment to that effect.

SurferJeffAtGoogle: Oh, I think I've seen this problem before.  The issue with hard-coding 5567 is it doesn't respect what the user sets on the command line or via environment variables or in config files.

Program.cs in the getting-started samples demonstrates a solution that still respects all the other ways to set the port number.
https://github.com/GoogleCloudPlatform/getting-started-dotnet/blob/master/HelloWorld/Program.cs

cloud-sql/mysql/Program.cs
```diff
@@ -14,21 +14,68 @@
...
+
+    static class ProgramExtensions
+    {
+        // Google Cloud Run sets the PORT environment variable to tell this
+        // process which port to listen to.
+        public static IWebHostBuilder UsePortEnvironmentVariable(
+            this IWebHostBuilder builder)
+        {
+            string port = Environment.GetEnvironmentVariable("PORT");
+            if (!string.IsNullOrEmpty(port))
```
amanda-tarafa, line 74: Should you add and `else` here that uses `8080` if the port is not set on an environment variable?

jsimonweb: I just tested updating this line to be:
string port = Environment.GetEnvironmentVariable("PORT") ?? "8080";

When I start the cloudsql proxy locally and run the program after that change it starts up successfully with the command line message:

"Application started. Press Ctrl+C to shut down.
Now listening on: http://0.0.0.0:8080"

But when I visit "http://0.0.0.0:8080" in a browser I get the error message:

"This site can’t be reachedThe webpage at http://0.0.0.0:8080/ might be temporarily down or it may have moved permanently to a new web address. ERR_ADDRESS_INVALID"


Changing the line back to: 
string port = Environment.GetEnvironmentVariable("PORT");

When I start the cloudsql proxy locally and run the program after that change it
starts up successfully with the command line message:

"Now listening on: http://localhost:5000
Now listening on: https://localhost:5001
Application started. Press Ctrl+C to shut down."

Now when visit "http://localhost:5000" or "https://localhost:5001/" in a browser I see the web app load correctly in a functional state.

I was also able to successfully deploy and run the app (with the current port setting command) to Cloud Run and App Engine.

This is also currently being used for the getting-started sample app:
https://github.com/GoogleCloudPlatform/getting-started-dotnet/blob/master/HelloWorld/Program.cs

Please force merge if everything else looks okay. Thanks!

amanda-tarafa: OK! That sounds like something else is setting a default port then which is fine.
I'll let tests run now (I've deactivated IAP) and then you can merge on green.

amanda-tarafa:
Force merging, failures are unrelated.

PR #1036, opened 2020-04-24
[IAP]: Updates IAP samples.

amanda-tarafa:


amanda-tarafa:
Force merging since the failures have nothing to do with IAP.

PR #1037, opened 2020-04-24
fix: use DELETE FROM for consistency

skuruppu:


skuruppu:
If you can do a force-merge, that would be great. Thanks @amanda-tarafa.

Issue #1040, opened 2020-04-27
IAP: Skipping tests because they have started faling without changes on our side.


amanda-tarafa:
I'm looking into it.

amanda-tarafa:
Take a look at https://github.com/googleapis/google-api-dotnet-client/issues/1557 when fixing this. Might be related.

PR #1041, opened 2020-04-27
[IAP]: Skips IAP testing because the application it was being tested against has dissapeared.

amanda-tarafa:
See #1040


amanda-tarafa:
Force merging, this is deactivating tests that are failing and has and accompanying tracking issue.

PR #1045, opened 2020-05-06
Update CODEOWNERS

czahedi:
Adding storage team


CODEOWNERS
```diff
@@ -4,7 +4,7 @@ bigquery/   @jsimonweb
 bigtable/   @jsimonweb
```
amanda-tarafa, line 4: Shouldn't these be set the same as you did in https://github.com/googleapis/google-cloud-dotnet/pull/4947?

czahedi: @crwilcox to confirm, I think we hadn't received info from other teams about which should be included

czahedi: Yep going to add those.

PR #1047, opened 2020-05-08
fix(spanner): UpdateBackup test Expire time issue

### LINKED ISSUE #1046: Spanner: UpdateBackup test is failing
`
Error: Grpc.Core.RpcException: 'Status(StatusCode=InvalidArgument, Detail="Invalid request for backup projects/project-id/instances/instance-id/backups/backup-id. `expire_time` is either missing, in the wrong format, or is not within the minimum and maximum expire times supported. `expire_time` must be a Unix Epoch Time (UTC time since 1970-01-01 00:00:00) with microseconds granularity, and be at least 6 hours and at most 366 days from the time the CreateBackup request is processed. The CreateBackup request was processed at `create_time` (2020-05-05T04:42:13.754175Z), and the input `expire_time` is 2021-05-17T14:00:00Z.")
`

[Here ](https://source.cloud.google.com/results/invocations/0964d5f4-1007-4551-a11b-8371b97fd8de/targets/github%2Fdotnet-docs-samples%2Fspanner%2Fapi%2FSpannerTest%2FTestResults/tests) is Build Logs.

laljikanjareeya:
Fixes #1046


laljikanjareeya:
@skuruppu Looks like `TestGetBackups` is also failing. I have modify that as well `count` of backupId is greater than or equal to 5 as `create_time` and `expire_time` may not capture it.

skuruppu:
@laljikanjareeya would you be able to look into the spanner test timeout. I've run the presubmits a few times now and keep the [same error](https://source.cloud.google.com/results/invocations/44a2cb43-405f-46b1-b29e-3dbab7ae3539/targets/github%2Fdotnet-docs-samples%2Fspanner%2Fapi%2FSpannerTest/tests).

spanner/api/SpannerTest/Tests.cs
```diff
@@ -749,9 +748,9 @@ void TestGetBackups()
             Assert.Equal(0, output.ExitCode);
             // BackupId should be a result of each of the 7 ListBackups calls and
```
amanda-tarafa, line 749: Does this depends on the order the tests are run? If it's so maybe don't change on this PR but definetely let's create an issue to fix this.

amanda-tarafa:
@laljikanjareeya also, if these tests are not using the JUnit logger, can you please add the dependency and the parameter on the script, so that if they fail we get more info? Thanks.

amanda-tarafa:
Spanner is already using the JUnit logger, still, we don't get good diagnostics info when tests fail. I wonder if it's related to the command runner.

spanner/api/SpannerTest/Tests.cs
skuruppu: It shouldn't depend on the order. It's just that there are two ListBackups calls that filter on the create_time and expire_time. Because we decided to reuse an existing backup rather than create a new one each time the test runs, the create_time and expire_time are no longer predictable to test for. This is why the check is for less than 7.

amanda-tarafa:
Again [xUnit.net 00:00:02.54]     GoogleCloudSamples.Spanner.SpannerTests.TestGetBackupOperations [FAIL]

(And also some Logging tests that timed out that I'll skip now in a separate PR).

Would you consider deactivating TestGetBackupOperations on this PR so that it doesn't affect presubmits for other PRs and then creating a PR for fixing whatever it's wrong? That's what I've been doing with all the flaky tests.

skuruppu:
All done. Fingers crossed that the tests will pass.

amanda-tarafa:
Thanks, I'm deactivating some others that are flaky as well so hopefully we can have green presubmits while I fix the rest. (Fine to merge this if the failing tests are unrelated.)

laljikanjareeya:
@amanda-tarafa It's difficult to figure out what's going on as it doesn't get diagnostics, Let me know if I am missing anything in configuration part of test.

amanda-tarafa:
> @amanda-tarafa It's difficult to figure out what's going on as it doesn't get diagnostics, Let me know if I am missing anything in configuration part of test.

Yes, I agree, I'm not 100% certain but I believe is the use of the command runner that garbles the diagnostics info. Unfortunately that would require a rewrite of all tests. Let's start slow, if this one fails again, or any other Spanner test, let's rewrite that one as per the KMS canonical set, remove the command runner, and see what kind of info we get. If we confirm then we can start rewriting the rest (not just Spanner, but a lot of the others are similar).

amanda-tarafa:
It's now GoogleCloudSamples.Spanner.SpannerTests.TestQueryWithBytes [FAIL]

laljikanjareeya:
@amanda-tarafa  can we deactivate `TestQueryWithBytes` test in this PR and let's check if everything goes well, will create separate issue for same for tracking.
I am guessing problem is somewhere else.

amanda-tarafa:
Yes, please do so.

amanda-tarafa:
Monitoring and Logging failing so far, which I'm dealing with in #1067

amanda-tarafa:
GoogleCloudSamples.Spanner.SpannerTests.TestReadWriteUnderFundedTransaction [FAIL]

Fine by me with skipping this one as well, just to see if we can get this PR in. I'm pretty sure as well that the problem is not the individual tests but something general, either the infrastructure, quota, etc. These tests are just the ones that happen to fail. Still, maybe we can skip enough of them so that we get a green.

An alternative is to deactivate all of Spanner samples test and start on a rewrite. @skuruppu what do you think? My only issue with keeping them as is while we fix them is that it affects other's presubmits and then I have to force merge.

laljikanjareeya:
@amanda-tarafa I reviewed it and it always failing first test, seems like issue in instance setup.
I have added setup for instance as well if it doesn't exists will create before Initializing Database.

laljikanjareeya:
Now It's giving [TestResult](https://source.cloud.google.com/results/invocations/1f712c12-8efd-4120-9641-746372d32e3f/targets/github%2Fdotnet-docs-samples%2Fspanner%2Fapi%2FSpannerTest%2FTestResults/tests) and limit has been reached on maximum database(100) on the instance.

@amanda-tarafa Does this instance shared with other language's CI?  because we are cleaning up everything in C#.

Another solution is we can use separate instance for C# so it doesn't reach the limit.

 ```
Error: System.AggregateException : One or more errors occurred. (Resources have been exhausted.)
---- Google.Cloud.Spanner.Data.SpannerException : Resources have been exhausted.
-------- Grpc.Core.RpcException : Status(StatusCode=ResourceExhausted, Detail="Unable to create database 'my-db-bisljpehcqccixaecurq' because the instance 'projects/dotnet-docs-samples-tests/instances/my-instance' has already reached the maximum database limit (100). Please delete a database in the instance and try again, or choose a different instance.") 
```

amanda-tarafa:
No, we are not sharing this with other languages. Do you clean up also when the test fails or timeout?
I'll go in manually now and clean all test instances. There might be left over from way back.

laljikanjareeya:
Another thing I notice CI is using `projects/dotnet-docs-samples-tests/instances/my-instance` instance so environment variable for instance i.e `TEST_SPANNER_INSTANCE` not set on CI.

laljikanjareeya:
> I'll go in manually now and clean all test instances. There might be left over from way back.

you can use https://github.com/laljikanjareeya/spanner-db-cleanup Program to delete all the backups and database on instance.

amanda-tarafa:
> Another thing I notice CI is using `projects/dotnet-docs-samples-tests/instances/my-instance` instance so environment variable for instance i.e `TEST_SPANNER_INSTANCE` not set on CI.

I can look into that later as well. Although I think that env var is used mostly for when we run the tests locally in our personal projects.

laljikanjareeya:
> Do you clean up also when the test fails or timeout.
I will take a look on same.

amanda-tarafa:
Are you deleting the databases? There were 72 when I went in and without me doing anything, there are now like 10.
If you are deleting them then there's some cases in which the databases are not deleted after tests.
If not this is just a problem of the databases being mark for deletion and deleted at some point after, which is less worrisome. And we've been running the tests today like crazy.

Also, how many databases do the tests create in each run. Are there tests that could reuse a single database? Just something to take into account when attempting to fix things.

laljikanjareeya:
> Are you deleting the databases? There were 72 when I went in and without me doing anything, there are now like 10.

No I don't have any access on CI.

laljikanjareeya:
> Also, how many databases do the tests create in each run.

we are creating three databases in test setup:
1. Quick start.
2. Database query and operation tests.
3. Backup tests.

And on Dispose we are deleting all three databases.

amanda-tarafa:
> > Are you deleting the databases? There were 72 when I went in and without me doing anything, there are now like 10.
> 
> No I don't have any access on CI.

OK, so they probably were correctly marked for deletion, and only deleted a little afterwards? Not really sure what is happening here. 
I will run the tests again and see what happens.

And see [this error](https://source.cloud.google.com/results/invocations/36ccda41-0423-4820-99ee-de34a2721c4a/targets/github%2Fdotnet-docs-samples%2Fspanner%2Fapi%2FSpannerTest%2FTestResults/tests) that I got on #1067.

amanda-tarafa:
It's green after forcing Kokoro. I'll merge now but there's something going on with the Spanner tests that we'll have to look at.

PR #1053, opened 2020-05-26
Added Cloud Run Events C# samples

meteatamel:


run/events/README.md
```diff
@@ -0,0 +1,35 @@
...
+
+```sh
+MY_RUN_SERVICE_URL=$(gcloud run services describe "$MY_RUN_SERVICE" --format 'value(status.address.url)')
+curl $MY_RUN_SERVICE_URL
+```
+
+Test with a Cloud Event:
+
+```sh
+curl $MY_RUN_SERVICE_URL \
```
jskeet, line 24: The examples I've seen typically use all lower-case for the CloudEvent headers (`ce-id`, `ce-type` etc).

run/events/.gitignore
```diff
@@ -0,0 +1,332 @@
+## Ignore Visual Studio temporary files, build results, and
```
jskeet, line 1: Do we need a .gitignore file specifically for this directory? It looks like we do have about 12 of them, but I don't know whether that's intended to be the way forward or not.
@amanda-tarafa what are your thoughts?

run/events/Startup.cs
```diff
@@ -0,0 +1,54 @@
...
+        public void Configure(IApplicationBuilder app, IWebHostEnvironment env)
+        {
+            if (env.IsDevelopment())
+            {
+                app.UseDeveloperExceptionPage();
+            }
+
+            app.UseRouting();
+
+            app.UseEndpoints(endpoints =>
```
jskeet, line 38: I'd probably add a comment here for folks who've used ASP.NET Core in an MVC or WebAPI setting before, but haven't seen this way of responding to requests.

meteatamel: What sort of comment you had in mind? In ASP.NET Core 3.1, this is the default template you get when you run `dotnet new web`

run/events/.gitignore
meteatamel: I see, I didn't realize that. I can remove .gitignore then, it's not needed here.

run/events/README.md
meteatamel: Ok, changed README to have lowercase CloudEvent headers

run/events/Startup.cs
jskeet: Just something around "Sample responses to GET and POST requests" or the like.
I suspect most ASP.NET Core developers stick to MVC, Razor pages or WebAPI most of the time - this is probably lower level than they're used to.

meteatamel: Ok, I added a couple minor comments

run/events/Program.cs
```diff
@@ -0,0 +1,39 @@
...
+    {
+        public static void Main(string[] args)
+        {
+            CreateHostBuilder(args).Build().Run();
+        }
+
+        public static IHostBuilder CreateHostBuilder(string[] args)
+        {
+            string port = Environment.GetEnvironmentVariable("PORT") ?? "8080";
+            string url = String.Concat("http://0.0.0.0:", port);
```
jskeet, line 30: Just noticed this - any reason for using `String.Concat` explicitly rather than either:

```csharp
string url = $"http://0.0.0.0:{port}";`
```

or

```csharp
string url = "http://0.0.0.0" + port;`
```

? Not a blocker - just interested.

meteatamel: No special reason. Probably a legacy from Knative Serving C# sample.

PR #1055, opened 2020-06-01
Updates CODEOWNERS.

amanda-tarafa:
* Sets idiomacy reviewers on all samples.
  * Sets owners of some new samples.

@jskeet I've added you mostly for when I'm not available, you can ignore review requests for sample PRs otherwise.

### DIFF
--- a/CODEOWNERS
+++ b/CODEOWNERS
@@ -1,20 +1,34 @@
-*           @amanda-tarafa
+# There are two aspects that should be taken into account when reviewing samples on this repo.
+#   * That the samples are meaningful and consistent with similar samples in other languages
+#     and with what is stated in documentation. Some sample sets have specific owners that
+#     are in charge of this aspect.
+#   * The extent to which the sample is idiomatic C# and consistent with the rest of the samples in this repo.
+#     Currently @amanda-tarafa, @jskeet, in that order of availability, are in charge of this aspect.
+# Any PR will require two reviewers, one for each of the aspects mentioned above. For those sample sets
+# that don't have a specific owner, idiomatic reviewers will do their best effort to either review this
+# aspect themselves or find someone who can.
 
-bigquery/   @jsimonweb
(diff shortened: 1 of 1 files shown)


CODEOWNERS
```diff
@@ -1,20 +1,34 @@
-*           @amanda-tarafa
+# There are two aspects that should be taken into account when reviewing samples on this repo.
+#   * That the samples are meaningful and consistent with similar samples in other languages
+#     and with what is stated in documentation. Some sample sets have specific owners that
+#     are in charge of this aspect.
+#   * The idiomacy of the sample and the consistency with the rest of the samples on this repo.
```
jskeet, line 5: I think this should be "idiomaticity" - but it could be clearer as "the extent to which the sample is idiomatic C# and consistent with the rest of the samples in this repo"

CODEOWNERS
```diff
@@ -1,20 +1,34 @@
-*           @amanda-tarafa
+# There are two aspects that should be taken into account when reviewing samples on this repo.
+#   * That the samples are meaningful and consistent with similar samples in other languages
+#     and with what is stated in documentation. Some sample sets have specific owners that
+#     are in charge of this aspect.
+#   * The idiomacy of the sample and the consistency with the rest of the samples on this repo.
+#     Currently @amanda-tarafa, @jskeet, in that order of availability, are in charge of this aspect.
+# Any PR will requiere two reviewers, one for each of the aspects mentioned above. For those sample sets
```
jskeet, line 7:
```suggestion
# Any PR will require two reviewers, one for each of the aspects mentioned above. For those sample sets

```

CODEOWNERS
amanda-tarafa: Yes, will change.

CODEOWNERS
amanda-tarafa: Ahhhhhh

PR #1063, opened 2020-06-03
fix: update DELETE samples to match docs

skuruppu:
[Docs](https://cloud.google.com/spanner/docs/modify-mutation-api#deleting_rows_in_a_table) list a set of dot points on the multiple ways to perform deletes. The updated samples now contain these examples.

- Delete all the rows in a table.
- Delete a single row by specifying the key column values for the row.
- Delete a group of rows by creating a key range.
- Delete rows in an interleaved table by deleting the parent rows, if the interleaved table includes ON DELETE CASCADE in its schema definition.


spanner/api/Spanner/Program.cs
```diff
@@ -1769,23 +1769,44 @@ await connection.BeginTransactionAsync())
                 new Singer {SingerId = 5, FirstName = "David",
                     LastName = "Lomond"},
             };
+            List<Album> albums = new List<Album> {
+                new Album {SingerId = secondSingerId, AlbumId = 1,
```
simonz130, line 1773: nit: a space between { and first property - here and everywhere

spanner/api/Spanner/Program.cs
```diff
@@ -1769,23 +1769,44 @@ await connection.BeginTransactionAsync())
...
+            };
             // Create connection to Cloud Spanner.
             using (var connection = new SpannerConnection(connectionString))
             {
                 await connection.OpenAsync();
 
-                // Delete rows from the Singers table.
-                var cmd = connection.CreateDeleteCommand("Singers",
+                // Delete individual rows from the Albums table.
+                var cmd = connection.CreateDeleteCommand("Albums",
```
simonz130, line 1784: "Albums" (first parameter) should start on a new line

spanner/api/SpannerTest/Tests.cs
```diff
@@ -404,6 +404,19 @@ void TestStructParameters()
...
 
+        [Fact]
+        void TestDeleteSampleData()
+        {
+            ConsoleOutput output = _spannerCmd.Run("deleteSampleData",
+                _fixture.ProjectId, _fixture.InstanceId, _fixture.DatabaseId);
+            Assert.Contains("Deleted individual rows in Albums.", output.Stdout);
+            Assert.Contains($"2 row(s) deleted from Singers.", output.Stdout);
+            Assert.Contains($"7 row(s) deleted from Singers.", output.Stdout);
+            // Put the deleted sample data back.
```
simonz130, line 415: nit: add a space line between lines 414 and 415.

skuruppu:
Thanks for the review @simonz130. I made the requested changes.

spanner/api/Spanner/Program.cs
```diff
@@ -1757,35 +1757,44 @@ await connection.BeginTransactionAsync())
...
+                   "DELETE FROM Singers WHERE SingerId >= 3 AND SingerId < 5");
+                int rowCount = await cmd.ExecuteNonQueryAsync();
+                Console.WriteLine($"{rowCount} row(s) deleted from Singers.");
+
+                // Delete remaining Singers rows, which will also delete the remaining Albums rows
+                // since it was defined with ON DELETE CASCADE.
+                cmd = connection.CreateDmlCommand(
+                   "DELETE FROM Singers WHERE true");
+                rowCount = await cmd.ExecuteNonQueryAsync();
+                Console.WriteLine($"{rowCount} row(s) deleted from Singers.");
```
skuruppu, line 1797: In some of the other languages, there is a way to call delete with a KetSet and a KeyRange. E.g. in Python the same sample looks like this:

```
    # Delete individual rows
    albums_to_delete = spanner.KeySet(
        keys=[[2, 1], [2, 3]])

    # Delete a range of rows where the column key is >=3 and <5
    singers_range = spanner.KeyRange(start_closed=[3], end_open=[5])
    singers_to_delete = spanner.KeySet(
        ranges=[singers_range])

    # Delete remaining Singers rows, which will also delete the remaining 
    # Albums rows since it was defined with ON DELETE CASCADE
    remaining_singers = spanner.KeySet(all_=True)

    with database.batch() as batch:
        batch.delete('Albums', albums_to_delete)
        batch.delete('Singers', singers_to_delete)
        batch.delete('Singers', remaining_singers)
```

I searched through the code to see if I can do something similar here but I don't think we can. `DbParameterCollection` doesn't have a concept like that so I don't think we can construct a delete command that will be equivalent to what Python is doing. This is why my implementation here resorts to a hand-written query.

But let me know if there's something I missed.

spanner/api/Spanner/Program.cs
```diff
@@ -1757,35 +1757,44 @@ await connection.BeginTransactionAsync())
...
-                new Singer {SingerId = 3, FirstName = "Alice",
-                    LastName = "Trentor"},
-                new Singer {SingerId = 4, FirstName = "Lea",
-                    LastName = "Martin"},
-                new Singer {SingerId = 5, FirstName = "David",
-                    LastName = "Lomond"},
+            List<Album> albums = new List<Album> {
+                new Album { SingerId = secondSingerId, AlbumId = 1, AlbumTitle = "Green" },
+                new Album { SingerId = secondSingerId, AlbumId = 3, AlbumTitle = "Terrified" },
             };
```
amanda-tarafa, line 1763:
```suggestion
            List<Album> albums = new List<Album> 
            {
                new Album { SingerId = secondSingerId, AlbumId = 1, AlbumTitle = "Green" },
                new Album { SingerId = secondSingerId, AlbumId = 3, AlbumTitle = "Terrified" },
            };
```

spanner/api/Spanner/Program.cs
```diff
@@ -1757,35 +1757,44 @@ await connection.BeginTransactionAsync())
...
             using (var connection = new SpannerConnection(connectionString))
             {
                 await connection.OpenAsync();
 
-                // Delete rows from the Singers table.
-                var cmd = connection.CreateDeleteCommand("Singers",
+                // Delete individual rows from the Albums table.
+                var cmd = connection.CreateDeleteCommand(
+                    "Albums",
                     new SpannerParameterCollection {
```
amanda-tarafa, line 1772:
```suggestion
                    new SpannerParameterCollection 
                    {
```

spanner/api/Spanner/Program.cs
```diff
@@ -1757,35 +1757,44 @@ await connection.BeginTransactionAsync())
...
-                        {"SingerId", SpannerDbType.Int64}
-                    });
-                await Task.WhenAll(singers.Select(singer =>
+                        {"SingerId", SpannerDbType.Int64},
```
amanda-tarafa, line 1773: Spaces before and after {}, the linter will fail anyway. Here and everywhere.
```suggestion
                        { "SingerId", SpannerDbType.Int64 },
```

spanner/api/Spanner/Program.cs
```diff
@@ -1757,35 +1757,44 @@ await connection.BeginTransactionAsync())
...
+                        {"AlbumId", SpannerDbType.Int64}
+                    }
+                );
+                await Task.WhenAll(albums.Select(album =>
                 {
-                    cmd.Parameters["SingerId"].Value = singer.SingerId;
+                    cmd.Parameters["SingerId"].Value = album.SingerId;
```
amanda-tarafa, line 1779: This won't work when executing several commands in parallel as you are doing here. The Parameters collection will be the same for all executions so you are overriding the parameters values every time. You'll end up deleting only a few of the rows, those for which the values were copied to the request before the "next value change" happened.
You should either execute the commands sequentially or create a new command for each delete, i.e. add the command creation here.

spanner/api/Spanner/Program.cs
amanda-tarafa: Nope, we don't have anything like that, and we are sort of constraint by ADO.NET. It's possible we could add methods similar to those but it's not a usual pattern for ADO.NET providers as far as I know.

spanner/api/Spanner/Program.cs
```diff
@@ -1757,35 +1757,44 @@ await connection.BeginTransactionAsync())
...
+                // Delete a range of rows from the Singers table where the column key is >=3 and <5.
+                cmd = connection.CreateDmlCommand(
+                   "DELETE FROM Singers WHERE SingerId >= 3 AND SingerId < 5");
+                int rowCount = await cmd.ExecuteNonQueryAsync();
+                Console.WriteLine($"{rowCount} row(s) deleted from Singers.");
+
+                // Delete remaining Singers rows, which will also delete the remaining Albums rows
+                // since it was defined with ON DELETE CASCADE.
+                cmd = connection.CreateDmlCommand(
+                   "DELETE FROM Singers WHERE true");
```
amanda-tarafa, line 1795: Consider removing the "WHERE true", I don't think it's needed.

spanner/api/Spanner/Program.cs
```diff
@@ -1757,35 +1757,44 @@ await connection.BeginTransactionAsync())
...
+                rowCount = await cmd.ExecuteNonQueryAsync();
+                Console.WriteLine($"{rowCount} row(s) deleted from Singers.");
             }
         }
         // [END spanner_delete_data]
```
amanda-tarafa, line 1800: This sample now does a lot of things. Would it be possible to split it in several methods, even if they are all enclosed by the same region tag. I'm thinking on testability, specially when we move away from command line runners, we would like to test against something other than the standard output, etc.

spanner/api/SpannerTest/Tests.cs
```diff
@@ -432,6 +432,20 @@ void TestStructParameters()
...
+        void TestDeleteSampleData()
+        {
+            ConsoleOutput output = _spannerCmd.Run("deleteSampleData",
+                _fixture.ProjectId, _fixture.InstanceId, _fixture.DatabaseId);
+            Assert.Contains("Deleted individual rows in Albums.", output.Stdout);
+            Assert.Contains($"2 row(s) deleted from Singers.", output.Stdout);
+            Assert.Contains($"7 row(s) deleted from Singers.", output.Stdout);
+
+            // Put the deleted sample data back.
+            _spannerCmd.Run("insertSampleData",
```
amanda-tarafa, line 445: What if this fails, it will spoil all other tests possibly. As a setup for this test, can we add data (or better yet, specific tables) for the delete operation to use freely?

spanner/api/Spanner/Program.cs
skuruppu: The WHERE clause is unfortunately required for DELETEs in Cloud Spanner. This is to prevent accidentally deleting all data when you meant to delete only a few rows.

https://cloud.google.com/spanner/docs/dml-syntax#where_clause

spanner/api/Spanner/Program.cs
skuruppu: Thanks, that's what I suspected. It's ok, I don't think we need to add the methods. If a user requests it, we can reconsider.

spanner/api/Spanner/Program.cs
skuruppu: Oh interesting, I didn't realize and for some reason, it didn't come up with the tests. Thanks for spotting this.

spanner/api/Spanner/Program.cs
skuruppu: I separated it out into their own functions. I would like to keep them within a single region tag when we move it out to separate files since that would make it easier to pull the samples into the docs. Hope that's ok.

spanner/api/SpannerTest/Tests.cs
skuruppu: That's fair enough. I set it up such that the the function that executes "deleteSampleData" sets up the tables and data, then drops the tables afterwards. I did it this way because I just didn't want to introduce more command line functions.

Let me know if this is ok or you'd prefer for me to setup the test data in the test itself. We can also cleanup all this during the refactoring.

spanner/api/Spanner/Program.cs
```diff
@@ -1749,14 +1749,88 @@ await connection.BeginTransactionAsync())
...
         // [START spanner_delete_data]
-        public static async Task DeleteSampleDataAsync(
+        public static async Task DeleteIndividualRowsAsync(
+            string projectId, string instanceId, string databaseId)
+        {
+            const int singerId = 2;
+            string connectionString =
+                $"Data Source=projects/{projectId}/instances/{instanceId}"
+                + $"/databases/{databaseId}";
+            List<Album> albums = new List<Album> {
```
amanda-tarafa, line 1759: nit: Bracket on next line unless collection initialization is all in one line. Check here and everywhere.
```suggestion
            List<Album> albums = new List<Album>
            {
```

spanner/api/Spanner/Program.cs
```diff
@@ -1749,14 +1749,88 @@ await connection.BeginTransactionAsync())
...
+            using (var connection = new SpannerConnection(connectionString))
+            {
+                await connection.OpenAsync();
+
+                // Delete individual rows from the UpcomingAlbums table.
+                await Task.WhenAll(albums.Select(album =>
+                {
+                    var cmd = connection.CreateDeleteCommand(
+                        "UpcomingAlbums",
+                        new SpannerParameterCollection {
```
amanda-tarafa, line 1773: nit: bracket.

spanner/api/Spanner/Program.cs
amanda-tarafa: Yes, that's totally fine, there's nothing wrong with having multiple methods, or even type definitions, in the same region tag. A region tag though should be fully defined in a single file, else it's really hard for maintainers to a) follow what a sample (the code that's actually shown) does and b) test what a sampel does.
It's also fine to have several tags show exactly the same code. The case might be that language 1 does things different for sample a and sample b, but for language 2, samples a and b are exactly the same. You would have two region tags, in language 1 you have to separate samples for each of those tags, in language 2 you have the same sample enclosed by both tags.

spanner/api/Spanner/Program.cs
amanda-tarafa: Unresolved this because this still needs fixing.

spanner/api/Spanner/Program.cs
```diff
@@ -1769,29 +1843,91 @@ await connection.BeginTransactionAsync())
                 new Singer {SingerId = 5, FirstName = "David",
                     LastName = "Lomond"},
             };
+            List<Album> albums = new List<Album> {
```
amanda-tarafa, line 1846: nit: bracket to next line.

spanner/api/Spanner/Program.cs
```diff
@@ -1769,29 +1843,91 @@ await connection.BeginTransactionAsync())
...
             // Create connection to Cloud Spanner.
             using (var connection = new SpannerConnection(connectionString))
             {
                 await connection.OpenAsync();
 
-                // Delete rows from the Singers table.
-                var cmd = connection.CreateDeleteCommand("Singers",
+                // Insert rows into the Singers table.
+                var cmd = connection.CreateInsertCommand("Singers",
                     new SpannerParameterCollection {
```
amanda-tarafa, line 1865: nit: bracket.

spanner/api/Spanner/Program.cs
```diff
@@ -1769,29 +1843,91 @@ await connection.BeginTransactionAsync())
...
                 {
                     cmd.Parameters["SingerId"].Value = singer.SingerId;
+                    cmd.Parameters["FirstName"].Value = singer.FirstName;
+                    cmd.Parameters["LastName"].Value = singer.LastName;
+                    return cmd.ExecuteNonQueryAsync();
+                }));
+
+                // Insert rows into the Albums table.
+                cmd = connection.CreateInsertCommand("Albums",
+                    new SpannerParameterCollection {
```
amanda-tarafa, line 1880: nit: bracket

spanner/api/Spanner/Program.cs
```diff
@@ -1860,7 +1996,31 @@ await connection.BeginTransactionAsync())
...
+                        await connection.CreateDdlCommand($"DROP TABLE {table}")
+                            .ExecuteNonQueryAsync();
+                    }
+                    catch (SpannerException e) when (e.ErrorCode == ErrorCode.NotFound)
+                    {
+                        // Table does not exist.  Not a problem.
+                    }
+                }
+                return 0;
+            }
```
amanda-tarafa, line 2022: the foreach seems a little bit overkill :)
```suggestion
            using (var connection = new SpannerConnection(connectionString))
            {
                DeleteTable("UpcomingAlbums");
                DeleteTable("UpcomingSingers");
                return 0;
            }
            void DeleteTable(string table)
            {
                try
                    {
                        await connection.CreateDdlCommand($"DROP TABLE {table}")
                            .ExecuteNonQueryAsync();
                    }
                    catch (SpannerException e) when (e.ErrorCode == ErrorCode.NotFound)
                    {
                        // Table does not exist.  Not a problem.
                    }
            }
```

spanner/api/Spanner/Program.cs
amanda-tarafa: Since you are opening the connection yourself the first await actually happens when you send the request, so this very particular example was not a problem. But, if you weren't opening the connection then the first await would happen before sending the request, to open the connection, or if you were doing this inside threads (we had an issue exactly because of that the other day), then it will be a problem. It's best to always show the prefered approach, which is to have individual parameter collections or commands, instead of explaining all this nuances I think.

spanner/api/Spanner/Program.cs
amanda-tarafa: Ah makes sense.

spanner/api/SpannerTest/Tests.cs
amanda-tarafa: This is fine for now. On rewrite I'd like to have proper test setup and teardown.

skuruppu:
Thanks for the review @amanda-tarafa. I addressed all the remaining feedback and the presubmits are now running. Happy for you force-merge if nothing unusual fails. But otherwise, I will take a look at any failing tests next week.

Have a great weekend!

amanda-tarafa:
Merging. One other Spanner test failed but we have #1065 for that.

Issue #924, opened 2019-12-18
Reactivate Sudukumb WebApp tests after fixing Chrome Driver issue.


amanda-tarafa:
We are getting the following error when trying to run Sudokumb WebApp tests which use Selenium.

> This version of chromedriver only supports chrome version 77

I've deactivated these tests (#922) until the issue is fixed. I want to check all other possible failures and get CI as green as possible to be able to merge several PRs we have pending. 

@SurferJeffAtGoogle any pointers as to what this error may be are welcomed.

PR #1069, opened 2020-06-04
fix: remove sudokumb sample

SurferJeffAtGoogle:
fixes https://github.com/GoogleCloudPlatform/dotnet-docs-samples/issues/924

I used it for one presentation, but never had a chance to write blog posts about it.  I'll resurrect it if I have time to write those posts.  In the meantime, I don't won't its maintenance burdening anyone else.


kms/api/CloudKmsAspNetSample/appsecrets.json
```diff
@@ -0,0 +1,5 @@
+{
```
sethvargo, line 1: Is this change relevant?

SurferJeffAtGoogle: Nope, I'll roll it back.

amanda-tarafa:
Force mergin, the errors in CI are unrelated.

Issue #1027, opened 2020-04-20
Update F# AppEngine Flex samples to .NET Core 3.1 and reactivate tests.


amanda-tarafa:
We have updated the Docker image to support .NET Core 3.1. In the process we've had to remove support for .NET Core 1.1 which is not supported by MS anymore and can't be properly pulled into the Docker image.
This work was done as part of #1013

amanda-tarafa:
Skipping building and testing of F# sample in #1026

PR #1070, opened 2020-06-04
fix: update f# sample to .net core 3.1

SurferJeffAtGoogle:
and re-enable its test.

Fixes https://github.com/GoogleCloudPlatform/dotnet-docs-samples/issues/1027


appengine/flexible/F#/F#.fsproj
```diff
@@ -11,9 +11,6 @@
   </ItemGroup>
   <ItemGroup>
     <PackageReference Include="Google.Cloud.Vision.V1" Version="1.0.0-beta05" />
```
amanda-tarafa, line 13: Can you update this to 2.0.0? Depending on what Vision is being used for, a couple of changes might be needed. Thanks.
https://www.nuget.org/packages/Google.Cloud.Vision.V1/

SurferJeffAtGoogle: Done.

amanda-tarafa:
Force merging, the errors are because we have reached IAM quota.

Issue #1065, opened 2020-06-03
[Spanner] Deactivate Backup tests


skuruppu:
The test times out so we're skipping this for now ([example](https://source.cloud.google.com/results/invocations/f255d775-e3ca-4231-b34d-8b70b669a0d7/targets/github%2Fdotnet-docs-samples%2Fspanner%2Fapi%2FSpannerTest/tests)). We should look into this and re-enable when we figure it out.

skuruppu:
Test will be skipped in #1047.

amanda-tarafa:
Thanks! Assigned to you but mostly so you can reassign. When I'm done with all the rest, if this is still around I will grab it.

laljikanjareeya:
And `TestQueryWithBytes` is also failing . I've skipped it also.

skuruppu:
`GoogleCloudSamples.Spanner.SpannerTests.TestUpdateBackup` timed out in this [run](https://source.cloud.google.com/results/invocations/ca33fcee-ddc2-48cd-88a5-a6859f3ef242/targets/github%2Fdotnet-docs-samples%2Fspanner%2Fapi%2FSpannerTest/tests). I think we're going to have to skip pretty much all the backups tests. I'm hoping to look into this tomorrow.

amanda-tarafa:
Yep, on nightly CI running, Spanner is still timing out from time to time. I'm fine with deactivating the problematic tests. On rewrite we can address whatever is making them fail I think.

amanda-tarafa:
@skuruppu , @laljikanjareeya Can't remember if this issue is still relevant. Ain't the backups being done just once now after the rewrite of some of the samples? It's fine if there's still some looese ends related to this, we can keep open until everything is addressed.

laljikanjareeya:
> @skuruppu , @laljikanjareeya Can't remember if this issue is still relevant. Ain't the backups being done just once now after the rewrite of some of the samples? It's fine if there's still some looese ends related to this, we can keep open until everything is addressed.

We are not deleting backup every time test runs, This will not be an issue but still backup sample is pending to add(will add in part 4) so let keep this open for while.

PR #1072, opened 2020-06-05
fix: add a code sample for CreateInstance

skuruppu:


spanner/api/SpannerTest/Tests.cs
```diff
@@ -217,22 +217,23 @@ async void InitializeInstance()
...
+                    {
+                        _spannerCmd.Run("createInstance", _fixture.ProjectId, _fixture.InstanceId);
+                        rethrow = false;
+                        break;
+                    }
+                }
+                if (rethrow)
+                {
+                    throw;
+                }
```
skuruppu, line 236: @laljikanjareeya catching this exception doesn't work unfortunately. I'm not sure I understand why.

The error printed out looks like this so I'm a bit confused by why it's not caught correctly.

```
  Error Message:
   System.AggregateException : One or more errors occurred. (Object not found.)
---- Google.Cloud.Spanner.Data.SpannerException : Object not found.
-------- Grpc.Core.RpcException : Status(StatusCode=NotFound, Detail="Instance not found: projects/appdev-soda-spanner-staging/instances/test-create-instance")
  Stack Trace:
     at System.Threading.Tasks.Task.Wait(Int32 millisecondsTimeout, CancellationToken cancellationToken)
   at System.Threading.Tasks.Task.Wait()
   at GoogleCloudSamples.Spanner.Program.CreateSampleDatabase(String projectId, String instanceId, String databaseId) in /google-cloud-dotnet/spanner/api/Spanner/Program.cs:line 1929
   at GoogleCloudSamples.Spanner.Program.<>c.<Main>b__124_1(CreateSampleDatabaseOptions opts) in /google-cloud-dotnet/spanner/api/Spanner/Program.cs:line 3689
```

laljikanjareeya: @skuruppu `instanceAdminClient.GetInstanceAsync` will throw `RpcException` not `AggregateException`  because we are not using command line for getting instance, so below code should work fine.

```csharp
async void InitializeInstance()
{
    InstanceAdminClient instanceAdminClient = await InstanceAdminClient.CreateAsync();
    string name = $"projects/{_fixture.ProjectId}/instances/{_fixture.InstanceId}";
    try
    {
        Instance response = await instanceAdminClient.GetInstanceAsync(name);
    }
    catch (RpcException ex) when (ex.Status.StatusCode == StatusCode.NotFound)
    {
        _spannerCmd.Run("createInstance", _fixture.ProjectId, _fixture.InstanceId);
    }
}
```

spanner/api/Spanner/CreateInstance.cs
```diff
@@ -0,0 +1,73 @@
...
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+// [START spanner_create_instance]
+
+using Google.Api.Gax.ResourceNames;
+using Google.Cloud.Spanner.Admin.Instance.V1;
+using Google.Cloud.Spanner.Common.V1;
+using Google.LongRunning;
+using Google.Protobuf.WellKnownTypes;
```
laljikanjareeya, line 21: please remove this unnecessary namespace.

spanner/api/SpannerTest/Tests.cs
amanda-tarafa: A couple of things happening here:

1. If you look closely, the error you are getting is in `CreateSampleDatabase` and not in `InitializeInstance`. What's happening is exactly what I was describing a couple of comments above. `InitializeInstance` is not being awaited so there's no guarantee that when `CreateSampleDatabase` executes the instance has actually finished creation. And that's why you are getting this error.
2. As for when you get the actual exception or an AggregateException, take a look at [Exception Handling (Task Parallel library)](https://docs.microsoft.com/en-us/dotnet/standard/parallel-programming/exception-handling-task-parallel-library). Notice that because of the command runner we are calling Wait() on async methods (which is different from awaiting) and that's why we end up getting the AggregateException.

laljikanjareeya:
@googlebot I consent

laljikanjareeya:
@googlebot I fixed it

skuruppu:
@googlebot I fixed it.

spanner/api/Spanner/CreateInstance.cs
```diff
@@ -0,0 +1,73 @@
...
+        {
+            // Create the InstanceAdminClient instance.
+            InstanceAdminClient instanceAdminClient = InstanceAdminClient.Create();
+
+            // Initialize request parameters.
+            Instance instance = new Instance
+            {
+                InstanceName = InstanceName.FromProjectInstance(projectId, instanceId),
+                ConfigAsInstanceConfigName =
+                    InstanceConfigName.FromProjectInstanceConfig(projectId, "regional-us-central1"),
```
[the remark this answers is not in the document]
skuruppu: Originally I had this but for the other languages it was unconventional to accept arguments to the command line tool that wasn't just the project, instance, database IDs. So I removed it. (Most other languages still use the command line tool)

spanner/api/Spanner/CreateInstance.cs
```diff
@@ -0,0 +1,73 @@
...
+using Google.Cloud.Spanner.Common.V1;
+using Google.LongRunning;
+using Google.Protobuf.WellKnownTypes;
+using System;
+
+namespace GoogleCloudSamples.Spanner
+{
+    public class CreateInstance
+    {
+        public static object SpannerCreateInstance(string projectId, string instanceId)
```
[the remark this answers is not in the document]
skuruppu: For reasons I don't quite understand, the [generated code for Spanner](https://github.com/googleapis/google-cloud-dotnet/search?q=throwonanyerror+spanner&unscoped_q=throwonanyerror+spanner) doesn't have ThrowOnAnyError(). It's only available on some APIs like [BigQuery](https://github.com/googleapis/google-cloud-dotnet/search?p=2&q=throwonanyerror&unscoped_q=throwonanyerror).

spanner/api/SpannerTest/Tests.cs
skuruppu: Oh wow, thank you both for explaining what's going on here. Now I understand.

1. I've now made made `InitializeInstance` synchronous. This was because we couldn't call an async method within a lock.

2. Now that I realize where the exception was coming from, it makes sense why I was seeing an AggregateException instead of an RpcException. It was because the exception was coming from `CreateSampleDatabase` as @amanda-tarafa explained. Now that that issue is fixed, it throws and `RpcException` as expected here.

spanner/api/SpannerTest/Tests.cs
```diff
@@ -217,22 +217,23 @@ async void InitializeInstance()
                 string name = $"projects/{_fixture.ProjectId}/instances/{_fixture.InstanceId}";
```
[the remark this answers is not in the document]
skuruppu: I couldn't make this async unfortunately because the compiler complains that we're calling await inside a lock (line 136). But I've now made it synchronous and everything works.

spanner/api/Spanner/Program.cs
```diff
@@ -37,6 +37,15 @@ class DefaultOptions
         public string databaseId { get; set; }
     }
 
+    [Verb("createInstance", HelpText = "Create a Cloud Spanner instance in your project.")]
```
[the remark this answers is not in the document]
skuruppu: I had no idea that I could get the tests running without it but turns out I can so I've removed this code :) Thanks for the suggestion.

skuruppu:
PTAL

Thanks for the review feedback. It's much cleaner now :)

skuruppu:
Hmmm `UpdateBackup` timed out in the test. I will take a look at these tests hopefully tomorrow as a part of https://github.com/GoogleCloudPlatform/dotnet-docs-samples/issues/1065.

spanner/api/Spanner/CreateInstance.cs
```diff
@@ -0,0 +1,66 @@
...
+    {
+        // Create the InstanceAdminClient instance.
+        InstanceAdminClient instanceAdminClient = InstanceAdminClient.Create();
+
+        // Initialize request parameters.
+        Instance instance = new Instance
+        {
+            InstanceName = InstanceName.FromProjectInstance(projectId, instanceId),
+            ConfigAsInstanceConfigName =
+                InstanceConfigName.FromProjectInstanceConfig(projectId, "regional-us-central1"),
```
amanda-tarafa, line 35: Just a comment: When we do the rewrite we would ideally remove all the command line runner bits. At that point I belive we should move away from attempting command line related consistency with samples in other languages. Are you OK with that?

spanner/api/Spanner/CreateInstance.cs
```diff
@@ -0,0 +1,66 @@
...
+
+        // Make the CreateInstance request.
+        Operation<Instance, CreateInstanceMetadata> response =
+            instanceAdminClient.CreateInstance(parentAsProjectName, instanceId, instance);
+
+        Console.WriteLine("Waiting for the operation to finish.");
+
+        // Poll until the returned long-running operation is complete.
+        Operation<Instance, CreateInstanceMetadata> completedResponse =
+            response.PollUntilCompleted();
```
amanda-tarafa, line 53: My bad with ThrowOnAnyErro(). That's something we write for BigQuery, manually. For grpc long running operations the correct way of checking errors is how it's being done in this sample.

spanner/api/Spanner/CreateInstance.cs
skuruppu: Yeah I think that should be fine. It's just from the perspective of the docs team, they try to write one description of what the sample is doing and ideally the description would apply consistently to all 8 clients. But some small minor variations should be fine as long as main thing the sample is trying to show is consistent.

skuruppu:
@amanda-tarafa if you can force merge, that would be great. I will skip some more tests in a separate PR.

PR #1074, opened 2020-06-08
feat: adds create BigQuery job sample

telpirion:
This sample adds the following region tag:

* bigquery_create_job

The Python canonical sample is here:
https://github.com/googleapis/python-bigquery/blob/master/samples/create_job.py

### DIFF
--- a/bigquery/api/Snippets/CreateJob.cs
+++ b/bigquery/api/Snippets/CreateJob.cs
@@ -20,7 +20,7 @@ using System.Collections.Generic;
 
 public class BigQueryCreateJob
 {
-    public void CreateJob(string projectId = "your-project-id")
+    public BigQueryJob CreateJob(string projectId = "your-project-id")
     {
         string query = @"
             SELECT country_name from `bigquery-public-data.utility_us.country_code_iso";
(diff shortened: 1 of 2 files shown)


telpirion:
It looks like some unrelated tests are failing (IAM) :(.

bigquery/api/Snippets/CreateJob.cs
```diff
@@ -0,0 +1,50 @@
...
+            JobLocation = "us",
+            JobIdPrefix = "code_sample_",
+            Labels = new Dictionary<string, string>
+            {
+                ["example-label"] = "example-value"
+            },
+            MaximumBytesBilled = 1000000
+        };
+
+        BigQueryJob queryJob = client.CreateQueryJob(
```
amanda-tarafa, line 42: No that we don't go through the command line, we can return the queryJob and use that to tests against.

bigquery/api/Snippets/Test.cs
```diff
@@ -85,6 +85,15 @@ public void TestCreateDataset()
         dataset.Delete();
     }
 
+    [Fact]
+    public void TestCreateJob()
+    {
+        var snippet = new BigQueryCreateJob();
+        snippet.CreateJob(_projectId);
+        string output = _stringOut.ToString();
```
amanda-tarafa, line 93: And instead of depending on the output, we check the return of the CreateJob method, for instance, that is no null.

telpirion:
Thank you!

PR #1075, opened 2020-06-08
fix: renables and fixes DetectIntentTextsTest

### LINKED ISSUE #935: Dialogflow.DetectIntentTextsTest not passing
The test defined in [DetectIntentTextsTest.TestDetectIntentFromTexts](https://github.com/GoogleCloudPlatform/dotnet-docs-samples/blob/master/dialogflow/api/Test/DetectIntentTextsTest.cs is not passing, it seems to be very strict. It's OK to loosen it up, we just want to test that the sample makese sense.
It should be checked by someone who is familiar with the Dialogflow API and can write a good enough test.

telpirion:
Changes DetectIntentTextsTest so that it checks for snippet string literals rather than API output.

**Note:** This PR partially re-enables tests for the Dialogflow project (see #947). It filters the tests to select only the DetectIntentTextsTest.  I will adjust the filter as I re-enable more of the DF tests.

Fixes #935

### DIFF
--- a/dialogflow/api/Test/DetectIntentTextsTest.cs
+++ b/dialogflow/api/Test/DetectIntentTextsTest.cs
@@ -28,7 +28,6 @@ namespace GoogleCloudSamples
             Assert.Equal(0, ExitCode);
 
             Assert.Contains("Query text:", Stdout);
-            Assert.Contains("Intent detected:", Stdout);
             Assert.Contains("Intent confidence:", Stdout);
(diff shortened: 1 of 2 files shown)


telpirion:
It looks like some unrelated tests are failing (IAM) :(.

amanda-tarafa:
I will force merge. IAM failures are because we have blown quota.

telpirion:
Yay! Thank you.

PR #1076, opened 2020-06-09
Convert Metadata to new style

pheonixblade9:


dlp/api/Snippets/DeidentifyWithFpe.cs
```diff
@@ -0,0 +1,84 @@
...
+
+public class DeidentifyWithFpe
+{
+    public static DeidentifyContentResponse Deidentify(
+        string projectId,
+        string dataValue,
+        IEnumerable<InfoType> infoTypes,
+        string keyName,
+        string wrappedKey,
+        string alphabet)
```
amanda-tarafa, line 32: This shouldn't be a string, it should be FfxCommonNativeAlphabet, and then there's no need to parse the value. This is leftover from the command line runner. Change here and everywhere that proper types are used instead of string

dlp/api/Snippets/DeidentifyWithDateShift.cs
```diff
@@ -0,0 +1,141 @@
...
+        int lowerBoundDays,
+        int upperBoundDays,
+        string dateFields,
+        string contextField = "",
+        string keyName = "",
+        string wrappedKey = "")
+    {
+        var dlp = DlpServiceClient.Create();
+
+        // Read file
```
amanda-tarafa, line 37: Is it an explicit objective of this sample to show how to read and parse values from a CSV? Could we receive a list of dates and values instead so the users are not lost in all the CSV related code and can focus on DLP? I believe this to be left over from command line running as well. At least consider moving the CSV processing code to a separate method.

dlp/api/Snippets/DeidentifyWithDateShift.cs
```diff
@@ -0,0 +1,141 @@
...
+        var protoRows = csvRows.Select(CsvRow =>
+        {
+            var rowValues = CsvRow.Split(',');
+            var protoValues = rowValues.Select(RowValue =>
+            {
+                if (System.DateTime.TryParse(RowValue, out var parsedDate))
+                {
+                    return new Value
+                    {
+                        DateValue = new Google.Type.Date
```
amanda-tarafa, line 53: Google.Type.Date.FromDateTime

dlp/api/Snippets/DeidentifyWithDateShift.cs
```diff
@@ -0,0 +1,141 @@
...
+        // Construct + execute the request
+        var dateShiftConfig = new DateShiftConfig
+        {
+            LowerBoundDays = lowerBoundDays,
+            UpperBoundDays = upperBoundDays
+        };
+        var hasKeyName = !string.IsNullOrEmpty(keyName);
+        var hasWrappedKey = !string.IsNullOrEmpty(wrappedKey);
+        var hasContext = !string.IsNullOrEmpty(contextField);
+        if (hasKeyName && hasWrappedKey && hasContext)
```
amanda-tarafa, line 88: For simplicity, can we assume that these will be present always?

dlp/api/Snippets/DeidentifyWithDateShift.cs
```diff
@@ -0,0 +1,141 @@
...
+                            Fields = { dateFieldList }
+                        }
+                    }
+            }
+        };
+
+        var response = dlp.DeidentifyContent(
+            new DeidentifyContentRequest
+            {
+                Parent = $"projects/{projectId}",
```
amanda-tarafa, line 126: ParentAsProjectName, if you can't do that there's a missing annotation in the proto.

dlp/api/Snippets/DeidentifyWithDateShift.cs
pheonixblade9: I'm pretty much copying everything over wholesale and intend to do a pass to make things better, as well. Right now I'm just focused on getting the samples ported over.

dlp/api/Snippets/DeidentifyWithDateShift.cs
pheonixblade9: I'm not sure what the implications of this being in an incorrect state are, so probably makes sense to leave it as-is for now.

PR #1083, opened 2020-06-11
Add risk analysis and trigger snippets and tests

pheonixblade9:


dlp/api/Snippets.Tests/TriggersCreateTests.cs
```diff
@@ -0,0 +1,57 @@
+using System.IO;
```
amanda-tarafa, line 1: These should go after the license with the rest. Include the change in one of the next PRs.

PR #1101, opened 2020-06-15
Create sample for detecting intent stream using mic.

ssilb4:
I created sample code for detecting intent stream using mic.


amanda-tarafa:
Actually, this just looks like a copy paste from a sample we already have: [DialogflowSamples/DetectIntentStream.cs](https://github.com/GoogleCloudPlatform/dotnet-docs-samples/blob/master/dialogflow/api/DialogflowSamples/DetectIntentStream.cs).
Will leave open for a couple of days in case I'm missing something, else, I'll close as this is increasingly looking like spam.

ssilb4:
First, Sry for my english.

Actually as you say, it is very similar to DialogflowSamples/DetectIntentStream.cs .
I just thought "streaming sample" is not to use audio file but to use mic.

And I saw that there are similar files in speech sample. In [Recognize.cs](https://github.com/GoogleCloudPlatform/dotnet-docs-samples/blob/master/speech/api/Recognize/Recognize.cs) , StreamingMicRecognizeAsync and StreamingRecognizeAsync. so I added this file.

And documents have differences in speech-to-text and dialogflow. In [speech-to-text](https://cloud.google.com/speech-to-text/docs/streaming-recognize), it has using mic sample, but in [dialogflow documents](https://cloud.google.com/dialogflow/docs/how/detect-intent-stream) , it has only streaming file.

As newbie like me, It is hard to only change reading audio file to streaming audio using mic.  and I dont understand why they are so different in documents. Anyway, it is my fault. I just make spam. sry.

amanda-tarafa:
@ssilb4 You don't have to apologize, what you were trying to do is the right thing, I see now the difference in the code you wrote and the one we have. I'm sorry for having jumped so fast to the conclusion that this might be spam, it is not.

I'll keep the PR closed for now, but will asked internally if we want to show a sample like the one you've written. If yes, and provided you do all I requested on my first comment, then we can reopen the PR for review and merge.

Thanks again, and don't feel that because you are a "newbie" you can't contribute, please do so as much as you can. Including a description in your PRs will help with avoiding this kind of missunderstanding in the future, and that's valid for any repo you might want to contribute to.

ssilb4:
Thanks.

PR #1102, opened 2020-06-15
Convert snippets to use global region endpoint

pheonixblade9:


sushicw:
This should fix internal bug b/158100365 for C# samples.

PR #1106, opened 2020-06-18
[asset] Add sample code for new RPCs

yuyifan-google:


asset/api/Asset.Samples.Tests/ExportAssetsTest.cs
```diff
@@ -0,0 +1,41 @@
...
+    }
+
+    [Fact]
+    public void TestExportAssets()
+    {
+        // Run the sample code.
+        var result = _sample.ExportAssets(_fixture.BucketName, _fixture.ProjectId);
+
+        string expectedOutput = String.Format("\"outputConfig\": {{ \"gcsDestination\": {{ \"uri\": \"gs://{0}/my-assets.txt\" }} }}", _fixture.BucketName);
+        Assert.Contains(expectedOutput, result.ToString());
```
amanda-tarafa, line 39: Can you check the actual field instead of converting this to string. It's really harder to maintaing this code if at some point we want to test for something else.
Here and everywhere where similar code might be present.

asset/api/Asset.Samples.Tests/BatchGetAssetsHistoryTest.cs
```diff
@@ -0,0 +1,41 @@
+/*
+ * Copyright (c) 2018 Google LLC.
```
amanda-tarafa, line 2: nit: 2020

asset/api/Asset.Samples.Tests/BatchGetAssetsHistoryTest.cs
```diff
@@ -0,0 +1,41 @@
...
+    }
+
+    [Fact]
+    public void TestBatchGetAssetsHistory()
+    {
+        // Run the sample code.
+        string assetName = String.Format("//storage.googleapis.com/{0}", _fixture.BucketName);
+        var result = _sample.BatchGetAssetsHistory(new string[] { assetName }, _fixture.ProjectId);
+
+        Assert.Contains(assetName, result.ToString());
```
amanda-tarafa, line 39: Here as well, check the actual field.

asset/api/Asset.Samples/SearchAllResources.cs
```diff
@@ -0,0 +1,49 @@
...
+        };
+        request.AssetTypes.AddRange(assetTypes ?? new string[] { });
+
+        // Call the API.
+        PagedEnumerable<SearchAllResourcesResponse, ResourceSearchResult> response = client.SearchAllResources(request);
+
+        // Return the first page.
+        IEnumerator<SearchAllResourcesResponse> enumerator = response.AsRawResponses().GetEnumerator();
+        enumerator.MoveNext();
+        return enumerator.Current;
```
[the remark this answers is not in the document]
yuyifan-google: Thanks for the info!

asset/api/Asset.Samples.Tests/AssetFixture.cs
```diff
@@ -0,0 +1,66 @@
...
+        };
+        _bigQueryClient.CreateDataset(datasetId: DatasetId, dataset);
+
+        // Wait 10 seconds to let resource creation events go to CAI.
+        Thread.Sleep(10000);
+    }
+
+    public void Dispose()
+    {
+        _bucketFixture.Dispose();
```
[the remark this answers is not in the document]
yuyifan-google: It should. https://github.com/GoogleCloudPlatform/dotnet-docs-samples/blob/master/testutil/TestBucket.cs#L24

asset/api/Asset.Samples.Tests/SearchAllResourcesTest.cs
```diff
@@ -0,0 +1,42 @@
...
+    {
+        _fixture = fixture;
+        _sample = new SearchAllResourcesSample();
+    }
+
+    [Fact]
+    public void TestSearchAllResources()
+    {
+        // Run the sample code.
+        string scope = String.Format("projects/{0}", _fixture.ProjectId);
```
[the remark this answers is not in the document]
yuyifan-google: For the test, all I have is a project id, so the scope can only be 'projects/PROJECT_ID'. But for our api, scope is not limited to projects. It can also be 'folders/...' or 'organizations/...'
I don't want to put some test-specific things into the sample code, and users can always check our api doc to see the values are acceptable.

asset/api/Asset.Samples.Tests/ExportAssetsTest.cs
yuyifan-google: These RPCs are owned by our sister team (we share the same public api). I'm not familiar with these RPCs and how to meaningfully check the results, so I'd rather not change their existing tests. I can file a bug for them to update the tests.

asset/api/Asset.Samples.Tests/SearchAllResourcesTest.cs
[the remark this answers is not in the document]
amanda-tarafa: Ok, that's fair.

asset/api/Asset.Samples.Tests/Asset.Samples.Tests.csproj
```diff
@@ -0,0 +1,21 @@
+<Project Sdk="Microsoft.NET.Sdk">
+
+  <PropertyGroup>
+    <TargetFramework>netstandard2.1</TargetFramework>
```
amanda-tarafa, line 4: OK, so netstandard* is a specification, and then you have concrete implementations of that which target different platforms etc.
Libraries should almost always target netstandard* so they can be used by all implementations, as where "running" projects, like test projects should target implementations. In samples, we are shifting to targetting netcoreapp3.1.
```suggestion
    <TargetFramework>netcoreapp3.1</TargetFramework>
```

asset/api/Asset.Samples/SearchAllResources.cs
```diff
@@ -0,0 +1,47 @@
...
+// [START asset_quickstart_search_all_resources]
+
+using Google.Api.Gax;
+using Google.Cloud.Asset.V1;
+using System.Collections.Generic;
+using System.Linq;
+
+public class SearchAllResourcesSample
+{
+    public SearchAllResourcesResponse SearchAllResources(string scope, string query = "", string[] assetTypes = null, int pageSize = 0, string orderBy = "")
```
amanda-tarafa, line 24: Sorry for not mentioning this before:
If the default values for parameters are meaningful, it's fine they are there, but in this case, they are really not, so better just remove them, here and everywhere.

asset/api/Asset.Samples/BatchGetAssetsHistory.cs
```diff
@@ -0,0 +1,48 @@
+// Copyright(c) 2018 Google LLC.
```
amanda-tarafa, line 1: nit: 2020
And check all new files

asset/api/Asset.Samples/ExportAssets.cs
```diff
@@ -0,0 +1,45 @@
+// Copyright(c) 2018 Google LLC.
```
amanda-tarafa, line 1: nit: year

asset/api/Asset.Samples/SearchAllResources.cs
```diff
@@ -0,0 +1,47 @@
...
+
+        // Build the request.
+        SearchAllResourcesRequest request = new SearchAllResourcesRequest
+        {
+            Scope = scope,
+            Query = query,
+            PageSize = pageSize,
+            OrderBy = orderBy,
+        };
+        request.AssetTypes.AddRange(assetTypes ?? new string[] { });
```
amanda-tarafa, line 37: And then you can probably remove the null validation here.

asset/api/Asset.Samples.Tests/ExportAssetsTest.cs
amanda-tarafa: OK, but it's as simple as this:
```suggestion
        Assert.Equals($"gs://{_fixture.BucketName}/my-assets.txt", result.OutputConfig.GcsDestination.Uri);
```
ToString is only generating a JSON like representation of the actual object.
I'm fine if you don't want to change this, but please do create an issue for the other team to do so. This is very fiddly code.

yuyifan-google:
Yes please. Thanks you!

amanda-tarafa:
On this last execution one of the Assets test failed on Windows. It passed on Linux and before so it's flaky. Can you check?

yuyifan-google:
Tests work fine locally.
I notice a 'RemoteException' in the test log. Could it be some kind of occasional internal error of the kokoro tests? It happened to me before for another language that the tests failed to build. Could you rerun the tests to see if that particular test still fail to build? Thanks.

amanda-tarafa:
I'll check this failure, it might be CI on windows.

amanda-tarafa:
I'm merging as is, the error we are seeing on Assets' tests is because of Power Sherll. I have a PR ready to submit after this one.

PR #1111, opened 2020-06-22
spanner: cleanup old backups and databases before starting new test cycle

### LINKED ISSUE #1110: spanner: cleanup old databases and backups before starting new test cycle
Here is the [Test Result](https://source.cloud.google.com/results/invocations/19dac484-be94-48e3-8f56-8bff0b5013d1/targets/github%2Fdotnet-docs-samples%2Fspanner%2Fapi%2FSpannerTest%2FTestResults/tests) instance reached  the maximum database limit (100).

//cc
@skuruppu

laljikanjareeya:
Fixes #1110


spanner/api/SpannerTest/Tests.cs
```diff
@@ -133,13 +133,81 @@ public SpannerTests(SpannerFixture fixture)
...
         }
 
+        void DeleteStaleBackupsAndDatabases()
+        {
+            DatabaseAdminClient databaseAdminClient = DatabaseAdminClient.Create();
+            var instanceName = InstanceName.FromProjectInstance(_fixture.ProjectId, _fixture.InstanceId);
+            var databases = databaseAdminClient.ListDatabases(instanceName).Where(c => c.DatabaseName.DatabaseId.StartsWith("my-db-")).ToList();
+            var databasesToDelete = new List<string>();
+            // delete all the databases created before 5 hrs.
+            var timestamp = DateTime.UtcNow.AddHours(-5).ToTimestamp().Seconds;
```
skuruppu, line 151: In Node.js we made this 4 hours.

```suggestion
            var timestamp = DateTime.UtcNow.AddHours(-4).ToTimestamp().Seconds;

```

spanner/api/SpannerTest/Tests.cs
```diff
@@ -133,13 +133,81 @@ public SpannerTests(SpannerFixture fixture)
...
+                var databaseId = database.DatabaseName.DatabaseId.Replace("my-db-", "");
+                if (long.TryParse(databaseId, out long dbCreationTime) && dbCreationTime <= timestamp)
+                {
+                    databasesToDelete.Add(database.DatabaseName.DatabaseId);
+                }
+            });
+
+            Console.Out.WriteLineAsync($"{databasesToDelete.Count} old databases found.");
+
+            //get backups
```
skuruppu, line 163:
```suggestion
            // Get backups.
```

spanner/api/SpannerTest/Tests.cs
```diff
@@ -133,13 +133,81 @@ public SpannerTests(SpannerFixture fixture)
...
+            ListBackupsRequest request = new ListBackupsRequest
+            {
+                ParentAsInstanceName = instanceName,
+                Filter = $"database:my-db-"
+            };
+            var backups = databaseAdminClient.ListBackups(request).ToList();
+
+            //backups for databases to delete.
```
skuruppu, line 171:
```suggestion
            // Backups that belong to the databases to be deleted.
```

spanner/api/SpannerTest/Tests.cs
```diff
@@ -133,13 +133,81 @@ public SpannerTests(SpannerFixture fixture)
...
+            var backupsToDelete = backups.Where(c => databasesToDelete.Contains(DatabaseName.Parse(c.Database).DatabaseId)).ToList();
+
+            Console.Out.WriteLineAsync($"{backupsToDelete.Count} old backups found.");
+
+            //delete the backups
```
skuruppu, line 176:
```suggestion
            // Delete the backups.
```

spanner/api/SpannerTest/Tests.cs
```diff
@@ -133,13 +133,81 @@ public SpannerTests(SpannerFixture fixture)
...
+            backupsToDelete.ForEach(backup =>
+            {
+                try
+                {
+                    DeleteBackup.SpannerDeleteBackup(_fixture.ProjectId, _fixture.InstanceId, backup.BackupName.BackupId);
+                }
+                catch (Exception) { }
+            });
+
+            //delete the databases
```
skuruppu, line 186:
```suggestion
            // Delete the databases.
```

spanner/api/SpannerTest/Tests.cs
```diff
@@ -96,19 +101,14 @@ public void Dispose()
...
             Environment.GetEnvironmentVariable("TEST_SPANNER_INSTANCE") ?? "my-instance";
-        private static readonly string s_randomDatabaseName = "my-db-"
-            + TestUtil.RandomName();
-        private static readonly string s_randomToBeCancelledBackupName = "my-backup-"
-            + TestUtil.RandomName();
-        private static readonly string s_randomRestoredDatabaseName = "my-db-"
-            + TestUtil.RandomName();
-        public string DatabaseId =
-            Environment.GetEnvironmentVariable("TEST_SPANNER_DATABASE") ?? s_randomDatabaseName;
+        private static readonly string s_randomDatabaseName = $"my-db-{DateTime.UtcNow.ToTimestamp().Seconds}";
```
amanda-tarafa, line 108: Remove this from here...
```suggestion
```

spanner/api/SpannerTest/Tests.cs
```diff
@@ -96,19 +101,14 @@ public void Dispose()
...
+        private static readonly string s_randomToBeCancelledBackupName = $"my-backup-{DateTime.UtcNow.ToTimestamp().Seconds}";
+        private static readonly string s_randomRestoredDatabaseName = $"my-db-{DateTime.UtcNow.AddMinutes(1).ToTimestamp().Seconds}";
+        public string DatabaseId = Environment.GetEnvironmentVariable("TEST_SPANNER_DATABASE") ?? s_randomDatabaseName;
```
amanda-tarafa, line 111: ... and then here:
```suggestion
        public string DatabaseId = Environment.GetEnvironmentVariable("TEST_SPANNER_DATABASE") ?? $"my-db-{DateTime.UtcNow.ToTimestamp().Seconds}";

```

spanner/api/SpannerTest/Tests.cs
```diff
@@ -96,19 +101,14 @@ public void Dispose()
...
-            + TestUtil.RandomName();
-        private static readonly string s_randomToBeCancelledBackupName = "my-backup-"
-            + TestUtil.RandomName();
-        private static readonly string s_randomRestoredDatabaseName = "my-db-"
-            + TestUtil.RandomName();
-        public string DatabaseId =
-            Environment.GetEnvironmentVariable("TEST_SPANNER_DATABASE") ?? s_randomDatabaseName;
+        private static readonly string s_randomDatabaseName = $"my-db-{DateTime.UtcNow.ToTimestamp().Seconds}";
+        private static readonly string s_randomToBeCancelledBackupName = $"my-backup-{DateTime.UtcNow.ToTimestamp().Seconds}";
+        private static readonly string s_randomRestoredDatabaseName = $"my-db-{DateTime.UtcNow.AddMinutes(1).ToTimestamp().Seconds}";
```
amanda-tarafa, line 110:
```suggestion
        private static readonly string s_randomRestoredDatabaseName = $"my-restore-db-{DateTime.UtcNow.ToTimestamp().Seconds}";

```

spanner/api/SpannerTest/Tests.cs
```diff
@@ -96,19 +101,14 @@ public void Dispose()
...
-        private static readonly string s_randomDatabaseName = "my-db-"
-            + TestUtil.RandomName();
-        private static readonly string s_randomToBeCancelledBackupName = "my-backup-"
-            + TestUtil.RandomName();
-        private static readonly string s_randomRestoredDatabaseName = "my-db-"
-            + TestUtil.RandomName();
-        public string DatabaseId =
-            Environment.GetEnvironmentVariable("TEST_SPANNER_DATABASE") ?? s_randomDatabaseName;
+        private static readonly string s_randomDatabaseName = $"my-db-{DateTime.UtcNow.ToTimestamp().Seconds}";
+        private static readonly string s_randomToBeCancelledBackupName = $"my-backup-{DateTime.UtcNow.ToTimestamp().Seconds}";
```
amanda-tarafa, line 109: Seconds will go from 0 to 59, you might want TotalSeconds or even TotalMilliseconds, or better still, format the date to something like "yyyyMMddHHmmss" so it's quickly readable and if we need to look for something that has failed we can find it quickly.
Here and in all the others.

spanner/api/SpannerTest/Tests.cs
```diff
@@ -96,19 +101,14 @@ public void Dispose()
...
+        private static readonly string s_randomRestoredDatabaseName = $"my-db-{DateTime.UtcNow.AddMinutes(1).ToTimestamp().Seconds}";
+        public string DatabaseId = Environment.GetEnvironmentVariable("TEST_SPANNER_DATABASE") ?? s_randomDatabaseName;
         public string BackupDatabaseId = "my-test-database";
         public string BackupId = "my-test-database-backup";
         public string ToBeCancelledBackupId = s_randomToBeCancelledBackupName;
```
amanda-tarafa, line 114: Why don't these need to be random? Just asking, I'm fine if there's no need for it.

spanner/api/SpannerTest/Tests.cs
```diff
@@ -133,13 +133,82 @@ public SpannerTests(SpannerFixture fixture)
...
 
+        void DeleteStaleBackupsAndDatabases()
+        {
+            DatabaseAdminClient databaseAdminClient = DatabaseAdminClient.Create();
+            var instanceName = InstanceName.FromProjectInstance(_fixture.ProjectId, _fixture.InstanceId);
+            var databases = databaseAdminClient.ListDatabases(instanceName).Where(c => c.DatabaseName.DatabaseId.StartsWith("my-db-")).ToList();
+            var databasesToDelete = new List<string>();
+
+            // Delete all the databases created before 4 hrs.
+            var timestamp = DateTime.UtcNow.AddHours(-4).ToTimestamp().Seconds;
```
amanda-tarafa, line 152: This won't work as you think it will, again, you need something like TotalSeconds instead. Or better still if you just format the date and time. Or probably the database entity has a CreatedOn field, you can then create the database with a name from a Guid and use the CreatedOn filed to delete the older ones.

spanner/api/SpannerTest/Tests.cs
```diff
@@ -133,13 +133,82 @@ public SpannerTests(SpannerFixture fixture)
...
+            databases.ForEach(database =>
```
amanda-tarafa, line 153: Always prefer `foreach (var database in databases) { ... }`

spanner/api/SpannerTest/Tests.cs
```diff
@@ -133,13 +133,82 @@ public SpannerTests(SpannerFixture fixture)
...
+            };
+            var backups = databaseAdminClient.ListBackups(request).ToList();
+
+            // Backups that belong to the databases to be deleted.
+            var backupsToDelete = backups.Where(c => databasesToDelete.Contains(DatabaseName.Parse(c.Database).DatabaseId)).ToList();
+
+            Console.Out.WriteLineAsync($"{backupsToDelete.Count} old backups found.");
+
+            // Delete the backups.
+            backupsToDelete.ForEach(backup =>
```
amanda-tarafa, line 178: Same here, always prefer `foreach`. And everywhere else.

spanner/api/SpannerTest/Tests.cs
```diff
@@ -133,13 +133,82 @@ public SpannerTests(SpannerFixture fixture)
...
+                catch (Exception) { }
+            });
+        }
+
+        void InitializeInstance()
+        {
+            InstanceAdminClient instanceAdminClient = InstanceAdminClient.Create();
+            try
+            {
+                string name = $"projects/{_fixture.ProjectId}/instances/{_fixture.InstanceId}";
```
amanda-tarafa, line 203: There's probably an InstanceName class or there should be one. If it's missing, the protos are missing an annotation.

spanner/api/SpannerTest/Tests.cs
```diff
@@ -133,13 +133,82 @@ public SpannerTests(SpannerFixture fixture)
...
                     InitializeBackup();
                 }
             }
         }
 
+        void DeleteStaleBackupsAndDatabases()
+        {
+            DatabaseAdminClient databaseAdminClient = DatabaseAdminClient.Create();
+            var instanceName = InstanceName.FromProjectInstance(_fixture.ProjectId, _fixture.InstanceId);
+            var databases = databaseAdminClient.ListDatabases(instanceName).Where(c => c.DatabaseName.DatabaseId.StartsWith("my-db-")).ToList();
```
amanda-tarafa, line 148: In general, you can avoid the ToList().

spanner/api/SpannerTest/Tests.cs
laljikanjareeya: seconds of UTC time since Unix epoch 1970-01-01T00:00:00Z. Must be from 0001-01-01T00:00:00Z to 9999-12-31T23:59:59Z inclusive.

laljikanjareeya: it starts from `new DateTime(1970, 1, 1,0, 0, 0, DateTimeKind.Utc)`

amanda-tarafa: Oh I see, you are calling ToTimestamp(), which I believe is a an extension method defined in Protobuf.WellKnownTypes.
So, I would prefer not to use protobuf code to do something that we can do with .NET: how about `DateTimeOffset.UtcNow.ToUnixTimeSeconds();`

spanner/api/SpannerTest/Tests.cs
laljikanjareeya: Removed it from everywhere.

spanner/api/SpannerTest/Tests.cs
```diff
@@ -96,23 +100,14 @@ public void Dispose()
             catch (RpcException ex) when (ex.Status.StatusCode == StatusCode.NotFound) { }
         }
 
-        public string ProjectId { get; private set; } =
-            Environment.GetEnvironmentVariable("GOOGLE_PROJECT_ID");
+        public string ProjectId { get; private set; } = Environment.GetEnvironmentVariable("GOOGLE_PROJECT_ID");
```
amanda-tarafa, line 103: You probably don't need the private set, remove it.

spanner/api/SpannerTest/Tests.cs
```diff
@@ -96,23 +100,14 @@ public void Dispose()
...
-            Environment.GetEnvironmentVariable("TEST_SPANNER_INSTANCE") ?? "my-instance";
-        private static readonly string s_randomDatabaseName = "my-db-"
-            + TestUtil.RandomName();
-        private static readonly string s_randomToBeCancelledBackupName = "my-backup-"
-            + TestUtil.RandomName();
-        private static readonly string s_randomRestoredDatabaseName = "my-db-"
-            + TestUtil.RandomName();
-        public string DatabaseId =
-            Environment.GetEnvironmentVariable("TEST_SPANNER_DATABASE") ?? s_randomDatabaseName;
+        public string InstanceId { get; private set; } = Environment.GetEnvironmentVariable("TEST_SPANNER_INSTANCE") ?? "my-instance";
```
amanda-tarafa, line 105: Same here, probably don't need the private set.

spanner/api/SpannerTest/Tests.cs
```diff
@@ -96,23 +100,14 @@ public void Dispose()
...
+        public string DatabaseId = Environment.GetEnvironmentVariable("TEST_SPANNER_DATABASE") ?? $"my-db-{DateTimeOffset.UtcNow.ToUnixTimeSeconds()}";
```
amanda-tarafa, line 106: Make this one a property with only a getter. And same with all the rest that follow.

spanner/api/SpannerTest/Tests.cs
```diff
@@ -96,23 +100,14 @@ public void Dispose()
...
         public string BackupDatabaseId = "my-test-database";
         public string BackupId = "my-test-database-backup";
-        public string ToBeCancelledBackupId = s_randomToBeCancelledBackupName;
-        public string RestoredDatabaseId = s_randomRestoredDatabaseName;
+        public string ToBeCancelledBackupId = $"my-backup-{DateTimeOffset.UtcNow.ToUnixTimeSeconds()}";
+        public string RestoredDatabaseId = $"my-restore-db-{DateTimeOffset.UtcNow.ToUnixTimeSeconds()}";
         public bool s_initializedDatabase { get; set; } = false;
```
amanda-tarafa, line 111: The naming is wrong here either

`public bool DatabaseInitialized {get; set;}` (do you need the set to be public?)
or `private static bool s_databaseInitialized`
or `private bool _databaseInitialized`

spanner/api/SpannerTest/Tests.cs
```diff
@@ -133,13 +128,83 @@ public SpannerTests(SpannerFixture fixture)
                 if (!_fixture.s_initializedDatabase)
                 {
                     _fixture.s_initializedDatabase = true;
```
amanda-tarafa, line 130: Move this to after all the initializations have happened, just in case some of hem fail. On rewrite we will move all of this to the fixture, and there won't be a need to ask whether initialization has happened or not.

spanner/api/SpannerTest/Tests.cs
```diff
@@ -133,13 +128,83 @@ public SpannerTests(SpannerFixture fixture)
...
+        {
+            DatabaseAdminClient databaseAdminClient = DatabaseAdminClient.Create();
+            var instanceName = InstanceName.FromProjectInstance(_fixture.ProjectId, _fixture.InstanceId);
+            var databases = databaseAdminClient.ListDatabases(instanceName)
+                .Where(c => c.DatabaseName.DatabaseId.StartsWith("my-db-") || c.DatabaseName.DatabaseId.StartsWith("my-restore-db-"));
+            var databasesToDelete = new List<string>();
+
+            // Delete all the databases created before 4 hrs.
+            var timestamp = DateTimeOffset.UtcNow.AddHours(-4).ToUnixTimeSeconds();
+            foreach (var database in databases)
```
amanda-tarafa, line 149: Ain't it easier to use the database creation time here? It's fine to use a timestamp for the name, but parsing that back to get the creation time seems a little too much.

laljikanjareeya: I was thinking on same but [In proto](https://github.com/googleapis/googleapis/blob/cd804bab06e46dd1a4f16c32155fd3cddb931b52/google/spanner/admin/database/v1/spanner_database_admin.proto#L364-L365) it says that the create_time might not be available every time(If exists).

laljikanjareeya:
@amanda-tarafa  Create backup failed on `my-test-database` database and we kept that name for restore and backup test.
we kept Backup for `my-test-database` with id `my-test-database-backup` and we are not deleting it at all.

Exception says that backup is already in progress for `my-test-database`.

`
Error: Grpc.Core.RpcException : Status(StatusCode=FailedPrecondition, Detail="Cannot create backup (projects/dotnet-docs-samples-tests/instances/my-instance/backups/my-backup-1594291057) for database (projects/dotnet-docs-samples-tests/instances/my-instance/databases/my-test-database) because the maximum number of pending backups (1) for the database has been reached. Please retry the operation once the pending backups complete.")
`

amanda-tarafa:
@laljikanjareeya Yes, you are right, these are not the new databases. We'll discuss offline the best fix for this.

spanner/api/SpannerTest/Tests.cs
laljikanjareeya: This probably will remove in upcoming rewrite.

spanner/api/SpannerTest/Tests.cs
laljikanjareeya: This will remove in upcoming rewrite.

PR #1120, opened 2020-07-05
Update dotnet yaml comment to reference dotnet instead of python

danmastrow:


amanda-tarafa:
@jskeet ignore the above comment, I misread this as being in functions.

PR #1140, opened 2020-07-22
Redirects stderr to stdout when running tests.

amanda-tarafa:
* This avoids PS failing when dotnet test writes to stderr.
  * Will hopefully improve the quality of test failure messages.

See: https://stackoverflow.com/a/20950421/1122643

I'm testing first with Asset, if it fixes the failure we had there, I'll do this for all files.

(Draft while trial and error)

### DIFF
--- a/BuildTools.psm1
+++ b/BuildTools.psm1
@@ -337,15 +337,6 @@ function Run-TestScripts($TimeoutSeconds=300) {
     Run-TestScriptsOnce $scripts $TimeoutSeconds 'Starting' $results
     # Rename all the test logs to a name Sponge will find.
     Get-ChildItem -Recurse TestResults.xml | Rename-Item -Force -NewName 01_sponge_log.xml
-    # Retry the failures once.
-    $failed = $results['Failed']
-    if ($failed) {
-        $results['Failed'] = @()
-        Run-TestScriptsOnce ($failed | Get-Item) $TimeoutSeconds `
-            'Retrying' $results
-        # Rename all the test logs to a name Sponge will find.
-        Get-ChildItem -Recurse TestResults.xml | Rename-Item -Force -NewName 02_sponge_log.xml
-    }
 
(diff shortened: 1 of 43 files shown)


amanda-tarafa:
FYI @yuyifan-google

amanda-tarafa:
@laljikanjareeya , @skuruppu This fix should make the fail messages less cryptic. I've seen for the Asset samples how we've gone from a Fail <TestName> (similar to what we had in Spanner) to having the whole detail of the failure, exception, stack trace, etc.

amanda-tarafa:
@jskeet This is ready for review, one commit at a time is best.

jskeet:
Great, will look tomorrow

laljikanjareeya:
> @laljikanjareeya , @skuruppu This fix should make the fail messages less cryptic. I've seen for the Asset samples how we've gone from a Fail (similar to what we had in Spanner) to having the whole detail of the failure, exception, stack trace, etc.

@amanda-tarafa  Thanks. This is helpful when something is failing.

BuildTools.psm1
```diff
@@ -263,7 +263,7 @@ function Run-TestScriptsOnce([array]$Scripts, [int]$TimeoutSeconds,
         Write-Output "$verb $relativePath..."
         $job = Start-Job -ArgumentList $relativePath, $script.Directory, `
             ('.\"{0}"' -f $script.Name) {
-            $ErrorActionPreference = "Stop"
+            $ErrorActionPreference = "Continue"
```
jskeet, line 266: Is this still required, if we're redirecting stderr to stdout anyway? I'm fine with it, but I thought I'd mention it seeming redundant.

amanda-tarafa: Yes, it's actually the combination of both that works, redirecting to stdout doesn't work if this is set to "Stop".

PR #1149, opened 2020-07-29
Add Functions sample for ImageMagick

jskeet:
See https://cloud.google.com/functions/docs/tutorials/imagemagick
for examples in other languages.

Currently there's no automated test for this - see README.md for
details.

### DIFF
--- a/functions/imagemagick/ImageMagick/Function.cs
+++ b/functions/imagemagick/ImageMagick/Function.cs
@@ -1,0 +1,164 @@
+// Copyright 2020 Google LLC
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     https://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+// [START functions_imagemagick_setup]
+using CloudNative.CloudEvents;
+using Google.Cloud.Functions.Framework;
+using Google.Cloud.Functions.Invoker.DependencyInjection;
+using Google.Cloud.Storage.V1;
+using Google.Cloud.Vision.V1;
+using Google.Events.SystemTextJson.Cloud.Storage.V1;
+using Grpc.Core;
+using Microsoft.Extensions.DependencyInjection;
+using Microsoft.Extensions.Logging;
+using System;
+using System.Diagnostics;
+using System.IO;
+using System.Threading;
+using System.Threading.Tasks;
+
+// Register the Startup class to configure dependency injection
+[assembly: FunctionsStartup(typeof(ImageMagick.Startup))]
+
+namespace ImageMagick
+{
+    // Dependency injection configuration, executed during server startup.
+    public class Startup : FunctionsStartup
+    {
+        public override void Configure(IFunctionsHostBuilder builder)
+        {
+            builder.Services
+                .AddSingleton(ImageAnnotatorClient.Create())
+                .AddSingleton(StorageClient.Create());
+        }
+    }
+
+    public class Function : ICloudEventFunction<StorageObjectData>
+    {
+        /// <summary>
+        /// The bucket to store blurred images in. An alternative to using environment variables here would be to
+        /// fetch it from IConfiguration.
+        /// </summary>
+        private static readonly string s_blurredBucketName = Environment.GetEnvironmentVariable("BLURRED_BUCKET_NAME");
+
+        private readonly ImageAnnotatorClient _visionClient;
+        private readonly StorageClient _storageClient;
+        private readonly ILogger _logger;
+
+        public Function(ImageAnnotatorClient visionClient, StorageClient storageClient, ILogger<Function> logger) =>
+            (_visionClient, _storageClient, _logger) = (visionClient, storageClient, logger);
+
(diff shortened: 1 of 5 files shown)


functions/imagemagick/ImageMagick/Function.cs
```diff
@@ -0,0 +1,171 @@
...
+            builder.Services
+                .AddSingleton(ImageAnnotatorClient.Create())
+                .AddSingleton(StorageClient.Create());
+        }
+    }
+
+    /// <summary>
+    /// A function that can be triggered in responses to changes in Google Cloud Storage.
+    /// The type argument (StorageObjectData in this case) determines how the event payload is deserialized.
+    /// The event must be deployed so that the trigger matches the expected payload type. (For example,
```
amanda-tarafa, line 50: nit: The event => The function?

functions/imagemagick/ImageMagick/Function.cs
```diff
@@ -0,0 +1,171 @@
...
+            {
+                await _storageClient.DownloadObjectAsync(data.Bucket, data.Name, output, cancellationToken: cancellationToken);
+            }
+
+            // Construct the ImageMagick command
+            string blurredImageFile = Path.GetTempFileName();
+            // Command-line arguments for ImageMagick.
+            // Paths are wrapped in quotes in case they contain spaces.
+            string arguments = $"\"{originalImageFile}\" -blur 0x8, \"{blurredImageFile}\"";
+            ProcessStartInfo psi = new ProcessStartInfo("convert", arguments);
```
amanda-tarafa, line 137: I think you are not using `psi` anywhere.

functions/imagemagick/ImageMagick/Function.cs
```diff
@@ -0,0 +1,171 @@
...
+
+            Process process = Process.Start("convert", arguments);
```
amanda-tarafa, line 139: ImageMagick command is called `convert`? If that's explained somewhere in the docs, that's fine, but it's really confusing otherwise, or for people who just see the code samples.

functions/imagemagick/ImageMagick/Function.cs
```diff
@@ -0,0 +1,171 @@
...
+
+            string uri = $"gs://{s_blurredBucketName}/{data.Name}";
+            _logger.LogInformation("Blurred image uploaded to: {uri}", uri);
+
+            // Remove images from the file system.
+            File.Delete(originalImageFile);
+            File.Delete(blurredImageFile);
+        }
+        // [END functions_imagemagick_blur]
+        // [START functions_imagemagick_setup]
```
amanda-tarafa, line 168: I'm fine with leaving this as is. But one alternative could be to have a function called AnalyzeImageAsync and from HandleAsync only call that function, so on `functions_imagemagick_setup` you can include the whole function code. Then move BlurImageAsync and the new AnalyzeImageAsync to a helper class as static methods and show those in their respectives sections.

functions/imagemagick/ImageMagick/Function.cs
```diff
@@ -0,0 +1,171 @@
...
+            // Perform safe search detection using the Vision API.
+            Image image = Image.FromUri(gcsUri);
+            SafeSearchAnnotation annotation;
+            try
+            {
+                annotation = await _visionClient.DetectSafeSearchAsync(image);
+            }
+            // If the call to the Vision API fails, log the error but let the function complete normally.
+            // If the exceptions weren't caught (and just propagated) the event would be retried.
+            // See the "Best Practices" section in the documentation for more details about retry.
```
amanda-tarafa, line 94: nit: Is there a link that's stable?

functions/imagemagick/runTests.ps1
```diff
@@ -0,0 +1,18 @@
...
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+# There are no tests to run at the moment, but let's check
+# it builds. See README.md for more details.
+
+dotnet build
```
amanda-tarafa, line 18: Add the redirect errout to stdout bit just in case.

functions/imagemagick/ImageMagick/Function.cs
jskeet: Correct, will remove, thanks.

functions/imagemagick/ImageMagick/Function.cs
jskeet: It's in the docs, but I'll add a comment.

functions/imagemagick/ImageMagick/Function.cs
jskeet: I think I'd prefer to keep the separation the way it is, as that's what other languages do - but we *could* do all of that with partial classes. Admittedly the partial-ness would show up in the docs too.

functions/imagemagick/runTests.ps1
jskeet: Done.

functions/imagemagick/ImageMagick/Function.cs
jskeet: I don't know how stable any such link would be, to be honest. I'd prefer to stay a bit vague for now.

jskeet:
Second commit is now present - PTAL, and I'll rebase to a single commit when ready.

functions/imagemagick/ImageMagick/Function.cs
amanda-tarafa: Yep, partial class is good too. But It's fine to keep as is, it's just the brackets.

jskeet:
Failures are unrelated to Functions. Merging.

PR #1150, opened 2020-08-10
Cloud SQL: Update MySQL/Postgres connectivity samples to demonstrate TCP and Unix Socket connections

shubha-rajan:
* Refactored samples to demonstrate connection using both Unix Sockets and TCP. 
* Modified samples to use environment variables for consistency with other language samples.
* Rewrite READMEs for consistency with other connectivity samples.


cloud-sql/postgres/Startup.cs
```diff
@@ -120,11 +102,57 @@ DbConnection NewPostgreSqlConnection()
...
+        }
+
+        DbConnection NewPostgreSqlUnixSocketConnection()
+        {
+            // [START cloud_sql_postgres_dotnet_ado_connection_socket]
+            String dbSocketDir = Environment.GetEnvironmentVariable("DB_SOCKET_PATH") ?? "/cloudsql";
+            String instanceConnectionName = Environment.GetEnvironmentVariable("INSTANCE_CONNECTION_NAME");
+            var connectionString = new NpgsqlConnectionStringBuilder()
+            {
+                // Connecting to a local proxy that does not support ssl.
```
kurtisvg, line 139: nit: (in a couple other places as well) 
```suggestion
                // The Cloud SQL proxy provides encryption between the proxy and service. 

```

shubha-rajan: fixed

cloud-sql/postgres/app.yaml
```diff
@@ -2,5 +2,11 @@
 runtime: aspnetcore
 env: flex
 
+env_variables:
```
amanda-tarafa, line 5: I understand these are samples and that this change actually is meant to make the samples more similar across language. But having a DB password in an environment variable is really not a good practice to show/follow.
I would prefer that you'd reconsider and leave the connections string details on [configuration files](https://docs.microsoft.com/en-us/dotnet/framework/data/adonet/connection-strings-and-configuration-files), which is common in .NET and can be encrypted. Else, do please add a comment on the readme, saying that this is done for the purposes of samples only and that storing a plain text DB password in an environment variable is not secure.

amanda-tarafa: And this comment applies to both the postgres and the sql sample.

kurtisvg: @shubha-rajan Can you make sure [we add the secret manager message](https://github.com/GoogleCloudPlatform/python-docs-samples/blob/948696d8a43f7d6b489a574ae356971471ebcdc7/cloud-sql/mysql/sqlalchemy/main.py#L99) about plaintext secrets around the env vars? 

We should make sure this is on our checklist of things when trying to standardize these samples as well

amanda-tarafa (approved):
LGTM.
Because of the flakyness in other tests I'll have to force merge, so let me know when you are ready.

shubha-rajan:
@amanda-tarafa Tests are running and passing for all 3 database engines and this is ready to merge.

amanda-tarafa (approved):
LGTM!

PR #1151, opened 2020-08-11
Update DeleteNotificationConfigSnippets.g.cs

cluebyte:
Fix typo in the region tag, removing an extraneous "]". The region tag should work after this fix.


amanda-tarafa:
This is a trivial change, and failures in CI are not related. I'm force merging.

PR #1157, opened 2020-08-25
run/events: Use Unmarshal Types Rather Than dynamic. Fixes #1156

### LINKED ISSUE #1156: run/events: Use Unmarshal Types Rather Than dynamic
## Actual

Currently in our run/events samples, we don't use types for "unmarshaling" our JSON payload:

https://github.com/GoogleCloudPlatform/dotnet-docs-samples/tree/master/run

i.e.

```cs
dynamic cloudEventData = JsonConvert.DeserializeObject(body);
```

## Expected

We should use our type library to not have dynamic types. Not sure exactly how it would look in .NET

```cs
MessagePublishedData cloudEventData = ???(body);
```

Type Library:

https://github.com/googleapis/google-cloudevents-dotnet/tree/master/src/Google.Events.Protobuf/Cloud/PubSub/V1

---

I'm not very familiar with .NET. @jskeet can provide more information on the types for .NET.

meteatamel:


meteatamel:
Note that I only made PubSub sample to use the strong event type. The Storage sample only reads CloudEvent headers and I think we want to keep it that way, so it can be used both in Managed and Anthos versions of eventing.

run/events-pubsub/Startup.cs
```diff
@@ -46,38 +45,47 @@ public void Configure(IApplicationBuilder app, IWebHostEnvironment env, ILogger<
...
-                    dynamic cloudEventData = JsonConvert.DeserializeObject(body);
-                    if (cloudEventData == null)
-                    {
-                        context.Response.StatusCode = 400;
-                        await context.Response.WriteAsync("Bad request: No Pub/Sub message received");
-                        return;
-                    }
+                var messagePublishedData = CloudEventConverters.ConvertCloudEventData<MessagePublishedData>(cloudEvent);
+                var pubSubMessage = messagePublishedData?.Message;
+                if (messagePublishedData == null || pubSubMessage == null)
```
amanda-tarafa, line 59: nit: no need to ask for messagePublishData here.
```suggestion
                if (pubSubMessage == null)

```

run/events-pubsub/Startup.cs
```diff
@@ -46,38 +45,47 @@ public void Configure(IApplicationBuilder app, IWebHostEnvironment env, ILogger<
...
+                logger.LogInformation("Received CloudEvent\n" + GetEventLog(cloudEvent));
 
-                    dynamic cloudEventData = JsonConvert.DeserializeObject(body);
-                    if (cloudEventData == null)
-                    {
-                        context.Response.StatusCode = 400;
-                        await context.Response.WriteAsync("Bad request: No Pub/Sub message received");
-                        return;
-                    }
+                var messagePublishedData = CloudEventConverters.ConvertCloudEventData<MessagePublishedData>(cloudEvent);
```
amanda-tarafa, line 57: This one will most probably throw instead of just returing null if it can't convert. I'm OK with not using try-catch here, but then I think you souldn't check for null either, else, it gives the impression that if it can't convert, it will return null.

run/events-pubsub/Startup.cs
```diff
@@ -46,38 +45,47 @@ public void Configure(IApplicationBuilder app, IWebHostEnvironment env, ILogger<
...
 
-                    dynamic pubSubMessage = cloudEventData["message"];
-                    if (pubSubMessage == null)
-                    {
-                        context.Response.StatusCode = 400;
-                        await context.Response.WriteAsync("Bad request: Invalid Pub/Sub message format");
-                        return;
-                    }
+                var data = pubSubMessage.Data;
+                logger.LogInformation($"Data: {data.ToBase64()}");
```
amanda-tarafa, line 67: nit: Why do you show this now?

amanda-tarafa (commented):
A couple of questions mainly.

run/events-pubsub/Startup.cs
meteatamel: Good point, I'll fix

run/events-pubsub/Startup.cs
meteatamel: I tried to reuse the existing code but you're right. Let me see what it actually does (throw or return null) and change the code accordingly.

run/events-pubsub/Startup.cs
meteatamel: I found it helpful to see the base64 encoded string first, before decoding it later..I don't feel strongly about showing it though and can remove if you want.

amanda-tarafa: No, not my heart set on it, if you think it's going to be helpful, that's fine.

run/events-pubsub/Startup.cs
meteatamel: Turns out, `messagePublishedData` is never null. It can just be empty if it's not the right format. So, I changed the code to only check for `pubSubMessage` null, which is possible if there is not `message` field. Also, `cloudEvent` cannot be null either, even if the HTTP Post is not in CloudEvent format. So, I removed the null check there as well.

run/events-pubsub/Startup.cs
meteatamel: I checked and `CloudEventConverters.ConvertCloudEventData` does not throw or return null when the format is not right. It simply returns an empty `MessagePublishedData`. I changed the code to account for that.

run/events-pubsub/Startup.cs
meteatamel: Ok, I'll keep it then.

run/events-pubsub/Startup.cs
jskeet: It's unfortunate that that's pretty much a requirement - events have to be able to add new fields, so we can't throw if we see fields we don't expect... which would be the case if something was "entirely the wrong format".
I believe it *will* throw if it's "somewhat the wrong format" - e.g. an expected-to-be-numeric field that had an object value.
Will think about whether there's anything else we can do about that...

meteatamel:
Let me know if I need to rebase or something to get the PR merged?

amanda-tarafa (approved):
LGTM.
Since you are the owner of these, but you can't review your own PRs, I'll have to force merge (we require two reviews). Let me know if it's ready and I'll do it.

meteatamel:
Yes, it's ready from my side

PR #1171, opened 2020-09-03
Functions samples

jskeet:
(As always, I'd recommend reviewing one commit at a time.)

### DIFF
--- a/functions/firebase/Firebase.Tests/Firebase.Tests.csproj
+++ b/functions/firebase/Firebase.Tests/Firebase.Tests.csproj
@@ -1,0 +1,26 @@
+<Project Sdk="Microsoft.NET.Sdk">
+
+  <PropertyGroup>
+    <TargetFramework>netcoreapp3.1</TargetFramework>
+
+    <IsPackable>false</IsPackable>
+  </PropertyGroup>
+
+  <ItemGroup>
+    <PackageReference Include="JUnitTestLogger" Version="1.1.0" />
+    <PackageReference Include="Google.Cloud.Functions.Invoker.Testing" Version="1.0.0-alpha11" />
+    <PackageReference Include="Microsoft.NET.Test.Sdk" Version="16.5.0" />
+    <PackageReference Include="xunit" Version="2.4.0" />
+    <PackageReference Include="xunit.runner.visualstudio" Version="2.4.0" />
+    <PackageReference Include="coverlet.collector" Version="1.2.0" />
+  </ItemGroup>
+
+  <ItemGroup>
+    <ProjectReference Include="..\FirebaseAuth\FirebaseAuth.csproj" />
+    <ProjectReference Include="..\FirebaseFirestore\FirebaseFirestore.csproj" />
+    <ProjectReference Include="..\FirebaseRemoteConfig\FirebaseRemoteConfig.csproj" />
+    <ProjectReference Include="..\FirebaseRtdb\FirebaseRtdb.csproj" />
+    <ProjectReference Include="..\FirestoreReactive\FirestoreReactive.csproj" />
+  </ItemGroup>
+
+</Project>
--- a/functions/firebase/FirestoreReactive/Function.cs
+++ b/functions/firebase/FirestoreReactive/Function.cs
@@ -1,0 +1,83 @@
+// Copyright 2020 Google LLC
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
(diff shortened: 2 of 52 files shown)


functions/update-google-events.sh
```diff
@@ -0,0 +1,16 @@
+#!/bin/sh
```
amanda-tarafa, line 1: This could be covered later on by the bot that updates dependencies, once I've managed to get it to create one PR per "product" (or at least top level folder) so it can be assigned to owners.

functions/update-google-events.sh
```diff
@@ -0,0 +1,16 @@
...
+  echo "Usage: $script version"
+  echo "e.g. $script 1.2.0"
+  exit 1
+fi
+
+for proj in $(find . -name '*proj')
+do
+  sed -i -e "s/Include=\"Google.Events\" Version=\".*\"/Include=\"Google.Events\" Version=\"$1\"/g" $proj
+  sed -i -e "s/Include=\"Google.Events.Protobuf\" Version=\".*\"/Include=\"Google.Events.Protobuf\" Version=\"$1\"/g" $proj
+  sed -i -e "s/Include=\"Google.Events.SystemTextJson\" Version=\".*\"/Include=\"Google.Events.SystemTextJson\" Version=\"$1\"/g" $proj
```
amanda-tarafa, line 15: Is this back in? I mean the JSON converters?

functions/firebase/Firebase.Tests/Firebase.Tests.csproj
```diff
@@ -0,0 +1,25 @@
+<Project Sdk="Microsoft.NET.Sdk">
+
+  <PropertyGroup>
+    <TargetFramework>netcoreapp3.1</TargetFramework>
+
+    <IsPackable>false</IsPackable>
+  </PropertyGroup>
+
+  <ItemGroup>
+    <PackageReference Include="Google.Cloud.Functions.Invoker.Testing" Version="1.0.0-alpha11" />
```
amanda-tarafa, line 10: You are missing the JUnit dependency, that's why the Firebase tests are failing.

functions/firebase/FirestoreReactive/Function.cs
```diff
@@ -0,0 +1,83 @@
...
+
+            string newValue = currentValue.ToUpperInvariant();
+            if (newValue == currentValue)
+            {
+                _logger.LogInformation("Value is already upper-cased; no replacement necessary");
+                return;
+            }
+
+            // The CloudEvent subject is "documents/x/y/...".
+            // The Firestore SDK FirestoreDb.Document method expects a reference relative to
```
amanda-tarafa, line 69: Just as a comment, obviously not something you can change in samples:
But shouldn't the CloudEvent subject of a Firestore event match whatever Firestore API is expecting? Is that what you are referring to "This may be simplified over time?"

functions/firebase/FirestoreReactive/Function.cs
```diff
@@ -0,0 +1,83 @@
...
+
+namespace FirestoreReactive
+{
+    public class Startup : FunctionsStartup
+    {
+        public override void ConfigureServices(WebHostBuilderContext context, IServiceCollection services) =>
+            services.AddSingleton(FirestoreDb.Create());
+    }
+
+    public class Function : ICloudEventFunction<DocumentEventData>
```
amanda-tarafa, line 39: I think this one doesn't have a test.

amanda-tarafa (approved):
There's a test missing, and also the reference to JUnit in the new Firebase project (those tests are failing)
Otherwise LGTM

functions/update-google-events.sh
jskeet: Not at the moment - but I left it in the replacement as it doesn't hurt :)

functions/update-google-events.sh
jskeet: Right - I'll typically want to be able to do this in a controlled way, often just after releasing. We can delete it if we want though :)

functions/firebase/FirestoreReactive/Function.cs
jskeet: Yes - I expect to add another method in Firestore that accepts something starting "documents/"

functions/firebase/FirestoreReactive/Function.cs
jskeet: You're right. The tricky part is that we'd need to fake out FirestoreClient. I can potentially do that... not sure whether it's worth it though. Thoughts welcome :)

PR #1180, opened 2020-09-11
Give the Firestore Reactive sample the correct tag

jskeet:
(It's a pity that it's tagged as Firebase when it's actually Firestore, but never mind... that's the tag everything uses.)


jskeet:
There may be another of these to add...

jskeet:
No, there isn't - I just missed out a sample altogether.

amanda-tarafa (approved):
LGTM

PR #1188, opened 2020-09-16
Spanner: add NUMERIC samples

skuruppu:


spanner/api/Spanner.Samples.Tests/QueryDataWithNumericParameterTest.cs
```diff
@@ -0,0 +1,44 @@
...
+    [Fact]
+    public async Task TestQueryDataWithNumericParameter()
+    {
+        await _spannerFixture.CreateVenuesTableAndInsertDataAsync();
+
+        AddNumericColumnAsyncSample addColumnSample = new AddNumericColumnAsyncSample();
+        await addColumnSample.AddNumericColumnAsync(_spannerFixture.ProjectId, _spannerFixture.InstanceId, _spannerFixture.DatabaseId);
+
+        UpdateDataWithNumericAsyncSample updateNumericSample = new UpdateDataWithNumericAsyncSample();
+        await updateNumericSample.UpdateDataWithNumericAsync(_spannerFixture.ProjectId, _spannerFixture.InstanceId, _spannerFixture.DatabaseId);
```
skuruppu, line 38: Let me know if it makes more sense to add these to the fixture. I didn't because I didn't want to slow down all other tests since this setup is only required for this test.

spanner/api/Spanner.Samples.Tests/QueryDataWithNumericParameterTest.cs
```diff
@@ -0,0 +1,44 @@
...
+
+    public QueryDataWithNumericParameterTest(SpannerFixture spannerFixture)
+    {
+        _spannerFixture = spannerFixture;
+    }
+
+    [Fact]
+    public async Task TestQueryDataWithNumericParameter()
+    {
+        await _spannerFixture.CreateVenuesTableAndInsertDataAsync();
```
skuruppu, line 32: To give some background, in other languages, we still have the datatypes samples so the NUMERIC samples follow on from that. But we decided to remove it in C# since those samples weren't pulled into the docs. The downside of it is that we have to do this extra setup step.

spanner/api/Spanner.Samples.Tests/QueryDataWithNumericParameterTest.cs
amanda-tarafa: This is fine here, if in the future it is need for more samples, we can move it to the fixture then.

spanner/api/Spanner.Samples.Tests/QueryDataWithNumericParameterTest.cs
```diff
@@ -0,0 +1,44 @@
...
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+using System.Collections.Generic;
+using System.Threading.Tasks;
+using Xunit;
+
+[Collection(nameof(SpannerFixture))]
+public class QueryDataWithNumericParameterTest
```
amanda-tarafa, line 20: The main thing here is that we prefer to have one test per sample. Even if there's a little of code duplication or execution. With something like this, if the test failed, you don't know which part failed at first glance, sometimes the stack trace gets truncated on the test output, etc.

spanner/api/Spanner.Samples.Tests/SpannerFixture.cs
```diff
@@ -220,4 +220,50 @@ public async Task RefillMarketingBudgetsAsync(int firstAlbumBudget, int secondAl
             await cmd.ExecuteNonQueryAsync();
         }
     }
+
+    public async Task CreateVenuesTableAndInsertDataAsync()
+    {
+        string connectionString = $"Data Source=projects/{ProjectId}/instances/{InstanceId}/" +
+            $"databases/{DatabaseId}";
+        // Create connection to Cloud Spanner.
+        using (var connection = new SpannerConnection(connectionString))
```
amanda-tarafa, line 229: nit: (and remove brackets and indent)
```suggestion
        using var connection = new SpannerConnection(connectionString);
```

spanner/api/Spanner.Samples.Tests/SpannerFixture.cs
```diff
@@ -220,4 +220,50 @@ public async Task RefillMarketingBudgetsAsync(int firstAlbumBudget, int secondAl
...
+        {
+            // Define create table statement for Venues.
+            string createTableStatement =
+            @"CREATE TABLE Venues (
+                     VenueId INT64 NOT NULL,
+                     VenueName STRING(1024),
+                 ) PRIMARY KEY (VenueId)";
+
+            var cmd = connection.CreateDdlCommand(createTableStatement);
```
amanda-tarafa, line 238:
```suggestion
            using var cmd = connection.CreateDdlCommand(createTableStatement);
```

spanner/api/Spanner.Samples.Tests/SpannerFixture.cs
```diff
@@ -220,4 +220,50 @@ public async Task RefillMarketingBudgetsAsync(int firstAlbumBudget, int secondAl
...
+            {
+                new Venue { VenueId = 4, VenueName = "Venue 4" },
+                new Venue { VenueId = 19, VenueName = "Venue 19" },
+                new Venue { VenueId = 42, VenueName = "Venue 42" },
+            };
+
+            await Task.WhenAll(venues.Select(venue =>
+            {
+                // Insert rows into the Venues table.
+                var cmd = connection.CreateInsertCommand("Venues", new SpannerParameterCollection
```
amanda-tarafa, line 251:
```suggestion
                using var cmd = connection.CreateInsertCommand("Venues", new SpannerParameterCollection
```

spanner/api/Spanner.Samples/QueryDataWithNumericParameterAsync.cs
```diff
@@ -0,0 +1,57 @@
...
+        public SpannerNumeric Revenue { get; set; }
+    }
+
+    public async Task<List<Venue>> QueryDataWithNumericParameterAsync(string projectId, string instanceId, string databaseId)
+    {
+        string connectionString = $"Data Source=projects/{projectId}/instances/{instanceId}" +
+            $"/databases/{databaseId}";
+
+        using var connection = new SpannerConnection(connectionString);
+        var cmd = connection.CreateSelectCommand(
```
amanda-tarafa, line 36:
```suggestion
        using var cmd = connection.CreateSelectCommand(
```

spanner/api/Spanner.Samples/AddNumericColumnAsync.cs
```diff
@@ -0,0 +1,35 @@
...
+public class AddNumericColumnAsyncSample
+{
+    public async Task AddNumericColumnAsync(string projectId, string instanceId, string databaseId)
+    {
+        string connectionString = $"Data Source=projects/{projectId}/instances/{instanceId}"
+            + $"/databases/{databaseId}";
+        string alterStatement = "ALTER TABLE Venues ADD COLUMN Revenue NUMERIC";
+
+        using var connection = new SpannerConnection(connectionString);
+        var updateCmd = connection.CreateDdlCommand(alterStatement);
```
amanda-tarafa, line 30:
```suggestion
        using var updateCmd = connection.CreateDdlCommand(alterStatement);
```

amanda-tarafa (commented):
Mostly fine, main thing is to try and split the tests even if it means a little bit of duplication, or creating and deleting the venues table as needed.

spanner/api/Spanner.Samples.Tests/QueryDataWithNumericParameterTest.cs
skuruppu: Cool, that makes sense. I've now added separate tests. The only downside is that the tests are not asserting anything since the final test that everything works is really when you run the query data sample.

But I suppose at least we know that an exception wasn't thrown if each test completes.

skuruppu:
Thanks for the quick review @amanda-tarafa. I've made all the changes now so PTAL.

spanner/api/Spanner.Samples.Tests/QueryDataWithNumericParameterTest.cs
amanda-tarafa: Yes, this is fine. If the one adding the column fails, probably all will fail, but at least we can quicly know which test is to blame.

amanda-tarafa (approved):
LGTM.
If you can't merge because of unrelated failures, let me know when it can be merged and I'll do it.

laljikanjareeya (approved):
LGTM.

PR #1191, opened 2020-09-17
[Spanner]: Delete five oldest databases if instance reaches the database limit

### LINKED ISSUE #1189: [Spanner]: Instance Database limit (100) reached easily.
Alternatives:

- Keep decrementing the time to consider a DB stale.
  - Currently at 48hrs which already doesn't give us time to check on weekend failures.
  - If there's a lot of activity on the repo, and since all tests are executed on presubmit, we can always reach limit even if we delete tha databases inmediately after running the tests, we only need a set number of presubmit checks executing at the same time.
  - @laljikanjareeya You mentioned something about there being leftover DBs from before. I just tried deleting them on the UI but they have backups, I will try and submit a PR just for the sake of deleting them later, tweaking the fixture code to include this once DBs whose name is my-db-<not a timestamp>, and then close the PR. Maybe that fixes it and we don't have to consider anything else.
- Consider creating an instance per test run, and deleting the stale instances after 72hours.
  - It would address #1173 
  - What's the max amount of intances that we can have? We would created one per tests run, as opposed to many databases per test run?
  - How long does it take to create an instance?

@skuruppu and @laljikanjareeya thoughst welcomed.

laljikanjareeya:
Fixes #1189 

Before starting the test Check the databases available on instance, if it's more than  94 delete the old five databases.
Removed delete backup logic as we have separate database for same and we never delete it.

### DIFF
--- a/spanner/api/Spanner.Samples.Tests/SpannerFixture.cs
+++ b/spanner/api/Spanner.Samples.Tests/SpannerFixture.cs
@@ -41,7 +41,7 @@ public class SpannerFixture : IAsyncLifetime, ICollectionFixture<SpannerFixture>
         var isExistingInstance = InitializeInstance();
         if (isExistingInstance)
         {
-            await DeleteStaleBackupsAndDatabasesAsync();
+            await DeleteStaleDatabasesAsync();
         }
         await InitializeDatabaseAsync();
         await InitializeBackupAsync();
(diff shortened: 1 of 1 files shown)


spanner/api/Spanner.Samples.Tests/SpannerFixture.cs
```diff
@@ -69,51 +69,36 @@ private bool InitializeInstance()
...
+
+    /// <summary>
+    /// Deletes first 5 database if no of database is more than 94.
+    /// </summary>
+    private async Task DeleteStaleDatabasesAsync()
     {
         DatabaseAdminClient databaseAdminClient = DatabaseAdminClient.Create();
         var instanceName = InstanceName.FromProjectInstance(ProjectId, InstanceId);
         var databases = databaseAdminClient.ListDatabases(instanceName)
             .Where(c => c.DatabaseName.DatabaseId.StartsWith("my-db-") || c.DatabaseName.DatabaseId.StartsWith("my-restore-db-"));
```
amanda-tarafa, line 81: I think you might not want to filter here, just in case database other than "my-db-" and "my-restore-db-" are being created.

spanner/api/Spanner.Samples.Tests/SpannerFixture.cs
```diff
@@ -69,51 +69,36 @@ private bool InitializeInstance()
...
-        var backups = databaseAdminClient.ListBackups(request);
-
-        // Backups that belong to the databases to be deleted.
-        var backupsToDelete = backups.Where(c => databasesToDelete.Contains(DatabaseName.Parse(c.Database).DatabaseId));
-
-        await Console.Out.WriteLineAsync($"{backupsToDelete.Count()} old backups found.");
-
-        // Delete the backups.
-        foreach (var backup in backupsToDelete)
+        var databaseIdsWithCreatedTime = new Dictionary<string, long>();
```
amanda-tarafa, line 88: And I think you can substitute from here to line 99 with, so we don't have to iterate many tames and create a dictionary (can't add suggestion for deleted lines), note that you don't need the where if you write the OrderBy like that:
```suggestion
        var databasesToDelete = databases
            .OrderBy(db => 
                long.TryParse(
                    db.DatabaseName.DatabaseId.Replace("my-db-", "").Replace("my-restore-db-", ""),
                    long out creationDate) ?
                creationDate : long.MaxValue)
            .Take(5);    
```

amanda-tarafa (commented):
Thanks a lot for this! I like better than just deleting always. And if we still have failures we can change to delete the 10 oldest, etc.
I have a sugestion to improve the linq expression and avoid a couple of iterations, but apart from that, this is great.

spanner/api/Spanner.Samples.Tests/SpannerFixture.cs
laljikanjareeya: We have one database named `my-test-database` for backup so keeping this.

amanda-tarafa: You don't need to, because when you try to convert the name to a long, it will fail, you will order by long.MaxValue and it will be the last DB in the list. Unless we delete all DBs, which won't happen, this one won't ever be deleted.

laljikanjareeya: Ah yes.

amanda-tarafa (approved):
LGTM

amanda-tarafa:
I'm rerunning Kokoro because there's a Spanner error, but I think it was something transient server side. CreateBackup and CreateDatabase were aborted because "probably the schema changed during the operation. Retrying may succeed."

amanda-tarafa:
Spanner passed. I'm merging. I'll also close the issue, we can reopen if this happens again, but I think we are safe now.

PR #1196, opened 2020-09-28
Moved run/events-* to eventarc/ and other minor changes to match the …

### LINKED ISSUE #1195: Move run/events-* to eventarc/
The Events for Cloud Run product/API is not separate from Run.

Given that, we should move the `run/events-*` samples to `eventarc/` and change the README and region tags appropriately.

.NET samples:

https://github.com/GoogleCloudPlatform/dotnet-docs-samples/tree/master/run

Example:

https://github.com/GoogleCloudPlatform/nodejs-docs-samples/tree/master/eventarc

---

We should leave the README for `run` since we still have the Knative sample. Probably just use the old README.

meteatamel:
…Node.js repo. Fixes #1195


run/README.md
```diff
@@ -1,116 +0,0 @@
-# Google Cloud Run .NET Samples
```
amanda-tarafa, line 1: Just calling out that this one doesn't have an equivalent in the new location. That might be totally fine, just in case you missed it.

amanda-tarafa (approved):
LGTM

run/README.md
meteatamel: I removed it, as they removed it from Node.js samples as well. I don't think it's maintainable to have an uber README.md anyway. Each sample folder has its own README.md and that should be enough.

amanda-tarafa: Yes, I agree that an uber readme is not the best.
I'm rerunning Kokoro, there was a failure before that I couldn't quite pinpoint, and will merge on green unless you say otherwise.

meteatamel: Thanks @amanda-tarafa

grayside:
The Cloud Run docs depended on `run/.gcloudignore`, with the intention to move this into a helloworld sample here in the next weeks. I'll file a PR to restore this.

PR #1202, opened 2020-10-07
fix(run): restore missing gcloudignore

grayside:
#1196 removed this gcloudignore, which is used directly in the Cloud Run quickstart.

Since I am looking at this, I added some more context preamble and exclusion on git history matching some of the other gcloudignores in the Cloud Run samples.

This is in `run/helloworld/` instead of `run/` because in Q4 we plan to move the Cloud Run C# quickstart here.


meteatamel (approved):
Sorry I didn't realize .gcloudignore was removed and it was important

amanda-tarafa (approved):
LGTM

PR #1206, opened 2020-10-15
Update jobs sample to v4.

palmchou:


amanda-tarafa (changes requested):
These changes won't even build because you need to update the package reference on the .csproj which still references the library for the beta version of the API.

Also, looking at these samples, I see there are only a few tests, way less than there are samples, and the tests at not being run at the moment because the runTests.ps1 file is missing. Can you please add one (you can take a look at the one in [KMS](https://github.com/GoogleCloudPlatform/dotnet-docs-samples/blob/master/kms/api/runTests.ps1)) and also make sure that the few tests that we have are passing?

amanda-tarafa:
Closing as this PR is obsolete. Jobs samples have been removed.

PR #1207, opened 2020-10-15
Don't parse request body when there isn't one

jskeet:
It's a shame this is more long-winded, but we don't really want to
log an exception every time there's an empty request.


amanda-tarafa (approved):
LGTM

PR #1211, opened 2020-10-26
Pub/Sub: regional endpoint in publish ordered message

laljikanjareeya:
Here is the issue in NodeJS: https://github.com/googleapis/nodejs-pubsub/issues/1136


amanda-tarafa (approved):
LGTM code wise. I'll let @anguillanneuf confirm that this is what we want to show.
From the [docs](https://cloud.google.com/pubsub/docs/publisher#using_ordering_keys), I don't get that it is necessary to specify a regional endpoint, only needed if you want all the messages to end up in the same region.

anguillanneuf (approved):

amanda-tarafa:
Merging, as failures are unrelated.

PR #1223, opened 2020-11-25
feat(samples): add samples for analyzeIamPolicy and analyzeIamPolicyLongrunning

donghez-google:


asset/api/Asset.Samples/AnalyzeIamPolicyLongrunning.cs
```diff
@@ -0,0 +1,112 @@
...
+        {
+            AnalysisQuery =
+              new IamPolicyAnalysisQuery
+              {
+                  Scope = scope,
+                  ResourceSelector =
+                    new IamPolicyAnalysisQuery.Types.ResourceSelector
+                    {
+                        FullResourceName = fullResourceName,
+                    },
```
amanda-tarafa, line 40: better like this, here and everywhere.

```suggestion
                  ResourceSelector = new IamPolicyAnalysisQuery.Types.ResourceSelector
                  {
                      FullResourceName = fullResourceName,
                  },
```

asset/api/Asset.Samples/AnalyzeIamPolicyLongrunning.cs
```diff
@@ -0,0 +1,112 @@
...
+                    new IamPolicyAnalysisOutputConfig.Types.GcsDestination
+                    {
+                        Uri = uri,
+                    },
+              },
+        };
+
+        // Start the analyze long-running operation
+        var operation = client.AnalyzeIamPolicyLongrunning(request);
+        // Wait for it to complete (or fail)
```
amanda-tarafa, line 61: `PollUntilCompleted()` won't fail. Instead `operation.Faulted` will be true and `operation.Exception` will contain the exception.
Not sure in how much detail you want to get into here, but just though to flag this because this comments makes it look like PollUntilCompleted() will throw.

asset/api/Asset.Samples/AnalyzeIamPolicyLongrunning.cs
```diff
@@ -0,0 +1,112 @@
...
+        operation = operation.PollUntilCompleted();
+        // Return the metadata
+        return operation.Metadata.ToString();
+    }
+
+    public string AnalyzeIamPolicyLongrunningBigquery(string scope, string fullResourceName,
```
amanda-tarafa, line 67: We write one sample per file only.
And are these two samples in the same region tag? Still you can add the same region tag to both files, but it just seems weird to have two samples in the same region tag even if they are similar.

asset/api/Asset.Samples/AnalyzeIamPolicyLongrunning.cs
```diff
@@ -0,0 +1,112 @@
...
+                    },
+              },
+        };
+
+        // Start the analyze long-running operation
+        var operation = client.AnalyzeIamPolicyLongrunning(request);
+        // Wait for it to complete (or fail)
+        operation = operation.PollUntilCompleted();
+        // Return the metadata
+        return operation.Metadata.ToString();
```
amanda-tarafa, line 64: Please don't call ToString() here. If you want to check against the string in tests, then it's OKish, but the sample should just show the Metadata object being returned, which in this case is the request object.
(here and everywhere where this is done).

asset/api/Asset.Samples.Tests/AnalyzeIamPolicyLongrunningTest.cs
```diff
@@ -0,0 +1,59 @@
...
+
+    [Fact]
+    public void TestAnalyzeIamPolicyLongrunningGcs()
+    {
+        // Run the sample code.
+        string scope = String.Format("projects/{0}", _fixture.ProjectId);
+        string fullResourceName =
+            String.Format("//cloudresourcemanager.googleapis.com/projects/{0}", _fixture.ProjectId);
+        string uri = String.Format("gs://{0}/my-analysis.json", _fixture.BucketName);
+        string metadata = _sample.AnalyzeIamPolicyLongrunningGcs(scope, fullResourceName, uri);
```
amanda-tarafa, line 40: If instead of returning a string from the sample you return the actual object, which seems to always be the request, then you can just navigate to the property that interests you and check it's value.
That's clearer both for maintainers of the code, and for users that might see this code. Also, it will catch breaking changes and errors faster. (This comment applies everywhere)

amanda-tarafa (changes requested):
I've reviewed the long running samples only, but most requested changes there will apply to the non-longrunning as well.
Let me know if I can help.

asset/api/Asset.Samples.Tests/AssetFixture.cs
```diff
@@ -56,7 +53,8 @@ public AssetFixture()
     public void Dispose()
     {
         _bucketFixture.Dispose();
-        _bigQueryClient.DeleteDataset(datasetId: DatasetId);
+        _bigQueryClient.DeleteDataset(datasetId: DatasetId,
+                                      options: new DeleteDatasetOptions() { DeleteContents = true });
```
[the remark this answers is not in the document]
donghez-google: Done

asset/api/Asset.Samples/AnalyzeIamPolicyLongrunning.cs
donghez-google: Done

asset/api/Asset.Samples/AnalyzeIamPolicyLongrunning.cs
donghez-google: Got it. They were for the same RPC but with different output options, so I grouped them into one file. Now I split them into two as per your recommendation.

asset/api/Asset.Samples/AnalyzeIamPolicyLongrunning.cs
donghez-google: Done. Returned the request object.

asset/api/Asset.Samples.Tests/AnalyzeIamPolicyLongrunningTest.cs
donghez-google: Make sense. Done.

asset/api/Asset.Samples/AnalyzeIamPolicyLongrunning.cs
donghez-google: Just want to wait for the completion. Updated the comment. Thanks!

garywheeler69 (approved):

asset/api/Asset.Samples.Tests/AssetFixture.cs
```diff
@@ -56,7 +53,8 @@ public AssetFixture()
     public void Dispose()
     {
         _bucketFixture.Dispose();
```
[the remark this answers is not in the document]
donghez-google: Added

asset/api/Asset.Samples.Tests/AssetFixture.cs
```diff
@@ -55,8 +52,16 @@ public AssetFixture()
 
     public void Dispose()
     {
-        _bucketFixture.Dispose();
-        _bigQueryClient.DeleteDataset(datasetId: DatasetId);
+        try
+        {
+            _bucketFixture.Dispose();
+            _bigQueryClient.DeleteDataset(
+              datasetId: DatasetId, options: new DeleteDatasetOptions() { DeleteContents = true });
```
amanda-tarafa, line 59: nit: Indents are always 4 spaces. Check here and everywhere.
```suggestion
            _bigQueryClient.DeleteDataset(
                datasetId: DatasetId, options: new DeleteDatasetOptions() { DeleteContents = true });
```

asset/api/Asset.Samples.Tests/AnalyzeIamPolicyTest.cs
```diff
@@ -0,0 +1,43 @@
...
+    {
+        _fixture = fixture;
+        _sample = new AnalyzeIamPolicySample();
+    }
+
+    [Fact]
+    public void TestAnalyzeIamPolicy()
+    {
+        // Run the sample code.
+        string scope = String.Format("projects/{0}", _fixture.ProjectId);
```
amanda-tarafa, line 36:
```suggestion
        string scope = $"projects/{_fixture.ProjectId}";
```

asset/api/Asset.Samples.Tests/AnalyzeIamPolicyTest.cs
```diff
@@ -0,0 +1,43 @@
...
+        string fullResourceName =
+            String.Format("//cloudresourcemanager.googleapis.com/projects/{0}", _fixture.ProjectId);
```
amanda-tarafa, line 38:
```suggestion
        string fullResourceName =
            $"//cloudresourcemanager.googleapis.com/projects/{_fixture.ProjectId}";
```

asset/api/Asset.Samples.Tests/AnalyzeIamPolicyTest.cs
```diff
@@ -0,0 +1,43 @@
...
+        var result = _sample.AnalyzeIamPolicy(scope, fullResourceName);
+
+        Assert.Contains(fullResourceName, result.ToString());
```
amanda-tarafa, line 41: Shouldn't this better be:
```suggestion
        Assert.Equals(fullResourceName, result.MainAnalysis.AnalysisQuery.ResourceSelector.FullResourceName);
```

asset/api/Asset.Samples.Tests/AnalyzeIamPolicyLongrunningBigqueryTest.cs
```diff
@@ -0,0 +1,49 @@
...
+
+    [Fact]
+    public void TestAnalyzeIamPolicyLongrunningBigquery()
+    {
+        // Run the sample code.
+        string scope = String.Format("projects/{0}", _fixture.ProjectId);
+        string fullResourceName =
+            String.Format("//cloudresourcemanager.googleapis.com/projects/{0}", _fixture.ProjectId);
+        string dataset =
+            String.Format("projects/{0}/datasets/{1}", _fixture.ProjectId, _fixture.DatasetId);
```
amanda-tarafa, line 41: Same suggestion here as in the other test re how to build the strings.

asset/api/Asset.Samples.Tests/AnalyzeIamPolicyLongrunningGcsTest.cs
```diff
@@ -0,0 +1,46 @@
...
+    }
+
+    [Fact]
+    public void TestAnalyzeIamPolicyLongrunningGcs()
+    {
+        // Run the sample code.
+        string scope = String.Format("projects/{0}", _fixture.ProjectId);
+        string fullResourceName =
+            String.Format("//cloudresourcemanager.googleapis.com/projects/{0}", _fixture.ProjectId);
+        string uri = String.Format("gs://{0}/my-analysis.json", _fixture.BucketName);
```
amanda-tarafa, line 40: Same suggestions here as in the other test.

amanda-tarafa (changes requested):
Mostly nits, thanks!

asset/api/Asset.Samples.Tests/AnalyzeIamPolicyLongrunningBigqueryTest.cs
donghez-google: Done

asset/api/Asset.Samples.Tests/AnalyzeIamPolicyLongrunningGcsTest.cs
donghez-google: Done

donghez-google:
> Mostly nits, thanks!

Thanks for reviewing!

asset/api/Asset.Samples.Tests/AnalyzeIamPolicyLongrunningGcsTest.cs
```diff
@@ -0,0 +1,63 @@
...
+// the License.
+//
+// [START asset_quickstart_analyze_iam_policy_longrunning_gcs]
+
+using Google.Api.Gax;
+using Google.Cloud.Asset.V1;
+using System.Collections.Generic;
+using System.Linq;
+
+public class AnalyzeIamPolicyLongrunningGcsSample
```
amanda-tarafa, line 22: This one should be the test for the Longrunning GCS sample, and it seems to be the sample all over again. This was probably an accidental copy paste, since there was a test here before :).

amanda-tarafa (changes requested):
One test that got replaced by its sample ;)

asset/api/Asset.Samples.Tests/AnalyzeIamPolicyLongrunningGcsTest.cs
donghez-google: Good catch. Sorry it's a copy paste error by accident.

amanda-tarafa (approved):
LGTM

amanda-tarafa:
Failures are unrelated. We reached IAM quota. I can force merge when you say it's ready.

donghez-google:
> Failures are unrelated. We reached IAM quota. I can force merge when you say it's ready.

All tests passed in the latest run.

amanda-tarafa:
Merging... Thanks!

PR #1229, opened 2020-12-11
docs: Wrap samples with future prefix

Strykrol:
Standardizing Security Command Center samples to use 'securitycenter' prefixing. 

- Wrapped existing samples to keep published doclinks unbroken
- Replace region tags that aren't published. 

Once this PR is through, published sample inclusions will be updated to use the new prefix, then I'll come through again and remove the unused block wraps.


amanda-tarafa (approved):
LGTM.
@hsanchez94 Not sure if you want to take a look before merging. Let me know. Else I can force merge.

PR #1240, opened 2021-01-13
feat(IAP): Changes the IAP sample to use Application Default Credentials.

amanda-tarafa:
The [docs](https://cloud.google.com/iap/docs/authentication-howto#obtaining_an_oidc_token_for_the_default_service_account) where this sample is surfaced refers to using default credentials only.

### DIFF
--- a/identity-aware-proxy/IAPClient/IAPClient.cs
+++ b/identity-aware-proxy/IAPClient/IAPClient.cs
@@ -30,23 +30,20 @@ public class IAPClient
     /// </summary>
     /// <param name="iapClientId">The client ID observed on 
     /// https://console.cloud.google.com/apis/credentials. </param>
-    /// <param name="credentialsFilePath">Path to the credentials .json file
-    /// downloaded from https://console.cloud.google.com/apis/credentials.
-    /// </param>
     /// <param name="uri">HTTP URI to fetch.</param>
     /// <param name="cancellationToken">The token to propagate operation cancel notifications.</param>
     /// <returns>The HTTP response message.</returns>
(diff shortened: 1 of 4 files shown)


identity-aware-proxy/IAPClient/IAPClient.cs
```diff
@@ -62,16 +59,13 @@ public class IAPClient
...
     /// <param name="cancellationToken">The token to propagate operation cancel notifications.</param>
     /// <returns>The HTTP response message.</returns>
-    public async Task<OidcToken> GetOidcTokenAsync(string iapClientId, string credentialsFilePath, CancellationToken cancellationToken)
+    public async Task<OidcToken> GetOidcTokenAsync(string iapClientId, CancellationToken cancellationToken)
     {
-        // Read credentials from the credentials .json file.
+        // Obtain the application default credentials.
         GoogleCredential credential = await GoogleCredential
-            .FromFileAsync(credentialsFilePath, cancellationToken).ConfigureAwait(false);
+            .GetApplicationDefaultAsync(cancellationToken).ConfigureAwait(false);
```
jskeet, line 68: Do we need the ConfigureAwait here? We don't have it elsewhere, and without that we could unwrap the line.

jskeet (approved):
One suggestion, but basically fine.

identity-aware-proxy/IAPClient/IAPClient.cs
amanda-tarafa: I'll remove before merging. It's really mostly in case someone copy/pastes. But they should know...

amanda-tarafa:
Unrelated failures. Force merging.

PR #1244, opened 2021-01-25
Channel services csharp sample

Osmosis311:
This is a C# sample application for end-to-end customer provisioning on the Channel Services API.  A similar version to this (with annotations for regions for online documentation purposes) will live in the Channel Services documentation, alongside similar code labs in other languages:

https://cloud.google.com/channel/docs/codelabs/workspace/provisioning

The sample creates a single customer and provisions a single Google Workspace entitlement.

The camelCase for the user-defined values is inconsistent with general C# formatting (and the rest of the code in this sample), but was meant to be consistent with the code from all the languages, so the documentation can list these variables in a similar way when explaining how/what to replace.

### DIFF
--- a/channel/api/snippets/WorkspaceProvisioning/Program.cs
+++ b/channel/api/snippets/WorkspaceProvisioning/Program.cs
@@ -1,0 +1,156 @@
+using System;
+using Google.Cloud.Channel.V1;
+using Google.Apis.Auth.OAuth2;
+using System.IO;
+using System.Linq;
+using Google.Api.Gax;
+using Google.Type;
+
+namespace CodeLab
+{
+    class Program
+    {
+        /***************** REPLACE WITH YOUR OWN VALUES ********************************/
+        private static readonly String jsonPrivateKeyFile = "path/to/json_key_file.json";
+        private static readonly String resellerAdminUser = "admin@yourresellerdomain.com";
+        private static readonly String accountId = "C012345";
+        private static readonly String customerDomain = "example.com";
+        /*******************************************************************************/
+
+        private static readonly String accountParent = "accounts/" + accountId;
+
+        private static CloudChannelServiceClient cloudChannelServiceClient;
+
+        static void Main(string[] args)
+        {
+            GoogleCredential credential;
+
+            using (var stream = new FileStream(jsonPrivateKeyFile, FileMode.Open, FileAccess.Read))
+            {
+                credential =
+                    GoogleCredential.FromStream(stream)
+                        .CreateScoped(new string[] {"https://www.googleapis.com/auth/apps.order"})
+                        .CreateWithUser(resellerAdminUser);
+            }
+
+            var client =
+                new CloudChannelServiceClientBuilder
+                    {
+                        TokenAccessMethod = ((ITokenAccess) credential).GetAccessTokenForRequestAsync
+                    }
+                    .Build();
+
+            // For the purpose of this codelab, the code lists all offers and picks
+            // the first offer for Google Workspace Business Standard on an Annual
+            // plan. This is needed because offerIds vary from one account to another,
+            // but this is not a recommended model for your production integration
+            PagedEnumerable<ListOffersResponse, Offer> listOffersResponse =
+                client.ListOffers(new ListOffersRequest {Parent = accountParent});
+            Offer selectedOffer = new Offer();
+            String sampleOffer = "Google Workspace Business Standard";
+            String samplePlan = "Commitment";
+
+            foreach (Offer offer in listOffersResponse.ToList())
+            {
+                String offerName = offer.Sku.MarketingInfo.DisplayName;
+                String offerPlan = offer.Plan.PaymentPlan.ToString();
+                if (offerName == sampleOffer && offerPlan == samplePlan)
+                {
+                    selectedOffer = offer;
+                    break;
+                }
+            }
+
+            CheckCloudIdentityAccountsExistRequest checkCloudIdentityAccountsExistRequest =
+                new CheckCloudIdentityAccountsExistRequest
+                {
+                    Parent = accountParent,
+                    Domain = customerDomain
+                };
+
+            CheckCloudIdentityAccountsExistResponse checkIdentityResponse =
+                client.CheckCloudIdentityAccountsExist(checkCloudIdentityAccountsExistRequest);
+
+            if (checkIdentityResponse.CloudIdentityAccounts.Count > 0)
+            {
+                throw new Exception(
+                    "Cloud identity already exists. Customer must be transferred. Out of scope for this codelab");
+            }
+
+            PostalAddress postalAddress = new PostalAddress
+            {
+                AddressLines = {"123 Main St"},
+                PostalCode = "97224", RegionCode = "US"
+            };
+
+            Customer customer = new Customer
+            {
+                OrgDisplayName = "Acme Corp", OrgPostalAddress = postalAddress, Domain = customerDomain
+            };
+
+            CreateCustomerRequest createCustomerRequest =
+                new CreateCustomerRequest {Parent = accountParent, Customer = customer};
+
+            Customer createCustomerResponse = client.CreateCustomer(createCustomerRequest);
+
+            CloudIdentityInfo cloudIdentityInfo =
+                new CloudIdentityInfo
+                {
+                    AlternateEmail = "marty.mcfly@gmail.com",
+                    LanguageCode = "en-US"
+                };
+
+            AdminUser adminUser = new AdminUser
+            {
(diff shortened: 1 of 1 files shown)


channel/api/snippets/WorkspaceProvisioning/Program.cs
```diff
@@ -0,0 +1,174 @@
...
+                new ProvisionCloudIdentityRequest
+                {
+                    Customer = createCustomerResponse.Name,
+                    CloudIdentityInfo = cloudIdentityInfo,
+                    User = adminUser
+                };
+
+            var provisionCloudIdentityOp = client.ProvisionCloudIdentity(provisionCloudIdentityRequest);
+
+            PollSettings pollSettings = new PollSettings(Expiration.FromTimeout(TimeSpan.FromSeconds(30)),
```
max-mathieu, line 128: Did you confirm this was required to do polling? I'd rather be consistent across languages in the codelab

channel/api/snippets/WorkspaceProvisioning/Program.cs
```diff
@@ -0,0 +1,174 @@
...
+using Google.Api.Gax;
+using Google.Cloud.Channel.V1;
+using Google.Type;
+using System;
+using System.IO;
+using System.Linq;
+
+namespace CodeLab
+{
+    class Program
```
max-mathieu, line 11: Can you add regions so the snippets are extracted correctly in https://cloud.google.com/channel/docs/codelabs/workspace/provisioning

See https://github.com/max-mathieu/nodejs-channel/blob/master/samples/workspace-codelab.js for the new regions we have to use

channel/api/snippets/WorkspaceProvisioning/Program.cs
```diff
@@ -0,0 +1,174 @@
...
+            {
+                String offerName = offer.Sku.MarketingInfo.DisplayName;
+                String offerPlan = offer.Plan.PaymentPlan.ToString();
+                if (offerName == sampleOffer && offerPlan == samplePlan)
+                {
+                    selectedOffer = offer;
+                    break;
+                }
+            }
+
```
max-mathieu, line 67: Can you add some console logs like in https://github.com/max-mathieu/nodejs-channel/blob/master/samples/workspace-codelab.js

The idea is for the person going through the codelab to possibly run those one step at a time and get used to the various resources' structure.

channel/api/snippets/WorkspaceProvisioning/Program.cs
```diff
@@ -0,0 +1,174 @@
...
+        {
+            GoogleCredential credential;
+
+            using (var stream = new FileStream(jsonPrivateKeyFile, FileMode.Open, FileAccess.Read))
+            {
+                credential =
+                    GoogleCredential.FromStream(stream)
+                        .CreateScoped(new string[] {"https://www.googleapis.com/auth/apps.order"})
+                        .CreateWithUser(resellerAdminUser);
+            }
```
amanda-tarafa, line 34: Any particular reason you don't use Application Default Credentials specifying that they need to be a  set to a service account credential (for impersonation)?
```suggestion
            GoogleCredential credential = GoogleCredential
                .GetApplicationDefault()
                .CreateScoped(CloudChannelServiceClient.DefaultScopes)
                .CreateWithUser(resellerAdminUser);
```

channel/api/snippets/WorkspaceProvisioning/Program.cs
```diff
@@ -0,0 +1,174 @@
...
+            foreach (Offer offer in listOffersResponse.ToList())
+            {
+                String offerName = offer.Sku.MarketingInfo.DisplayName;
+                String offerPlan = offer.Plan.PaymentPlan.ToString();
+                if (offerName == sampleOffer && offerPlan == samplePlan)
+                {
+                    selectedOffer = offer;
+                    break;
+                }
+            }
```
amanda-tarafa, line 66: First suggestion would be to not call ToList(), so you don't fetch all offers.
But really, all of this could be written in a more idiomatic way:
```suggestion
            selectedOffer = listOffersResponse.Where(o => o.Sku.MarketingInfo.DisplayName == sampleOffer && o.Plan.PaymentPlan == samplePlan).FirstOrDefault();
```

channel/api/snippets/WorkspaceProvisioning/Program.cs
```diff
@@ -0,0 +1,174 @@
...
+
+            if (checkIdentityResponse.CloudIdentityAccounts.Count > 0)
+            {
+                throw new Exception(
+                    "Cloud identity already exists. Customer must be transferred. Out of scope for this codelab");
+            }
+
+            PostalAddress postalAddress = new PostalAddress
+            {
+                AddressLines = {
```
amanda-tarafa, line 86: Here and everywhere, curly brackets on a new line.

```suggestion
                AddressLines = 
                {
```

channel/api/snippets/WorkspaceProvisioning/Program.cs
amanda-tarafa: I can confirm this is not needed.

channel/api/snippets/WorkspaceProvisioning/Program.cs
amanda-tarafa: +1 On this repo we usually don't store samples that are not shown in docs.
Addign region tags will also help with knowing where to split this in individual methods and that will help greatly with testing.

channel/api/snippets/WorkspaceProvisioning/Program.cs
```diff
@@ -0,0 +1,174 @@
...
+
+            PollSettings pollSettings = new PollSettings(Expiration.FromTimeout(TimeSpan.FromSeconds(30)),
+                TimeSpan.FromSeconds(1));
+
+            provisionCloudIdentityOp.PollUntilCompleted(pollSettings);
+
+            RenewalSettings renewalSettings =
+                new RenewalSettings
+                {
+                    EnableRenewal = true, PaymentPlan = PaymentPlan.Commitment,
```
amanda-tarafa, line 136:
```suggestion
                    EnableRenewal = true,
                    PaymentPlan = PaymentPlan.Commitment,
```

amanda-tarafa (changes requested):
I've left some specific comments in the code. Apart from that:
- This is missing tests.
- This is missing project and solution files.
- Ideally we split the snippets in individual methods, and have a file per snippet (see the KMS sample set) and I can't see anything in the codelab that would be affected by that.

Let me know how I can help.

Osmosis311:
@amanda-tarafa thanks for the review!  Made some of the requested changes, still more to do.  Next step is splitting out into separate functions.  May need some guidance on testing.

amanda-tarafa:
> @amanda-tarafa thanks for the review! Made some of the requested changes, still more to do. Next step is splitting out into separate functions. May need some guidance on testing.

Thanks! Just request my review again when you are ready. And ping me when you get to testing, I can help there no problem. Keep an eye on the [KMS sample set](https://github.com/GoogleCloudPlatform/dotnet-docs-samples/tree/master/kms/api) as well.

Osmosis311:
@amanda-tarafa happy to split up into separate methods, however, regarding the snippets, just want to make sure we're aligned: we're considering this entire file to be one "snippet," which accomplishes a single task: end-to-end provisioning of a customer + entitlement.  Basically mimicking what a reseller would go through to create a standard customer and their Workspace service.  In that light, does it make sense to keep it all in one file?  Or split into multiple files?

amanda-tarafa:
> @amanda-tarafa happy to split up into separate methods, however, regarding the snippets, just want to make sure we're aligned: we're considering this entire file to be one "snippet," which accomplishes a single task: end-to-end provisioning of a customer + entitlement. Basically mimicking what a reseller would go through to create a standard customer and their Workspace service. In that light, does it make sense to keep it all in one file? Or split into multiple files?

There are other split snippets on this repo that are also used as part of a tutorial that shows how to perform one or two end to end actions. That's fine. On the codelab, you don't show the whole thing at once, but you go step by step, so you really don't need the code to be all together. The main advantages to having the snippets split is that they are easier to maintain and test. For testing, you could split them in different methods in the same file, but we have found that this leads to maintenance issues and less clear code when code from one snippet is used in another, etc. That's why we now require a snippet per file (unless there's a good reason), where a snippet usually shows the usage of **one** service client method, i.e: creates the client, initializaes request parameters, makes the service call, handles the response.

amanda-tarafa:
Ping on this PR. If it's not relevant anymore we can close. Else please address requested changes so we can merge. Thanks.

amanda-tarafa:
@Osmosis311 I'm closing this PR as it's been stale for over a year. Do reopen or create a new one when you are ready to work on it. Thanks!

PR #1247, opened 2021-01-26
feat(storage): add PublicAccessPrevention samples

tritone:
Add snippets to support this new feature. See reference
samples at googleapis/nodejs-storage#1310 and library
PR at googleapis/google-cloud-dotnet#5821

PR #1248, opened 2021-01-27
fix(storage): update CustomTime and NonCurrentTime

tritone:
This updates the sample to account for a breaking change in the
apiary client. This PR also bumps the apiary library to the latest
released version.

### DIFF
--- a/storage/api/Storage.Samples/EnableBucketLifecycleManagement.cs
+++ b/storage/api/Storage.Samples/EnableBucketLifecycleManagement.cs
@@ -49,14 +49,12 @@ public class EnableBucketLifecycleManagementSample
             Console.WriteLine($"Age: \t{rule.Condition.Age}");
             Console.WriteLine($"Created Before: \t{rule.Condition.CreatedBefore}");
             Console.WriteLine($"Time Before: \t{rule.Condition.CustomTimeBefore}");
-            Console.WriteLine($"Time Before Raw: \t{rule.Condition.CustomTimeBefore}");
             Console.WriteLine($"Days Since Custom Time: \t{rule.Condition.DaysSinceCustomTime}");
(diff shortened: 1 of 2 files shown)


amanda-tarafa (approved):
LGTM.
But maybe change the library version on this PR instead of in #1247. The change will work either way, but it makes more sense once we are using the latest library version.

tritone:
> LGTM.
> But maybe change the library version on this PR instead of in #1247. The change will work either way, but it makes more sense once we are using the latest library version.

Done!

amanda-tarafa:
Thanks, let's wait for tests and @frankyn to confirm.

storage/api/Storage.Samples/EnableBucketLifecycleManagement.cs
```diff
@@ -49,14 +49,14 @@ public Bucket EnableBucketLifecycleManagement(string bucketName = "your-unique-b
             Console.WriteLine($"Age: \t{rule.Condition.Age}");
             Console.WriteLine($"Created Before: \t{rule.Condition.CreatedBefore}");
             Console.WriteLine($"Time Before: \t{rule.Condition.CustomTimeBefore}");
-            Console.WriteLine($"Time Before Raw: \t{rule.Condition.CustomTimeBeforeRaw}");
+            Console.WriteLine($"Time Before Raw: \t{rule.Condition.CustomTimeBefore}");
```
frankyn, line 52: I think this line should be deleted as it's printed above. Same below.

frankyn (changes requested):

storage/api/Storage.Samples/EnableBucketLifecycleManagement.cs
tritone: Good catch, done.

frankyn (approved):

PR #1256, opened 2021-02-09
feat: add LogCommitStats sample for Spanner

olavloite:
Adds a sample for `LogCommitStats` for Spanner.

### DIFF
--- a/spanner/api/Spanner.Samples/LogCommitStatsAsync.cs
+++ b/spanner/api/Spanner.Samples/LogCommitStatsAsync.cs
@@ -1,0 +1,87 @@
+﻿// Copyright 2021 Google Inc.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+// [START spanner_get_commit_stats]
+
+using Google.Cloud.Spanner.Data;
+using Google.Cloud.Spanner.V1;
+using Google.Cloud.Spanner.V1.Internal.Logging;
+using System;
+using System.Collections.Generic;
+using System.Diagnostics;
+using System.Threading.Tasks;
+
+public class LogCommitStatsAsyncSample
+{
+    public async Task<long> LogCommitStatsAsync(string projectId, string instanceId, string databaseId)
+    {
+        // Commit statistics are logged at level Info by the default logger.
+        // This sample uses a custom logger to access the commit statistics.
+        // See https://googleapis.github.io/google-cloud-dotnet/docs/Google.Cloud.Spanner.Data/logging.html
+        // for more information on how to use loggers.
+        var logger = new CommitStatsSampleLogger();
+        var options = new SessionPoolOptions();
+        var poolManager = SessionPoolManager.Create(options, logger);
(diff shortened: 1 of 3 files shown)


skuruppu (approved):

skuruppu:
Friendly ping on this PR @amanda-tarafa. This is feature is launching very soon.

spanner/api/Spanner.Samples/LogCommitStatsAsync.cs
```diff
@@ -0,0 +1,87 @@
...
+        Console.WriteLine($"{mutationCount} mutation(s) in transaction...");
+
+        return mutationCount;
+    }
+
+    /// <summary>
+    /// Sample logger that keeps a reference to the last seen commit response.
+    /// Use the default logger if you only want to log the commit stats.
+    /// </summary>
+    public class CommitStatsSampleLogger : Logger
```
amanda-tarafa, line 62: Why do we wan to show a custom logger? In particular this one that seem to add nothing (except maybe for our own testing purposes). I think the idea was that clients shouldn't have to change their code and if the only sample we have includes a custom logger, then I wonder if we are sending the wrong signal?

spanner/api/Spanner.Samples/LogCommitStatsAsync.cs
```diff
@@ -0,0 +1,87 @@
...
+        await connection.OpenAsync();
+
+        using var cmd = connection.CreateDmlCommand("INSERT Singers (SingerId, FirstName, LastName) VALUES (110, 'Virginia', 'Watson')");
+        var rowCount = await cmd.ExecuteNonQueryAsync();
+        var mutationCount = logger._lastCommitResponse.CommitStats.MutationCount;
+
+        Console.WriteLine($"{rowCount} row(s) inserted...");
+        Console.WriteLine($"{mutationCount} mutation(s) in transaction...");
+
+        return mutationCount;
```
amanda-tarafa, line 55: Remember that here we are not testing the feature, that's done in google-cloud-dotnet. We are testing the sample, and for that it's enough if you return `rowCount`, so the custom logger seems a little too much?

spanner/api/Spanner.Samples/LogCommitStatsAsync.cs
```diff
@@ -0,0 +1,87 @@
...
+    }
+
+    /// <summary>
+    /// Sample logger that keeps a reference to the last seen commit response.
```
amanda-tarafa, line 59: Why would this be useful. And would it be so useful that we want to show it in our only sample?

amanda-tarafa (commented):
I have a couple comments around using a custom logger.

(Sorry for the delay in review, this slip through.)

spanner/api/Spanner.Samples/LogCommitStatsAsync.cs
skuruppu: Ah this was because in other languages, the user is able to read the value of the mutation count in code. Whereas, they can't do that by default in C#. So we thought that it would be good to show the user how to read the value out from the logger if possible in case they need to make a decision based on that.

amanda-tarafa: Hmmm, still feels more like an "advance" scenerio for a second sample than the simple usage that I would show in the only sample. But I'm totally fine with leaving that decision to you.

amanda-tarafa (approved):
LGTM.

If there are unrelated failures, let me know and I can merge.

olavloite:
> LGTM.
> 
> If there are unrelated failures, let me know and I can merge.

The failures are unrelated (infrastructure failure), so please go ahead and merge.

> Hmmm, still feels more like an "advance" scenerio for a second sample than the simple usage that I would show in the only sample. But I'm totally fine with leaving that decision to you.

I've tried to indicate that users can also use the simple scenario in the comments. I think it's better to include the advanced scenario in the actual sample, as it corresponds better with the samples for the other clients (and thereby possibly also for the generic documentation).

PR #1263, opened 2021-02-19
fix: use earliest version time for CreateBackup

### LINKED ISSUE #1262: Spanner: Tests failing because of backup PITR
Sample failures [here](https://source.cloud.google.com/results/invocations/575b363e-34b8-4655-aff3-cf969b13cdd7/targets/github%2Fdotnet-docs-samples%2Fspanner%2Fapi%2FSpanner.Samples.Tests%2FTestResults/tests). Seems like the version time is "in the future".

olavloite:
Use the earliest version time of the test database instead of the local system time to ensure that the chosen version time is valid.

Fixes #1262

### DIFF
--- a/spanner/api/Spanner.Samples.Tests/CreateBackupTest.cs
+++ b/spanner/api/Spanner.Samples.Tests/CreateBackupTest.cs
@@ -12,6 +12,8 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
+using Google.Cloud.Spanner.Admin.Database.V1;
+using Google.Cloud.Spanner.Common.V1;
 using Google.Cloud.Spanner.Data;
 using Grpc.Core;
 using System;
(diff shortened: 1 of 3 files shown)


skuruppu (approved):
That's a great idea. Thank you.

thiagotnunes:
I think that the solution taken here is fair, but we have done something different in the Java Spanner tests, where we use the database `earliest version time`, instead of the server current timestamp (see https://github.com/googleapis/java-spanner/blob/master/google-cloud-spanner/src/test/java/com/google/cloud/spanner/it/ITPitrBackupAndRestore.java#L117). I would prefer if we did that unless you have concerns.

amanda-tarafa (approved):
LGTM

amanda-tarafa:
I like that we are not showing the user where we are getting the date from.

Failures are unrelated, merging...

PR #1268, opened 2021-03-01
samples(spanner): PITR samples backup fix

thiagotnunes:
Instead of using the earliest version time of the database, uses the current time (from spanner). If we used the earliest version time of the database instead we would be creating an empty backup, since the database we are backing up is not 1 hour old (default version retention period).

### DIFF
--- a/spanner/api/Spanner.Samples.Tests/CreateBackupTest.cs
+++ b/spanner/api/Spanner.Samples.Tests/CreateBackupTest.cs
@@ -1,4 +1,4 @@
-// Copyright 2021 Google Inc.
+// Copyright 2020 Google Inc.
 //
 // Licensed under the Apache License, Version 2.0 (the "License");
 // you may not use this file except in compliance with the License.
--- a/spanner/api/Spanner.Samples.Tests/CreateBackupTest.cs
+++ b/spanner/api/Spanner.Samples.Tests/CreateBackupTest.cs
@@ -1,4 +1,4 @@
-// Copyright 2020 Google Inc.
+// Copyright 2021 Google Inc.
 //
 // Licensed under the Apache License, Version 2.0 (the "License");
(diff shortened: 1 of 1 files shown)


thiagotnunes:
Test failures seem unrelated to the PR, they are failures for the `Storage.Samples.Tests` (not modified here).

spanner/api/Spanner.Samples.Tests/CreateBackupTest.cs
```diff
@@ -1,4 +1,4 @@
-// Copyright 2020 Google Inc.
+// Copyright 2021 Google Inc.
```
amanda-tarafa, line 1: nit: Please revert this change, we don't update the copyright year when we make modifications, the code is copyrighted since it was first published.

amanda-tarafa (approved):
LGTM
I'm fine with this change, but I wonder if it's really necessary. We are not showing users where we are getting the version date from so for them is irrelevant. And for testing the sample itself , it's also irrelevant if we created the backup with either date.
If you are looking at consistency across languages then that's understandable.

(We have some flakiness in this repo that we are slowly fixing, don't worry about unrelated failures.)

amanda-tarafa:
#1272 and #1274 for tracking flakiness seen on presubmits in this PR.

spanner/api/Spanner.Samples.Tests/CreateBackupTest.cs
thiagotnunes: Fixed.

thiagotnunes:
@amanda-tarafa yes, you are correct, we are looking to have all the languages consistently use the `current_timestamp()` function.

skuruppu (approved):

amanda-tarafa:
More flakiness tracked by #1277 and #1278. All failures are unrelated to this PR. Merging...

PR #1269, opened 2021-03-01
Update NuGet.Config

Youssef1313:

PR #1270, opened 2021-03-01
Don't mix private and public feeds

Youssef1313:
See [Microsoft's white paper](https://aka.ms/pkg-sec-wp) for more information about dependency confusion attack. Recreation of https://github.com/GoogleCloudPlatform/dotnet-docs-samples/pull/1269/

### DIFF
--- a/endpoints/getting-started/NuGet.Config
+++ b/endpoints/getting-started/NuGet.Config
@@ -1,8 +1,0 @@
-<?xml version="1.0" encoding="utf-8"?>
-<configuration>
-  <packageSources>
-    <!--To inherit the global NuGet package sources remove the <clear/> line below -->
-    <clear />
-    <add key="api.nuget.org" value="https://api.nuget.org/v3/index.json" />
-  </packageSources>
-</configuration>
--- a/endpoints/getting-started/NuGet.Config
+++ b/endpoints/getting-started/NuGet.Config
@@ -3,7 +3,6 @@
   <packageSources>
     <!--To inherit the global NuGet package sources remove the <clear/> line below -->
     <clear />
-    <add key="dotnet-core" value="https://www.myget.org/F/dotnet-core/api/v3/index.json" />
     <add key="api.nuget.org" value="https://api.nuget.org/v3/index.json" />
   </packageSources>
 </configuration>


Youssef1313:
@amanda-tarafa @jskeet If this caused build failures, one of the following actions will be needed:

- Use trustedSigners: requires .NET 5.0 to be specified in a global.json file.
- Create a private feed and have all needed packages mirrored to it.

It's, however, more likely that this can build successfully. Nearly all packages from dotnet-core feed are existing on NuGet.

jskeet:
It's unclear to me why https://www.myget.org/F/dotnet-core is regarded as a "private" feed. It's not private at all, is it?

I'm happy to get rid of it if we can, but I also think it's important to be clear about what's "private" and what's not.

Youssef1313:
> It's unclear to me why https://www.myget.org/F/dotnet-core is regarded as a "private" feed. It's not private at all, is it?
> 
> I'm happy to get rid of it if we can, but I also think it's important to be clear about what's "private" and what's not.

 What I mean by private is "it's a feed that not everyone can publish to". For example, Anyone can publish to nuget.org, so it's public.

The issue when these are mixed is as follows:

- Assume a package called "X" exists on myget.org/F/dotnet-core.
- Assume the same package doesn't exist on nuget.org.
- An attack uploads package called "X" to nuget.org.
- Now bad things can happen.

Limitations:

- Microsoft almost prefixes all packages with `Microsoft.*`.
- This is a reserved id in nuget.org, so an attacker can't upload package with the same name on nuget.org.
- Unfortunately, there can be exceptions. For example, Microsoft have a package called `XliffTasks` (I can't recall if it's in dotnet-core feed or another one). So a package like this was in future used, the attack can happen.

There is even a bigger issue. AFAIK, Microsoft is in the process of deleting all myget feeds and they're moving to AzDO feeds. That means an attacker could create a feed with the same name after Microsoft deletes it, and myget will allow uploading any package, regardless of reserved IDs (Reserved IDs are NuGet.org-only concept).

A small benefit (not intended by the PR) is faster build because `dotnet restore` will now only connect to one feed.

Youssef1313:
The deletion of myget feeds have actually caused a valid vulnerability in a Microsoft repository that I reported to MSRC and it was fixed in <https://github.com/microsoft/nodejstools>.

jskeet:
Right - that difference between "public" and "private" makes sense after explanation, but I'd suggest that if you're filing a lot of these pull requests, you make that clearer to start with. (It's not a very common use of the term "private" in my experience.)

I'd actually suggest that if we can, we should just delete the config file entirely. It's the only one in the repo, and if it's no longer needed to build this sample, it would be cleaner not to have it.

Youssef1313:
The existence of the file in repository ensures that the global machine-wide nuget.config (located in `%AppData%\NuGet\`) isn't used. For a clean machine (e.g, CI machine), it should only have nuget.org. But for dev machines, it can have other feeds. It's unlikely that this can cause a serious issue, but probably more safer to keep it.

jskeet:
I don't see why we'd want to keep it for *this one sample* but not for anything else (where nuget.org is basically assumed). Note that its presence makes it harder to test with local feeds.

If it's worthwhile having it at all, it should be consistent across the repository. But it's pretty clearly only present to bring in the feed that's now been removed - so let's remove the whole file instead.

amanda-tarafa:
I agree with just attempting to delete the file althogether, this is a singularity of this sample set. I'm running CI now to see if we have actual dependencies on the removed feed. If we do, we'll evaluate the alternatives.

amanda-tarafa:
@Youssef1313 We welcome contributions of course, but do please take a look  at [CONTRIBUTING.md](https://github.com/GoogleCloudPlatform/dotnet-docs-samples/blob/master/CONTRIBUTING.md) and in particular the section about [contributing a patch](https://github.com/GoogleCloudPlatform/dotnet-docs-samples/blob/master/CONTRIBUTING.md#contributing-a-patch). All the discussion happening on this PR should have happened first on an issue, and then a PR could have been submitted for review. In particular submitting a PR with a generic title and no description whatsoever is hardly going to be accepted or even reviewed in this repo and probably in most repos.
Again thanks a lot for this and future contributions.

Youssef1313:
@amanda-tarafa Agreed. Sorry for not adding details at the beginning.

---

Regarding the CI failure here, I see there are test failures - not build failures. I'm not sure if there is related to NuGet.config. Are there any known flaky tests in the repo?

amanda-tarafa:
Failures are unrelated. I've created #1271, #1272 and #1273 to address these flakes.

I've also built the quickstart locally after deleting the config file without issues.
I'll merge now.

amanda-tarafa (commented):
LGTM

Issue #1278, opened 2021-03-02
CI: On ocassion Google.Cloud* packages cannot be found.


amanda-tarafa:
CI sample [output](https://source.cloud.google.com/results/invocations/07e66c49-eaa9-4c8d-96ad-db51cbaa0bd8/targets/github%2Fdotnet-docs-samples%2Fasset%2Fquickstart%2FExportAssetsTest/tests).

I've seen it for several different libraries. It seems when it happens that packages are only being resolved from offline sources. Wonder if the machine looses connectivity from time to time. If that's the case, then there's not much we can do.

I'll add here sample output when I see the error again to try and find some commonality.

amanda-tarafa:
Another example of this with the [Transalate library](https://source.cloud.google.com/results/invocations/396e58ff-c11b-457a-9563-224671c6f12d/targets/github%2Fdotnet-docs-samples%2Ftranslate%2Fapi%2FTests/tests).

amanda-tarafa:
Another example with [Asset](https://source.cloud.google.com/results/invocations/37e8c6d0-73d0-4d84-800c-1ed57eecc6dd/targets/github%2Fdotnet-docs-samples%2Fasset%2Fquickstart%2FExportAssetsTest/tests).

amanda-tarafa:
[Asset](https://source.cloud.google.com/results/invocations/31dd34df-5b96-4977-b74b-e0df2c837eed/targets/github%2Fdotnet-docs-samples%2Fasset%2Fquickstart%2FExportAssetsTest/tests) again.

amanda-tarafa:
[Vision](https://source.cloud.google.com/results/invocations/a1fcbb86-86de-43e0-9d90-ff36536f9ac2/targets/github%2Fdotnet-docs-samples%2Fvision%2Fapi%2FVisionTest/tests) and [Asset](https://source.cloud.google.com/results/invocations/51075306-1922-43ba-a6fd-d5fcebb4d116/targets/github%2Fdotnet-docs-samples%2Fasset%2Fquickstart%2FExportAssetsTest/tests)

amanda-tarafa:
[Vision](https://source.cloud.google.com/results/invocations/b8700b79-7864-4021-8ef3-6e660e175436/targets/github%2Fdotnet-docs-samples%2Fvision%2Fapi%2FVisionTest/tests)

amanda-tarafa:
This continues to happen...

Issue #1272, opened 2021-03-01
Storage: HMAC tests not cleaning up after themselves. HMAC keys limit reached.


amanda-tarafa:
Seems like HMAC tests are not properly deleting HMAC keys and we have reached limit. All HMAC keys related tests are failing.

Sample [CI output](https://source.cloud.google.com/results/invocations/00fb02fd-8e97-40af-b459-c95f324ef2dd/targets/github%2Fdotnet-docs-samples%2Fstorage%2Fapi%2FStorage.Samples.Tests%2FTestResults/tests).

amanda-tarafa:
Actually all the tests are cleaning up after themselves. This seems to be because of sporadic failures over time in deleting the keys, and they have build up. I will clear the existing keys but keep this issue open for a while to see how soon we see it build back up.

amanda-tarafa:
I just deleted all HMAC keys. Let's give it a couple of weeks and see how soon it builds of egain.

amanda-tarafa:
And we have now [an error](https://source.cloud.google.com/results/invocations/24ac2550-b9b4-4959-b545-9a657e227787/targets/github%2Fdotnet-docs-samples%2Fstorage%2Fapi%2FStorage.Samples.Tests%2FTestResults/tests) which seems because we are trying to clean an HMAC key that was already delete.

amanda-tarafa:
We've reached limit [again](https://source.cloud.google.com/results/invocations/4d68d111-4399-4c1f-8037-daf4c2eadec9/targets/github%2Fdotnet-docs-samples%2Fstorage%2Fapi%2FStorage.Samples.Tests%2FTestResults/tests).

amanda-tarafa:
@tritone Assigning to you as well. There's some issue with the HMAC cleanup, it's there, but it might not be working always. HMAC keys are filling up quite quickly.

Assigning to you as well so that the first one of us with badnwidth can take a look. Thanks

tritone:
I did a little bit of investigation around this.

[...]

PR #1284, opened 2021-03-19
feat: add Spanner samples for customer managed encryption keys

olavloite:
(Work in progress, the tests will currently fail as the encryption keys have not yet been set up)

Adds samples for using customer managed encryption keys with Spanner.

@larkee and/or @skuruppu Access to the encryption keys (probably) also needs to be set up for this repository as well.


spanner/api/Spanner.Samples/RestoreDatabaseWithEncryptionAsync.cs
```diff
@@ -0,0 +1,63 @@
...
+            Console.WriteLine($"Error while restoring database: {completedResponse.Exception}");
+            throw completedResponse.Exception;
+        }
+
+        var database = completedResponse.Result;
+        var restoreInfo = database.RestoreInfo;
+        Console.WriteLine($"Database {restoreInfo.BackupInfo.SourceDatabase} " +
+            $"restored to {database.Name} " +
+            $"from backup {restoreInfo.BackupInfo.Backup} " +
+            $"using encryption key {kmsKeyName}");
```
skuruppu, line 59: In Node, I was able to get the key name from Database.EncryptionConfig to verify that the key that was set is the one that was being used for encryption. You'll have to do a separate DatabaseGet() request though.

olavloite: It now prints the encryption key that is set on the database.

amanda-tarafa:
Unrelated test failures being tracked:
* Storage in #1272 
* Asset in #1278

olavloite:
@skuruppu 

Permissions to create and/or use encryption keys seem to be missing for this repository:

```
  X DatabaseAndBackupWithEncryptionKeyAsyncTest.Test01_CreateDatabaseWithEncryptionKeyAsync [300ms]
  Error Message:
   Grpc.Core.RpcException : Status(StatusCode="FailedPrecondition", Detail="KMS Key provided is not usable: ServiceAccount: service-1061780273775@gcp-sa-spanner.iam.gserviceaccount.com KmsKeyName: projects/dotnet-docs-samples-tests/locations/us-central1/keyRings/spanner-test-keyring/cryptoKeys/spanner-test-key Error: generic::permission_denied: Permission 'cloudkms.cryptoKeyVersions.useToEncrypt' denied on resource 'projects/dotnet-docs-samples-tests/locations/us-central1/keyRings/spanner-test-keyring/cryptoKeys/spanner-test-key' (or it may not exist). [google.rpc.error_details_ext] { message: "Permission \'cloudkms.cryptoKeyVersions.useToEncrypt\' denied on resource \'projects/dotnet-docs-samples-tests/locations/us-central1/keyRings/spanner-test-keyring/cryptoKeys/spanner-test-key\' (or it may not exist)." }.", DebugException="Grpc.Core.Internal.CoreErrorDetailException: {"created":"@1616488607.874905326","description":"Error received from peer ipv4:173.194.217.95:443","file":"/var/local/git/grpc/src/core/lib/surface/call.cc","file_line":1062,"grpc_message":"KMS Key provided is not usable: ServiceAccount: service-1061780273775@gcp-sa-spanner.iam.gserviceaccount.com KmsKeyName: projects/dotnet-docs-samples-tests/locations/us-central1/keyRings/spanner-test-keyring/cryptoKeys/spanner-test-key Error: generic::permission_denied: Permission 'cloudkms.cryptoKeyVersions.useToEncrypt' denied on resource 'projects/dotnet-docs-samples-tests/locations/us-central1/keyRings/spanner-test-keyring/cryptoKeys/spanner-test-key' (or it may not exist). [google.rpc.error_details_ext] { message: "Permission \'cloudkms.cryptoKeyVersions.useToEncrypt\' denied on resource \'projects/dotnet-docs-samples-tests/locations/us-central1/keyRings/spanner-test-keyring/cryptoKeys/spanner-test-key\' (or it may not exist)." }.","grpc_status":9}")
```

amanda-tarafa (changes requested):
I took a look trying to figure out what was happening with the missing permissions and left a few change requests.
I'll send an internal mail about the permissions.

skuruppu:
Oh sad :( Now I remember why we decided to create a single static backup in the C# samples tests. Otherwise, the tests tend to timeout if you create a new backup in each run. So maybe it's not possible to add tests for these samples?

@amanda-tarafa what are your thoughts on this?

amanda-tarafa:
I think we can do similar, create a single static encrypted backup and use that for the restore test and even for the create backup operation and do the same that we do for the normal backup, accept that we might get an "already created" error.
That should be enough to make sure that the samples are good I think.

Are the tests in other languages being able to create backups at will without timeouts? I think there were timing issues with other languages as well. If it's just us, maybe we can look at Spanner configuration to make sure there's nothing different there.

olavloite:
> Are the tests in other languages being able to create backups at will without timeouts? I think there were timing issues with other languages as well. If it's just us, maybe we can look at Spanner configuration to make sure there's nothing different there.

Yes, but not significantly quicker than in .NET as far as I can see:
- Go: The [sample tests](https://github.com/GoogleCloudPlatform/golang-samples/pull/2001) create both a 'normal' backup and an encrypted backup at each run. Total run time is 28 minutes. The total timeout for the tests is 60 minutes.
- Java: The sample tests also create both a normal backup and an encrypted backup at each run. Total run time [for the last PR](https://source.cloud.google.com/results/invocations/ab86bd8c-4d66-4b70-b2a4-0d75f8e65399/targets) was just under 32 minutes. The total timeout for the tests is 60 minutes.

The test timeout for the .NET seems to be 30 minutes.

For both Go and Java the encryption tests for database-backup-restore are implemented as a single test that reuse the result from the preceding test step. The total run time for both Go and Java are around 11 minutes for a complete cycle of create-encrypted-database, backup-encrypted-database, restore-encrypted-database. Also, all the integration tests for Go and Java are executed in parallel, so these backup tests don't actually add 11 minutes of total build time.

amanda-tarafa:
Thanks @olavloite for the info.

- We usually have timeout per product set to no more than 10 minutes, but I think we have already made an exception for Spanner.
- Test runs are sharded, the shards run in parallel but not the individual tests in each shard, so tests for Spanner would effectively add to the overall time. This is not a problem for nightly runs, but definetely for presubmits, which already take a lot of time.
- I can work on having Spanner on it's own shard, but I don't have the bandwidth until next month. @olavloite  and @skuruppu are you both OK with working with a "static" backup for these tests, and after I've sharded out Spanner, work on properly creating backups?
- At some point after midyear, might be late 2021 or even 2022, I want to redo some of the testing infrastructure of this repo, and at that point, we probably won't have these issues, but better not count with that until there's a better ETA.

olavloite:
> * I can work on having Spanner on it's own shard, but I don't have the bandwidth until next month. @olavloite  and @skuruppu are you both OK with working with a "static" backup for these tests, and after I've sharded out Spanner, work on properly creating backups?

I would be OK with that as well, but I worry that we might even then get timeouts as these tests will effectively be added serially to the total build time:
1. Creating a database is relatively quick, so that should work.
2. Creating a backup would be skipped (in most cases), so that would normally not be a problem.
3. Restoring backups can also be very slow, so even when working with a static backup we would probably add something like 5+ minutes to the build time.

So my preference in that case would actually be to skip both creating backups and restoring them in the tests (at least for presubmits) until we have a more long-term solution.

skuruppu:
@olavloite is right. We should skip both backup and restore tests here until we have a long-term solution.

In general, we experience more timeouts due to backups in the mono repos as the tests are run more often. We tend to be fine with running the tests in the spanner-specific repos as the timeouts there are pretty rare.

skuruppu:
@amanda-tarafa we will need to merge these code samples in the next day or so. Are you ok with us skipping the backup and restore tests?

olavloite:
@amanda-tarafa The backup and restore tests for encrypted databases, including the creation of the static database and backup for these tests, are now skipped by default. The test for creating an encrypted database is still active.

The current build failure is unrelated to this change:

```
[xUnit.net 00:00:41.19]     BigQueryTest.TestExtractTable [FAIL]
  X BigQueryTest.TestExtractTable [167ms]
  Error Message:
   Google.GoogleApiException : Google.Apis.Requests.RequestError
An internal error occurred and the request could not be completed. [500]
Errors [
	Message[An internal error occurred and the request could not be completed.] Location[ - ] Reason[internalError] Domain[global]
]

  Stack Trace:
     at Google.Apis.Requests.ClientServiceRequest`1.ParseResponse(HttpResponseMessage response) in C:\Apiary\2020-03-06.14-34-34\Src\Support\Google.Apis\Requests\ClientServiceRequest.cs:line 250
   at Google.Apis.Requests.ClientServiceRequest`1.Execute() in C:\Apiary\2020-03-06.14-34-34\Src\Support\Google.Apis\Requests\ClientServiceRequest.cs:line 174
   at Google.Cloud.BigQuery.V2.BigQueryClientImpl.CreateExtractJob(TableReference tableReference, IEnumerable`1 destinationUris, CreateExtractJobOptions options) in T:\src\github\google-cloud-dotnet\releasebuild\apis\Google.Cloud.BigQuery.V2\Google.Cloud.BigQuery.V2\BigQueryClientImpl.JobCrud.cs:line 124
   at BigQueryExtractTable.ExtractTable(String projectId, String bucketName) in /tmpfs/src/github/dotnet-docs-samples/bigquery/api/Snippets/ExtractTable.cs:line 30
   at BigQueryTest.TestExtractTable() in /tmpfs/src/github/dotnet-docs-samples/bigquery/api/Snippets/Test.cs:line 144
OperationStopped: FAILED with exit code 1
```

amanda-tarafa:
On this right now, sorry for the late response.

spanner/api/Spanner.Samples.Tests/AlphabeticalTestOrderer.cs
```diff
@@ -0,0 +1,33 @@
...
+
+using System;
+using System.Collections.Generic;
+using System.Linq;
+using Xunit.Abstractions;
+using Xunit.Sdk;
+
+namespace Spanner.Samples.Tests
+{
+    public class AlphabeticalTestOrderer : ITestCaseOrderer
```
amanda-tarafa, line 23: I think you don't need this class anymore.

spanner/api/Spanner.Samples.Tests/SpannerFixture.cs
```diff
@@ -164,6 +231,75 @@ private async Task InitializeBackupAsync()
...
+    {
+        // Sample backup for encrypted restore test.
+        try
+        {
+            CreateDatabaseWithEncryptionKeyAsyncSample createDatabaseAsyncSample = new CreateDatabaseWithEncryptionKeyAsyncSample();
+            InsertDataAsyncSample insertDataAsyncSample = new InsertDataAsyncSample();
+            await createDatabaseAsyncSample.CreateDatabaseWithEncryptionKeyAsync(ProjectId, InstanceId, FixedEncryptedDatabaseId, KmsKeyName);
+            await insertDataAsyncSample.InsertDataAsync(ProjectId, InstanceId, FixedEncryptedDatabaseId);
+        }
+        catch (Exception e) when (e.ToString().Contains("Database already exists"))
```
amanda-tarafa, line 244: Could this better be:
```suggestion
        catch (RpcException e) when (e.StatusCode == StatusCode.AlreadyExists)
```

spanner/api/Spanner.Samples.Tests/CreateBackupWithEncryptionKeyAsyncTest.cs
```diff
@@ -0,0 +1,41 @@
...
+
+    public CreateBackupWithEncryptionKeyAsyncTest(SpannerFixture fixture)
+    {
+        _fixture = fixture;
+    }
+
+    [SkippableFact]
+    public async Task TestCreatBackupWithEncryptionKeyAsync()
+    {
+        Skip.If(!_fixture.RunCmekBackupSampleTests, "Spanner CMEK backup sample tests are disabled by default for performance reasons. Set the environment variable RUN_SPANNER_CMEK_BACKUP_SAMPLES_TESTS=true to enable the test.");
```
amanda-tarafa, line 35: nit: Consider adding this text here as a constant in the fixture and reusing that in all skippable tests.

amanda-tarafa (approved):
LGTM except for a couple of minor requests.
I've approved, I know these need to be merged soon, so if you don't have time to address my comments before, I'm happy to merge as is and create an issue for the rest to be addressed later.

amanda-tarafa:
Failures are internal Kokoro failures, the build finished OK.

spanner/api/Spanner.Samples.Tests/AlphabeticalTestOrderer.cs
olavloite: Nope, removed.

spanner/api/Spanner.Samples.Tests/SpannerFixture.cs
olavloite: I copied this from the other fixed database that is also created and used for this fixture, and I think the choice to check for this specific error message is intentional here, as there are three actions going on in the try block that could throw an `AlreadyExists` error:
1. The `CreateDatabase` command.
2. The creation of the tables in the database.
3. Inserting data into the tables.

Only the failure of the `CreateDatabase` operation should be handled specifically here.

Spanner does however include metadata in the trailers of the errors to indicate what kind of resource already existed or could not be found. We could use that to detect specific types of errors instead of having to rely on checking error messages. We [use this in the Java client](https://github.com/googleapis/java-spanner/blob/70c5b805def9e60775a08abd34d38c8dbd3031e0/google-cloud-spanner/src/main/java/com/google/cloud/spanner/SpannerExceptionFactory.java#L275) to detect `DatabaseNotFound` for example. I'll have a look and see if that is feasible for `AlreadyExists` as well.

spanner/api/Spanner.Samples.Tests/CreateBackupWithEncryptionKeyAsyncTest.cs
olavloite: Done

amanda-tarafa:
Let me know when you are ready for merging.

olavloite:
> Let me know when you are ready for merging.

@amanda-tarafa 
Thanks for the reminder. Yes, this is ready for merge.

Issue #1311, opened 2021-04-12
Cloud-SQL: SQL server tests are flaky.


amanda-tarafa:
I've seen [this](https://source.cloud.google.com/results/invocations/05c02d3e-9bd9-4630-ab0d-e3b384ffdcdd/targets/github%2Fdotnet-docs-samples%2Fcloud-sql%2Fsql-server%2FSqlServer/tests) a couple of times now.

amanda-tarafa:
@kurtisvg, @shubha-rajan Could you take a look, some of these are now always failing. Thanks.

kurtisvg:
Here's the error message:
```
fail: Microsoft.AspNetCore.Diagnostics.ExceptionHandlerMiddleware[1]
      An unhandled exception has occurred while executing the request.
System.Data.SqlClient.SqlException (0x80131904): A network-related or instance-specific error occurred while establishing a connection to SQL Server. The server was not found or was not accessible. Verify that the instance name is correct and that SQL Server is configured to allow remote connections. (provider: TCP Provider, error: 40 - Could not open a connection to SQL Server)
   at System.Data.SqlClient.SqlInternalConnectionTds..ctor(DbConnectionPoolIdentity identity, SqlConnectionString connectionOptions, SqlCredential credential, Object providerInfo, String newPassword, SecureString newSecurePassword, Boolean redirectedUserInstance, SqlConnectionString userConnectionOptions, SessionData reconnectSessionData, Boolean applyTransientFaultHandling)
   at System.Data.SqlClient.SqlConnectionFactory.CreateConnection(DbConnectionOptions options, DbConnectionPoolKey poolKey, Object poolGroupProviderInfo, DbConnectionPool pool, DbConnection owningConnection, DbConnectionOptions userOptions)
   at System.Data.ProviderBase.DbConnectionFactory.CreatePooledConnection(DbConnectionPool pool, DbConnection owningObject, DbConnectionOptions options, DbConnectionPoolKey poolKey, DbConnectionOptions userOptions)
   at System.Data.ProviderBase.DbConnectionPool.CreateObject(DbConnection owningObject, DbConnectionOptions userOptions, DbConnectionInternal oldConnection)
   at System.Data.ProviderBase.DbConnectionPool.UserCreateRequest(DbConnection owningObject, DbConnectionOptions userOptions, DbConnectionInternal oldConnection)
[...]

PR #1314, opened 2021-05-11
Fix PowerShell from statement to be from current location

TravisEz13:
Please use mcr.microsoft.com for the best support.  https://azure.microsoft.com/en-us/blog/microsoft-syndicates-container-catalog/

### DIFF
--- a/.kokoro/docker/Dockerfile
+++ b/.kokoro/docker/Dockerfile
@@ -1,5 +1,5 @@
 # start with a powershell image.
-FROM microsoft/powershell
+FROM mcr.microsoft.com/powershell
 
 RUN apt-get update
 RUN apt-get install wget


TravisEz13:
Change in minimal, should not require a CLA. Please accept.  We will quit publishing to the old location soon.

amanda-tarafa:
Thanks for this! I'm checking internally about the CLA bit and will update here as soon as I know more.

amanda-tarafa:
@TravisEz13 the instructions I'm getting is to close this PR with a note and do the change myself, explaining.

Would that be OK with you or do you want to reconsider signing the CLA?

TravisEz13:
CLA was signed

TravisEz13:
FYI: The example is using the latest tag which does change distribution periodically.  We moved from Ubuntu 16.04 to 20.04 as .NET dropped support for 16.04 in their latest release.  This is probably okay for your sample, but I just wanted to point it out.

amanda-tarafa:
Thanks!

I'll run CI before merging and we'll see if having the newest version works or not. If it doesn't I'll pin to whatever works until I have the time to update.

amanda-tarafa:
CI failures are unrelated to this PR and already being tracked here:

- Cloud Sql: #1311
- Storage: #1272

I'll force merge. Thanks again for this @TravisEz13

PR #1315, opened 2021-05-14
feat: add samples for JSON data type

olavloite:
Adds samples for Spanner JSON data type.

DO NOT MERGE: Depends on changes to the client library that have not yet been merged.

### DIFF
--- a/spanner/api/Spanner.Samples.Tests/UpdateDataWithJsonTest.cs
+++ b/spanner/api/Spanner.Samples.Tests/UpdateDataWithJsonTest.cs
@@ -1,0 +1,41 @@
+// Copyright 2021 Google Inc.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+using System.Threading.Tasks;
+using Xunit;
+
+[Collection(nameof(SpannerFixture))]
+public class UpdateDataWithJsonTest
+{
+    private readonly SpannerFixture _spannerFixture;
+
+    public UpdateDataWithJsonTest(SpannerFixture spannerFixture)
+    {
+        _spannerFixture = spannerFixture;
+    }
+
+    [Fact]
+    public async Task TestUpdateDataWithJson()
+    {
+        await _spannerFixture.CreateVenuesTableAndInsertDataAsync();
+
+        var addColumnSample = new AddJsonColumnAsyncSample();
+        await addColumnSample.AddJsonColumnAsync(_spannerFixture.ProjectId, _spannerFixture.InstanceId, _spannerFixture.DatabaseId);
+
+        var updateJsonSample = new UpdateDataWithJsonAsyncSample();
+        await updateJsonSample.UpdateDataWithJsonAsync(_spannerFixture.ProjectId, _spannerFixture.InstanceId, _spannerFixture.DatabaseId);
+
+        await _spannerFixture.DeleteVenuesTable();
+    }
+}
--- a/spanner/api/Spanner.Samples/AddJsonColumnAsync.cs
+++ b/spanner/api/Spanner.Samples/AddJsonColumnAsync.cs
@@ -1,0 +1,34 @@
+// Copyright 2021 Google Inc.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+// [START spanner_add_json_column]
+
+using Google.Cloud.Spanner.Data;
+using System;
+using System.Threading.Tasks;
+
+public class AddJsonColumnAsyncSample
+{
+    public async Task AddJsonColumnAsync(string projectId, string instanceId, string databaseId)
+    {
+        string connectionString = $"Data Source=projects/{projectId}/instances/{instanceId}/databases/{databaseId};Host=staging-wrenchworks.sandbox.googleapis.com";
+        string alterStatement = "ALTER TABLE Venues ADD COLUMN VenueDetails JSON";
+
+        using var connection = new SpannerConnection(connectionString);
+        using var updateCmd = connection.CreateDdlCommand(alterStatement);
+        await updateCmd.ExecuteNonQueryAsync();
+        Console.WriteLine("Added the VenueDetails column.");
+    }
+}
+// [END spanner_add_json_column]
--- a/spanner/api/Spanner.Samples/QueryDataWithJsonParameterAsync.cs
+++ b/spanner/api/Spanner.Samples/QueryDataWithJsonParameterAsync.cs
@@ -1,0 +1,61 @@
+// Copyright 2021 Google Inc.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+// [START spanner_query_with_json_parameter]
+
+using Google.Cloud.Spanner.Data;
+using System.Collections.Generic;
+using System.Threading.Tasks;
+
+public class QueryDataWithJsonParameterAsyncSample
+{
+    public class Venue
+    {
+        public int VenueId { get; set; }
+        public string VenueDetails { get; set; }
+    }
+
+    public async Task<List<Venue>> QueryDataWithJsonParameterAsync(string projectId, string instanceId, string databaseId)
+    {
+        string connectionString = $"Data Source=projects/{projectId}/instances/{instanceId}/databases/{databaseId}";
+
+        using var connection = new SpannerConnection(connectionString);
+        var jsonValue = @"{
+                            ""rating"": ""9"",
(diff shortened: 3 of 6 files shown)


spanner/api/Spanner.Samples/UpdateDataWithJsonAsync.cs
```diff
@@ -0,0 +1,123 @@
...
+                    }
+                },
+                {
+                    ""main hall"": {
+                        ""description"": ""this is the biggest space"",
+                        ""size"": ""200""
+                    }
+                }
+            ]" },
+            new Venue { VenueId = 42, VenueDetails = @"{
```
olavloite, line 60: (AFAICT this version of) .NET does not have standard support for serializing an object to JSON, which means that this JSON literal needs to be written manually as a string. Another option could be to add a dependency on an external library that supports serializing objects to JSON, such as `NewtonSoft.JSON` [(example)](https://dzone.com/articles/quickly-create-json-object). That would simplify this sample, but would at the same time also introduce an external dependency.

@amanda-tarafa @jskeet 
What are your thoughts regarding this?

jskeet: I think it's absolutely fine to add a dependency on Json.NET for a *sample* - we can change to a different dependency later on without any of the normal versioning concerns, etc. I'd be much more nervous about taking a dependency in the production code.

amanda-tarafa: Google.Cloud.Spanner.Data already has a (transitive) dependency on Newtosoft.Json so I think it's totally fine to use that here. Also, in .NET, Newtonsoft.Json is a failry standard way to deal with Json.
Users in .NET Core 3.1 or newer should know whether they want to use System.Text.Json instead, so that's fine (Maybe leave a comment "If you are using .NET Core 3.1 ...")

olavloite: Great, thanks. I've updated the samples to use `NewtonSoft` for serialization, and added a comment about using `System.Text.Json` for newer versions of .NET.

spanner/api/Spanner.Samples/AddJsonColumnAsync.cs
```diff
@@ -0,0 +1,34 @@
...
+
+using Google.Cloud.Spanner.Data;
+using System;
+using System.Threading.Tasks;
+
+public class AddJsonColumnAsyncSample
+{
+    public async Task AddJsonColumnAsync(string projectId, string instanceId, string databaseId)
+    {
+        string connectionString = $"Data Source=projects/{projectId}/instances/{instanceId}/databases/{databaseId};Host=staging-wrenchworks.sandbox.googleapis.com";
```
amanda-tarafa, line 25:
```suggestion
        string connectionString = $"Data Source=projects/{projectId}/instances/{instanceId}/databases/{databaseId}";
```

spanner/api/Spanner.Samples/QueryDataWithJsonParameterAsync.cs
```diff
@@ -0,0 +1,64 @@
...
+    {
+        string connectionString = $"Data Source=projects/{projectId}/instances/{instanceId}/databases/{databaseId}";
+
+        using var connection = new SpannerConnection(connectionString);
+        // If you are using .NET Core 3.1 or later, you can use System.Text.Json for serialization instead.
+        var jsonValue = JsonConvert.SerializeObject(new
+        {
+            rating = 9,
+            description = "This is a nice place.",
+        }, Formatting.None);
```
amanda-tarafa, line 40:
```suggestion
        });
```

amanda-tarafa (changes requested):
Mostly nits.

spanner/api/Spanner.Samples/UpdateDataWithJsonAsync.cs
```diff
@@ -0,0 +1,142 @@
...
+    {
+        string connectionString = $"Data Source=projects/{projectId}/instances/{instanceId}/databases/{databaseId}";
+        List<Venue> venues = new List<Venue>
+        {
+            // If you are using .NET Core 3.1 or later, you can use System.Text.Json for serialization instead.
+            new Venue { VenueId = 19, VenueDetails = JsonConvert.SerializeObject(new
+            {
+                rating = 9,
+                description = "This is a nice place.",
+            }, Formatting.None) },
```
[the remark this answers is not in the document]
olavloite: Yes, that looks better. Changed for all items in the list. Note that the sample data has also been changed, as the specs changed to include somewhat simpler sample data.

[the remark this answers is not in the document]
olavloite: Good point, removed.

spanner/api/Spanner.Samples/UpdateDataWithJsonAsync.cs
```diff
@@ -0,0 +1,142 @@
...
+                        {
+                            days = "Sun",
+                            open = new object[] { new { time = "8:00 AM–7:00 PM" } },
+                        }
+                    }
+                }
+            }, Formatting.None) },
+        };
+        // Create connection to Cloud Spanner.
+        using var connection = new SpannerConnection(connectionString);
```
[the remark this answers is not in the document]
olavloite: Done.

spanner/api/Spanner.Samples/UpdateDataWithJsonAsync.cs
```diff
@@ -0,0 +1,142 @@
...
+        await connection.OpenAsync();
+
+        await Task.WhenAll(venues.Select(venue =>
+        {
+            // Update rows in the Venues table.
+            using var cmd = connection.CreateUpdateCommand("Venues", new SpannerParameterCollection
```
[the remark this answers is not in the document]
olavloite: Correct. This column is specifically added for this sample, and is not used for anything else. Also, the table is only created for the specific tests that need it, and then deleted.

spanner/api/Spanner.Samples.Tests/UpdateDataWithJsonTest.cs
```diff
@@ -0,0 +1,41 @@
...
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+using System.Threading.Tasks;
+using Xunit;
+
+[Collection(nameof(SpannerFixture))]
+public class UpdateDataWithJsonTest
```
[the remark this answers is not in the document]
olavloite: Done.

spanner/api/Spanner.Samples.Tests/UpdateDataWithJsonTest.cs
```diff
@@ -0,0 +1,41 @@
...
+{
+    private readonly SpannerFixture _spannerFixture;
+
+    public UpdateDataWithJsonTest(SpannerFixture spannerFixture)
+    {
+        _spannerFixture = spannerFixture;
+    }
+
+    [Fact]
+    public async Task TestUpdateDataWithJson()
```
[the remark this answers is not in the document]
olavloite: Done.

spanner/api/Spanner.Samples.Tests/UpdateDataWithJsonTest.cs
```diff
@@ -0,0 +1,41 @@
...
+    {
+        await _spannerFixture.CreateVenuesTableAndInsertDataAsync();
```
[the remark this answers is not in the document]
olavloite: Check. It's created on demand for the tests that need it, and then deleted after the test.

spanner/api/Spanner.Samples.Tests/UpdateDataWithJsonTest.cs
```diff
@@ -0,0 +1,41 @@
...
+
+        var addColumnSample = new AddJsonColumnAsyncSample();
+        await addColumnSample.AddJsonColumnAsync(_spannerFixture.ProjectId, _spannerFixture.InstanceId, _spannerFixture.DatabaseId);
+
+        var updateJsonSample = new UpdateDataWithJsonAsyncSample();
+        await updateJsonSample.UpdateDataWithJsonAsync(_spannerFixture.ProjectId, _spannerFixture.InstanceId, _spannerFixture.DatabaseId);
```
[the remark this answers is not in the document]
olavloite: Yes, this is modelled after the existing `NUMERIC` samples. There's no output that could be reasonably validated here.

spanner/api/Spanner.Samples/AddJsonColumnAsync.cs
olavloite: Done

spanner/api/Spanner.Samples/QueryDataWithJsonParameterAsync.cs
olavloite: Done

spanner/api/Spanner.Samples/UpdateDataWithJsonAsync.cs
```diff
@@ -0,0 +1,97 @@
...
+                })
+            },
+            new Venue
+            {
+                VenueId = 4,
+                VenueDetails = JsonConvert.SerializeObject(new object[]
+                {
+                    new
+                    {
+                        name = (string) null,
```
amanda-tarafa, line 53: Can't this be changed by an actual value. It's just a very serialization specific detail and it looks ugly.

amanda-tarafa (commented):
Thanks, +1 to having simplified the data.
Do please check changes @laljikanjareeya is making on #1304 which will affect the code here.

spanner/api/Spanner.Samples/UpdateDataWithJsonAsync.cs
olavloite: This `null` value is set to be consistent with the samples for the other clients.

@zoercai Any thoughts on this? Do we want to keep this `null` in all clients, or should we consider setting an actual value?

zoercai: Yep I'm fine with having this be an actual value. I'll change the sample code in the spec so that the name is "room 1".

zoercai (approved):

amanda-tarafa:
@olavloite do ping me here when it's clear this needs to be merged and it's ready for review. I think we are waiting for a decision on https://github.com/googleapis/google-cloud-dotnet/pull/6390, the last relevant comment there seems to be to hold off on merge?

olavloite:
> @olavloite do ping me here when it's clear this needs to be merged and it's ready for review. I think we are waiting for a decision on [googleapis/google-cloud-dotnet#6390](https://github.com/googleapis/google-cloud-dotnet/pull/6390), the last relevant comment there seems to be to hold off on merge?

I'll let you know when we can proceed. This is waiting for #6390, and #6390 is waiting for the backend to be ready.

skuruppu:
@olavloite the release has been cut with JSON support so please update the PR and get it ready for review/merge.

olavloite:
@amanda-tarafa This PR is ready for review.

* I somehow can no longer remove the 'Do not merge' label from this PR.
* The current build errors are not Spanner related.

PR #1316, opened 2021-05-24
[Secret Manager] Updated code samples with etag support

gwhitehawk:


secretmanager/api/SecretManagerSample/Program.cs
```diff
@@ -42,6 +42,9 @@ class SecretOptions
 
         [Value(1, HelpText = "ID of the secret", Required = true)]
         public string SecretId { get; set; }
+
+	[Value(2, HelpText = "Etag of the secret", Required = false)]
```
sethvargo, line 46: Hey @gwhitehawk - I think there may be some formatting issues throughout. Can you run `dotnet format` or use your editor plugin to format the code?

secretmanager/api/SecretManagerSample/Program.cs
```diff
@@ -218,22 +224,24 @@ public static void CreateSecret(string projectId, string secretId)
 
         // [START secretmanager_delete_secret]
         /// <summary>
-        /// Delete an existing secret with the given name.
+        /// Delete an existing secret with the given name and etag (optional).
         /// </summary>
         /// <param name="projectId">ID of the project where the secret resides.</param>
         /// <param name="secretId">ID of the secret.</param>
+	/// <param name="etag">Current secret etag.</param>
```
sethvargo, line 231: Could we provide a quick explanation of what etags are/what they are used for?

sethvargo (changes requested):

secretmanager/api/SecretManagerSample/Program.cs
gwhitehawk: Oh oh, just used vim with lacking vimrc, fixed now.

secretmanager/api/SecretManagerSample/Program.cs
gwhitehawk: I think the comment that it should be the _current_ etag explains it quite sufficiently (as far as the code comment level of detail goes). But we have a dedicated doc that is more elaborate than what would fit on a single line here.

secretmanager/api/SecretManagerSample/Program.cs
```diff
@@ -66,6 +69,9 @@ class SecretVersionOptions
 
         [Value(2, HelpText = "Version of the secret", Required = true)]
         public string SecretVersion { get; set; }
+
+        [Value(3, HelpText = "Etag of the secret version", Required = false)]
```
sethvargo, line 73:
```suggestion
        [Value(3, HelpText = "Optional etag of the secret version", Required = false)]
```

secretmanager/api/SecretManagerSample/Program.cs
```diff
@@ -42,6 +42,9 @@ class SecretOptions
 
         [Value(1, HelpText = "ID of the secret", Required = true)]
         public string SecretId { get; set; }
+
+        [Value(2, HelpText = "Etag of the secret", Required = false)]
```
sethvargo, line 46:
```suggestion
        [Value(2, HelpText = "Optional etag of the secret", Required = false)]
```

secretmanager/api/SecretManagerSample/Program.cs
```diff
@@ -218,22 +224,24 @@ public static void CreateSecret(string projectId, string secretId)
 
         // [START secretmanager_delete_secret]
         /// <summary>
-        /// Delete an existing secret with the given name.
+        /// Delete an existing secret with the given name and etag (optional).
         /// </summary>
         /// <param name="projectId">ID of the project where the secret resides.</param>
         /// <param name="secretId">ID of the secret.</param>
+        /// <param name="etag">Current secret etag.</param>
         /// <example>
```
sethvargo, line 232: I think we should have 2 examples for each of these. The existing example as-is with no etag and then another example (copy-paste the whole `<example>` block) that is "with etags". That will make it clearer that etags aren't required.

sethvargo (commented):
Thank you for cleaning up the formatting. My concern is that, for someone unfamiliar with etags and check-and-set operations, this could be confusing. Is there some documentation we can point people towards? 

I also added some suggestions to make it clearer that etags are optional.

gwhitehawk:
Closing the PR, as I accidentally worked on an old fork without fetching remote.

PR #1320, opened 2021-06-10
samples(spanner): use an explicit transaction ...

skuruppu:
when inserting data using the mutations API so that the mutations are batched together and not executed one at a time.

### DIFF
--- a/spanner/api/Spanner.Samples/InsertDataAsync.cs
+++ b/spanner/api/Spanner.Samples/InsertDataAsync.cs
@@ -55,33 +55,39 @@ public class InsertDataAsyncSample
             new Album { SingerId = 2, AlbumId = 2, AlbumTitle = "Forever Hold your Peace" },
             new Album { SingerId = 2, AlbumId = 3, AlbumTitle = "Terrified" },
         };
+
         // Create connection to Cloud Spanner.
         using var connection = new SpannerConnection(connectionString);
         await connection.OpenAsync();
 
-        await Task.WhenAll(singers.Select(singer =>
+        await connection.RunWithRetriableTransactionAsync(async transaction =>
         {
-            // Insert rows into the Singers table.
-            using var cmd = connection.CreateInsertCommand("Singers", new SpannerParameterCollection
+            await Task.WhenAll(singers.Select(singer =>
             {
-                    { "SingerId", SpannerDbType.Int64, singer.SingerId },
-                    { "FirstName", SpannerDbType.String, singer.FirstName },
-                    { "LastName", SpannerDbType.String, singer.LastName }
-            });
-            return cmd.ExecuteNonQueryAsync();
-        }));
+                // Insert rows into the Singers table.
+                using var cmd = connection.CreateInsertCommand("Singers", new SpannerParameterCollection
+                {
+                        { "SingerId", SpannerDbType.Int64, singer.SingerId },
+                        { "FirstName", SpannerDbType.String, singer.FirstName },
+                        { "LastName", SpannerDbType.String, singer.LastName }
+                });
+                cmd.Transaction = transaction;
+                return cmd.ExecuteNonQueryAsync();
+            }));
 
-        await Task.WhenAll(albums.Select(album =>
-        {
-            // Insert rows into the Albums table.
-            using var cmd = connection.CreateInsertCommand("Albums", new SpannerParameterCollection
+            await Task.WhenAll(albums.Select(album =>
             {
-                    { "SingerId", SpannerDbType.Int64, album.SingerId },
-                    { "AlbumId", SpannerDbType.Int64, album.AlbumId },
-                    { "AlbumTitle", SpannerDbType.String,album.AlbumTitle }
-            });
-            return cmd.ExecuteNonQueryAsync();
-        }));
+                // Insert rows into the Albums table.
+                using var cmd = connection.CreateInsertCommand("Albums", new SpannerParameterCollection
+                {
+                        { "SingerId", SpannerDbType.Int64, album.SingerId },
+                        { "AlbumId", SpannerDbType.Int64, album.AlbumId },
+                        { "AlbumTitle", SpannerDbType.String,album.AlbumTitle }
+                });
+                cmd.Transaction = transaction;
+                return cmd.ExecuteNonQueryAsync();
+            }));
+        });
         Console.WriteLine("Data inserted.");
     }
 }
--- a/spanner/api/Spanner/Program.cs
+++ b/spanner/api/Spanner/Program.cs
@@ -1864,37 +1864,42 @@ namespace GoogleCloudSamples.Spanner
             {
                 await connection.OpenAsync();
 
-                // Insert rows into the Singers table.
-                var cmd = connection.CreateInsertCommand("Singers",
-                    new SpannerParameterCollection
-                    {
-                        { "SingerId", SpannerDbType.Int64 },
-                        { "FirstName", SpannerDbType.String },
-                        { "LastName", SpannerDbType.String }
-                    });
-                await Task.WhenAll(singers.Select(singer =>
+                await connection.RunWithRetriableTransactionAsync(async transaction =>
                 {
-                    cmd.Parameters["SingerId"].Value = singer.SingerId;
-                    cmd.Parameters["FirstName"].Value = singer.FirstName;
-                    cmd.Parameters["LastName"].Value = singer.LastName;
-                    return cmd.ExecuteNonQueryAsync();
-                }));
+                    // Insert rows into the Singers table.
+                    var cmd = connection.CreateInsertCommand("Singers",
+                        new SpannerParameterCollection
+                        {
+                            { "SingerId", SpannerDbType.Int64 },
+                            { "FirstName", SpannerDbType.String },
+                            { "LastName", SpannerDbType.String }
+                        });
+                    cmd.Transaction = transaction;
+                    await Task.WhenAll(singers.Select(singer =>
+                    {
+                        cmd.Parameters["SingerId"].Value = singer.SingerId;
+                        cmd.Parameters["FirstName"].Value = singer.FirstName;
+                        cmd.Parameters["LastName"].Value = singer.LastName;
+                        return cmd.ExecuteNonQueryAsync();
+                    }));
 
-                // Insert rows into the Albums table.
-                cmd = connection.CreateInsertCommand("Albums",
-                    new SpannerParameterCollection
+                    // Insert rows into the Albums table.
+                    cmd = connection.CreateInsertCommand("Albums",
+                        new SpannerParameterCollection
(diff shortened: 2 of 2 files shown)


skuruppu:
I get a test failure when I use an explicit transaction which was a bit surprising. Somehow the test logs are empty in the failed Kokoro build above. But I'm pasting the stack trace from my local run:

```
  X DeleteDataAsyncTest.TestDeleteDataAsync [1ms]
  Error Message:
   System.Transactions.TransactionAbortedException : The transaction has aborted.
---- Google.Cloud.Spanner.Data.SpannerException : Object not found.
-------- Grpc.Core.RpcException : Status(StatusCode="NotFound", Detail="Parent row for row [1,1] in table Albums is missing. Row cannot be written.", DebugException="Grpc.Core.Internal.CoreErrorDetailException: {"created":"@1623302645.713128270","description":"Error received from peer ipv6:[2404:6800:4006:80e::200a]:443","file":"/var/local/git/grpc/src/core/lib/surface/call.cc","file_line":1062,"grpc_message":"Parent row for row [1,1] in table Albums is missing. Row cannot be written.","grpc_status":5}")
  Stack Trace:
     at System.Transactions.TransactionStateAborted.EndCommit(InternalTransaction tx)
   at System.Transactions.CommittableTransaction.Commit()
   at System.Transactions.TransactionScope.InternalDispose()
   at System.Transactions.TransactionScope.Dispose()
   at InsertDataAsyncSample.InsertDataAsync(String projectId, String instanceId, String databaseId) in /google-cloud-dotnet/spanner/api/Spanner.Samples/InsertDataAsync.cs:line 92
   at SpannerFixture.InitializeDatabaseAsync() in /google-cloud-dotnet/spanner/api/Spanner.Samples.Tests/SpannerFixture.cs:line 183
   at SpannerFixture.InitializeAsync() in /google-cloud-dotnet/spanner/api/Spanner.Samples.Tests/SpannerFixture.cs:line 75
----- Inner Stack Trace -----
   at Google.Cloud.Spanner.Data.ExecuteHelper.WithErrorTranslationAndProfiling[T](Func`1 t, String name, Logger logger)
   at Google.Api.Gax.TaskExtensions.WaitWithUnwrappedExceptions(Task task)
   at Google.Api.Gax.TaskExtensions.ResultWithUnwrappedExceptions[T](Task`1 task)
   at Google.Cloud.Spanner.Data.SpannerTransaction.Commit(DateTime& timestamp)
   at Google.Cloud.Spanner.Data.SpannerTransaction.Commit()
   at Google.Cloud.Spanner.Data.VolatileResourceManager.<CommitToSpanner>b__13_0()
   at Google.Cloud.Spanner.Data.ExecuteHelper.WithErrorTranslationAndProfiling(Action t, String name, Logger logger)
   at Google.Cloud.Spanner.Data.VolatileResourceManager.CommitToSpanner()
   at Google.Cloud.Spanner.Data.VolatileResourceManager.SinglePhaseCommit(SinglePhaseEnlistment singlePhaseEnlistment)
----- Inner Stack Trace -----
   at Grpc.Gcp.GcpCallInvoker.<>c__DisplayClass27_0`2.<<AsyncUnaryCall>g__PostProcessPropagateResult|3>d.MoveNext()
--- End of stack trace from previous location where exception was thrown ---
   at Google.Api.Gax.Grpc.ApiCallRetryExtensions.<>c__DisplayClass0_0`2.<<WithRetry>b__0>d.MoveNext()
--- End of stack trace from previous location where exception was thrown ---
   at Google.Cloud.Spanner.V1.Internal.ExecuteHelper.WithSessionExpiryChecking[T](Task`1 task, Session session)
   at Google.Cloud.Spanner.V1.PooledSession.RecordSuccessAndExpiredSessions[T](Task`1 task)
   at Google.Cloud.Spanner.V1.PooledSession.CommitAsync(CommitRequest request, CallSettings callSettings)
   at Google.Cloud.Spanner.Data.SpannerTransaction.<>c__DisplayClass44_0.<<CommitAsync>b__0>d.MoveNext()
--- End of stack trace from previous location where exception was thrown ---
   at Google.Cloud.Spanner.Data.ExecuteHelper.WithErrorTranslationAndProfiling[T](Func`1 t, String name, Logger logger)
```

My code doesn't seem to be doing what I think it's doing but I'm not sure why.

spanner/api/Spanner.Samples/InsertDataAsync.cs
```diff
@@ -55,34 +56,40 @@ public async Task InsertDataAsync(string projectId, string instanceId, string da
             new Album { SingerId = 2, AlbumId = 2, AlbumTitle = "Forever Hold your Peace" },
             new Album { SingerId = 2, AlbumId = 3, AlbumTitle = "Terrified" },
         };
-        // Create connection to Cloud Spanner.
-        using var connection = new SpannerConnection(connectionString);
-        await connection.OpenAsync();
 
-        await Task.WhenAll(singers.Select(singer =>
+        using (TransactionScope scope = new TransactionScope(
```
amanda-tarafa, line 60: Just for the sake of figuring out what the error is, can you do a normal begin transaction from the connection, and set that transaction manually on all the commands? If that works, then the problem is in the ambient transaction mechanism.

amanda-tarafa: Ah, no, scratch that, the error is because when you get to execute the mutations that insert on the Album, the ones that inserted on the Singers table are not yet "effective".
So we are keeping the mutations in the order that we are receiving them, and in this code I'm pretty certain that all the Singers mutations will be received by the transaction before all the Albums mutations.
Can it be that in the backend Spanner does not necessarily respects that order?

(It was not failing before because each command was being individually executed).

I will try and debug this on Monday.

olavloite: I looked into this, and the problem is a race condition that occurs in the `VolatileResourceManager` implementation in the Spanner client, although I don't know whether that should be considered a bug in the `VolatileResourceManager` or wrong usage in this case (although my feeling says that this is a bug). I can reproduce this locally using a mock server and see exactly which requests and mutations are actually sent to Spanner in this use case.

What happens is the following:
1. The insertions of the `Singer` records are executed in parallel by [this call](https://github.com/GoogleCloudPlatform/dotnet-docs-samples/blob/a6a3332cf32010bff2b3e281c6031e7d00ce0bca/spanner/api/Spanner.Samples/InsertDataAsync.cs#L67) to `Task.WhenAll`.
2. Each of these calls will get to [this point](https://github.com/googleapis/google-cloud-dotnet/blob/00dcfae3f83ab591dcc5d372c1d5dc17594d7653/apis/Google.Cloud.Spanner.Data/Google.Cloud.Spanner.Data/VolatileResourceManager.cs#L203) in `VolatileResourceManager` and see a `null` value for `_transaction`. Each of the calls will therefore also call `_spannerConnection.BeginTransactionAsync` and assign the returned value to `_transaction`. 
3. The last of the calls in step 2 to finish will eventually assign the returned transaction to `_transaction`, and that is the `SpannerTransaction` that will be used for the remainder of the ambient transaction. This will normally be the last mutation, i.e. the Singer with id 5.
4. The insertions of the Albums will use the transaction that was created by the last Singer mutation, and this will fail as it will only contain Singer 5 (and not Singer 1).
5. The insertions of Singers 1 to 4 will be added to other transactions that are never committed. The sessions that are used by these transactions are (AFAICT) leaked from the pool.

Assuming that executing multiple commands in parallel on the same (ambient) transaction is a valid use case, then we need some synchronization in `VolatileResourceManager` that will ensure that only one `SpannerTransaction` is ever created and assigned to it, and that it can only be committed once (and probably a couple of other things that I haven't noticed yet).
If executing multiple commands in parallel on the same ambient transaction is not a valid use case, then we should still have a closer look at this, as I think this example will cause 4 sessions to be leaked from the pool (the 4 sessions that are associated with the transactions that are created by the first 4 Singer records).

amanda-tarafa:
I've added the do not merge label until we decide what to do on our internal thread.

skuruppu:
Phew, I can confirm that the tests now pass after using a `RetriableTransaction`. The failures in CI are unrelated to Spanner.

skuruppu:
@jskeet if you're comfortable with approving and merging this in, please go ahead. Otherwise, I'm more than happy to wait till @amanda-tarafa is back.

spanner/api/Spanner.Samples/InsertDataAsync.cs
skuruppu: Based on the internal discussion, I'm now using a `RetriableTransaction` so this is no longer an issue. This is also good in the sense that now we finally have a code sample that shows how to use a RetriableTransaction. Before, the only examples were in the snippets and tests in the main repo.

spanner/api/Spanner.Samples/InsertDataAsync.cs
```diff
@@ -55,34 +55,40 @@ public async Task InsertDataAsync(string projectId, string instanceId, string da
...
+                using var cmd = connection.CreateInsertCommand("Albums", new SpannerParameterCollection
+                {
+                        { "SingerId", SpannerDbType.Int64, album.SingerId },
+                        { "AlbumId", SpannerDbType.Int64, album.AlbumId },
+                        { "AlbumTitle", SpannerDbType.String,album.AlbumTitle }
+                });
+                cmd.Transaction = transaction;
+                return cmd.ExecuteNonQueryAsync();
+            }));
+            Console.WriteLine("Data inserted.");
```
olavloite, line 90: nit: I think it would be better to place this line after the transaction instead of inside the transaction. Any error that is returned by the `Commit` will now be preceded by the `Data inserted` log line (i.e. at this point the data is only buffered and not really inserted).

spanner/api/Spanner/Program.cs
```diff
@@ -1864,38 +1864,43 @@ await connection.BeginTransactionAsync())
...
+                        });
+                    cmd.Transaction = transaction;
+                    await Task.WhenAll(albums.Select(album =>
+                    {
+                        cmd.Parameters["SingerId"].Value = album.SingerId;
+                        cmd.Parameters["AlbumId"].Value = album.AlbumId;
+                        cmd.Parameters["AlbumTitle"].Value = album.AlbumTitle;
+                        return cmd.ExecuteNonQueryAsync();
+                    }));
+                    Console.WriteLine("Inserted data.");
```
olavloite, line 1902: Same as above, move outside of transaction.

olavloite (approved):
Small nit on where the `Console.WriteLine` is placed at the moment, otherwise LGTM.

spanner/api/Spanner.Samples/InsertDataAsync.cs
skuruppu: Great catch, thank you. I fixed it in both places.

amanda-tarafa (approved):
LGTM

PR #1325, opened 2021-06-18
CloudAsset: Add sample code for ListAssets

peter-zheng-g:


asset/api/Asset.Samples.Tests/ListAssetsTest.cs
```diff
@@ -0,0 +1,41 @@
+/*
+ * Copyright (c) 2020 Google LLC.
```
amanda-tarafa, line 2: Nit: 2021

asset/api/Asset.Samples.Tests/ListAssetsTest.cs
```diff
@@ -0,0 +1,41 @@
...
+    private readonly AssetFixture _fixture;
+    private readonly ListAssetsSample _sample;
+
+    public ListAssetsTest(AssetFixture fixture)
+    {
+        _fixture = fixture;
+        _sample = new ListAssetsSample();
+    }
+
+    [Fact(Skip = "https://github.com/GoogleCloudPlatform/dotnet-docs-samples/issues/1141")]
```
amanda-tarafa, line 32: Does this one fail as well?

asset/api/Asset.Samples.Tests/ListAssetsTest.cs
```diff
@@ -0,0 +1,41 @@
...
+    public void TestListAssets()
+    {
+        // Run the sample code.
+        var result = _sample.ListAssets(_fixture.ProjectId);
+
+        string assetName = String.Format("//storage.googleapis.com/{0}", _fixture.BucketName);
+        Assert.Contains(assetName, result.ToString());
```
amanda-tarafa, line 39: This will fail if the asset is not on the first page of results. You need to use Assert.Contains

amanda-tarafa (changes requested):

asset/api/Asset.Samples.Tests/ListAssetsTest.cs
peter-zheng-g: Yes, it'll be as flaky as the other test case. The issue is, for any change with the test project (such as a gcs bucket creation), it takes time (typically seconds) to propagate to CloudAsset backend. It'll be much less flaky if we can add a sleep/wait and some loops here.

asset/api/Asset.Samples.Tests/ListAssetsTest.cs
peter-zheng-g: I've added an additional parameter assetTypes to restrict the result to GCS bucket only, I guess this should avoid the issue?

asset/api/Asset.Samples.Tests/ListAssetsTest.cs
peter-zheng-g: I've removed the annotation and see how it goes.

asset/api/Asset.Samples.Tests/ListAssetsTest.cs
amanda-tarafa: I know that some of the other Asset tests are checking against the string representation of the responses and I've flagged that on review before, it's really not accurate. You should navigate the response until you find what you are looking for, something like this:

```
Assert.Contains(result, asset => asset.Name == assetName);
```
In this way you guarante that results are fetched until we find the one we are looking for, and also, that if by chance, there's another property for some other asset containing the value `assetName` we don't get a false positive.

amanda-tarafa (changes requested):

asset/api/Asset.Samples.Tests/ListAssetsTest.cs
peter-zheng-g: Thanks. I was not familiar with C#. Glad that C# has this way to get what we need.

amanda-tarafa (approved):
LGTM. Thanks

PR #1326, opened 2021-06-19
feat(Compute): Adds usage report export bucket samples.

amanda-tarafa:


compute/api/Compute.Samples.Tests/DisableUsageExportBucketAsyncTest.cs
```diff
@@ -0,0 +1,57 @@
...
+
+            await _sample.DisableUsageExportBucketAsync(_fixture.ProjectId);
+
+            project = await _fixture.ProjectsClient.GetAsync(_fixture.ProjectId);
+            var newUsageExportLocation = project.UsageExportLocation;
+
+            Assert.Null(newUsageExportLocation);
+
+            // As cleanup we just restore whatever config we had before.
+            // Strictly speaking we should use a separate project for this test, etc. But let's leave this
```
rsamborski, line 45: I agree. I've run into race condition while multiple tests were run by Kokoro on the same project (for different PHP versions simultaneously)

compute/api/Compute.Samples/SetUsageExportBucketAsync.cs
```diff
@@ -0,0 +1,54 @@
...
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+// [START compute_usage_report_set]
+// [START compute_instances_verify_default_value]
```
rsamborski, line 18: Is it OK to have the same region tag in two different files? We want to include SetUsageExportBucket and GetUsageExportBucket in one sample (compute_instances_verify_default_value), but I am not sure if this will work correctly. Can you confirm?

rsamborski (commented):
LGTM, just two minor comments/questions.

compute/api/Compute.Samples.Tests/DisableUsageExportBucketAsyncTest.cs
amanda-tarafa: I deactivated parallel test running, but still it can happen when the tests are run on Windows and Linux at roughly the same time. I'll leave it as is for now and if it's too flaky I'll modify it.

compute/api/Compute.Samples/SetUsageExportBucketAsync.cs
amanda-tarafa: Hmm now that you mention it I'm not certain if the code can be picked up from two different files. Would it be terrible to have two distinct region tags, one for checking input and another for checking output? Our general policy is to not have code for different samples on the same file.

rsamborski: It seems we can include region_tags from different files, but it will look a bit different. We are now discussing if we actually need the tag 'compute_instances_verify_default_value' as we may decide to use only 'compute_usage_report_set' and 'compute_usage_report_get' - this way there will be no such problem. I should have some decisions tomorrow and will let you know.

rsamborski: @amanda-tarafa We decided to refactor the snippets and get rid of the ‘compute_instances_verify_default_value’ region tag. Instead we will use ‘compute_usage_report_set’ and ‘compute_usage_report_get’ only. So the only thing for you to do is to remove the unnecessary tag and you should be OK to merge your code.

compute/api/Compute.Samples/OperationPoller.cs
```diff
@@ -22,6 +22,20 @@
 
 public static class OperationPoller
 {
+    public static async Task<Operation> PollUntilCompletedAsync(this Operation operation, string projectId)
+    {
+        GlobalOperationsClient operationClient = await GlobalOperationsClient.CreateAsync();
```
jskeet, line 27: Maybe rename to operationsClient to be consistent with the type name?

compute/api/Compute.Samples/SetUsageExportBucketAsync.cs
jskeet: Looks like this is still pending?

jskeet (approved):
Just one nit - otherwise looks okay.

amanda-tarafa:
Failures are unrelated and being tracked in #1311 

Force merging.

PR #1328, opened 2021-06-22
Update Functions to GA libraries

jskeet:


jskeet:
Test failures are all due to Spanner.

@amanda-tarafa is on vacation, and I'm a SME for both language and API, so I'll self-merge.

PR #1329, opened 2021-06-22
Update Eventarc sample to use latest libraries

jskeet:
Both CloudNative.CloudEvents and Google.Events.Protobuf have
changed significantly since this sample was originally written.

The `gcloud eventarc` command is now GA, but creating a trigger with
a `matching-criteria` parameter still requires the `beta` component.

### DIFF
--- a/eventarc/pubsub/README.md
+++ b/eventarc/pubsub/README.md
@@ -16,9 +16,9 @@ gcloud run deploy eventarc-pubsub \
 Create a Cloud Pub/Sub trigger:
 
 ```sh
-gcloud beta eventarc triggers create pubsub-trigger \
+gcloud eventarc triggers create pubsub-trigger \
   --destination-run-service eventarc-pubsub \
-  --matching-criteria "type=google.cloud.pubsub.topic.v1.messagePublished"
+  --event-filters "type=google.cloud.pubsub.topic.v1.messagePublished"
 ```
 
 ## Test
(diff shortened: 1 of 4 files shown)


eventarc/pubsub/README.md
```diff
@@ -16,7 +16,7 @@ gcloud run deploy eventarc-pubsub \
 Create a Cloud Pub/Sub trigger:
 
 ```sh
-gcloud beta events triggers create pubsub-trigger \
+gcloud beta eventarc triggers create pubsub-trigger \
```
meteatamel, line 19: We can also remove the 'beta' now

jskeet: I don't think so - the filtering part is still beta, as far as I can tell.

meteatamel: `gcloud eventarc triggers create` is definetely not beta: https://cloud.google.com/eventarc/docs/creating-triggers

jskeet: But the matching-criteria flag is. Put it this way: I tried it this morning without beta, and it failed. With beta, it works.

Perhaps it should use a different flag instead of matching-criteria?

meteatamel: Ah, yes, `matching-criteria` was renamed to `event-filters` in GA. I added this README before we had Eventarc docs. Now that we have Eventarc docs, maybe it's best to remove it? I doubt we'll be able to keep it up to date with all the changes, as evidenced here.

jskeet:
Updated - PTAL.

meteatamel (approved):

PR #1330, opened 2021-06-25
samples(spanner): create instance with processing units

skuruppu:


spanner/api/Spanner.Samples.Tests/CreateInstanceWithProcessingUnitsTest.cs
```diff
@@ -0,0 +1,35 @@
...
+    public CreateInstanceWithProcessingUnitsTest(SpannerFixture spannerFixture)
+    {
+        _spannerFixture = spannerFixture;
+    }
+
+    [Fact]
+    public void TestCreateInstanceWithProcessingUnits()
+    {
+        var createInstanceWithProcessingUnitsSample = new CreateInstanceWithProcessingUnitsSample();
+        var instance = createInstanceWithProcessingUnitsSample.CreateInstance(_spannerFixture.ProjectId, _spannerFixture.InstanceId));
```
amanda-tarafa, line 32: This InstanceId was already present on the fixture so I guess it is being used for something. It's probably the instance that has already been created and that is used for everything else. If that's the case this test will always fail.

amanda-tarafa (commented):
The comment I left plus this does not build?

```
CreateInstanceWithProcessingUnits.cs(36,13): error CS0117: 'Instance' does not contain a definition for 'ProcessingUnits' [/tmpfs/src/github/dotnet-docs-samples/spanner/api/Spanner.Samples/Spanner.Samples.csproj] FAILED with exit code 1 BUILD FAILED
```

spanner/api/Spanner.Samples.Tests/CreateInstanceWithProcessingUnitsTest.cs
skuruppu: Good catch :) Thanks.

skuruppu:
The failing tests are unrelated to Spanner AFAICT.

spanner/api/Spanner.Samples.Tests/SpannerFixture.cs
```diff
@@ -49,6 +49,9 @@ public class SpannerFixture : IAsyncLifetime, ICollectionFixture<SpannerFixture>
     public string FixedEncryptedDatabaseId { get; } = "fixed-enc-backup-db";
     public string FixedEncryptedBackupId { get; } = "fixed-enc-backup";
 
+    public string InstanceIdWithProcessingUnits { get; } = $"my-instance-processing-units-{DateTimeOffset.UtcNow.ToUnixTimeMilliseconds()}";
```
amanda-tarafa, line 52: On Dispose, I think you should attemp to delete this instance, inside a try catch so that we don't fail the tests if we can't delete it.

spanner/api/Spanner.Samples/CreateInstanceWithProcessingUnits.cs
```diff
@@ -0,0 +1,77 @@
...
+
+        if (completedResponse.IsFaulted)
+        {
+            Console.WriteLine($"Error while creating instance: {completedResponse.Exception}");
+            throw completedResponse.Exception;
+        }
+
+        Console.WriteLine($"Instance created successfully.");
+
+        // GetInstance to get the processing units using a field mask.
```
amanda-tarafa, line 63: We usually would have two separate samples for this, one for creating the instance, and one for obtaining the processing units. It's fine if docs have been designed with this as a single sample, just saying that it is going to be slightly different to the rest of the samples in that sense.

amanda-tarafa (changes requested):
One request to attempt cleanup of the created instance.

spanner/api/Spanner.Samples.Tests/SpannerFixture.cs
skuruppu: Ah I was thinking about this and I got somewhat confused by where to do the delete because all other resources are deleted when they're "stale".

Added a method to delete the instance on dispose.

spanner/api/Spanner.Samples/CreateInstanceWithProcessingUnits.cs
skuruppu: Yeah I totally understand. We unfortunately will only have one sample for this feature so we decided to put both the create and get in one place :(

skuruppu:
Thanks for the review feedback. PTAL @amanda-tarafa.

amanda-tarafa (approved):
LGTM

amanda-tarafa:
Failures are unrelated and being tracked here: #1311.

I'll force merge.

PR #1333, opened 2021-07-08
chore: implementing TW's feedback in the comments in Compute Samples

rsamborski:


compute/api/Compute.Samples/CreateInstanceAsync.cs
```diff
@@ -22,6 +22,7 @@
 public class CreateInstanceAsyncSample
 {
     public async Task CreateInstanceAsync(
+        // TODO(developer): Replace these variables before running the sample.
```
amanda-tarafa, line 25: This text is not C# idiomatic nor it matches what this code is. It should read something like (Here and everywhere):
```suggestion
        // TODO(developer): Set your own default values for these parameters or pass different values when calling this method.
```

compute/api/Compute.Samples/DeleteInstanceAsync.cs
```diff
@@ -23,13 +23,16 @@
...
         string zone = "us-central1-a",
         string machineName = "test-machine")
     {
 
-        // Initialize the client that will be used to send instance-related requests.
-        // You should reuse the same client for multiple requests.
+        // Initialize client that will be used to send requests. This client only needs to be created
+        // once, and can be reused for multiple requests. After completing all of your requests, call
+        // the `InstancesClient.close()` method on the client to safely
+        // clean up any remaining background resources.
```
amanda-tarafa, line 35: In C#, it wouldn't be IntancesClient.close(), it would be client.Dispose(), but none of these client classes are disposable (probably because they don't need to be) so, this part of the comment is not accurate. Here and everywhere.

```suggestion
        // Initialize client that will be used to send requests. This client only needs to be created
        // once, and can be reused for multiple requests.
```

compute/api/Compute.Samples/GetUsageExportBucketAsync.cs
```diff
@@ -37,8 +39,8 @@ public async Task<Project> GetUsageExportBucketAsync(string projectId = "your-pr
         else if (string.IsNullOrEmpty(project.UsageExportLocation.ReportNamePrefix))
         {
             Console.WriteLine(
-                $"A null or empty value for {nameof(UsageExportLocation.ReportNamePrefix)} " +
-                $"will result in reports being prefixed with `usage_gce`.");
+                $"Setting {nameof(UsageExportLocation.ReportNamePrefix)} " +
+                $"to empty value causes the report to have the default prefix of `usage_gce`.");
```
amanda-tarafa, line 43: null and empty are two different values in C#, they have the same effect on the prefix. Here and everywhere.
```suggestion
                $"to null or empty values causes the report to have the default prefix of `usage_gce`.");
```

amanda-tarafa (changes requested):
Some of the wording here is either not relevatn for C# or incorrectly describes the code.

@jskeet for a second pair of eyes.

compute/api/Compute.Samples/GetUsageExportBucketAsync.cs
jskeet: Possibly "to null or an empty string" - but generally, yes.

jskeet:
I agree with all Amanda's proposals.

rsamborski:
Updated per your recommendations.

amanda-tarafa (approved):
LGTM

Will wait on tests to go green and merge after.

amanda-tarafa:
Failures are unrelated and being tracked:

- #1311 for SQL
- #1272 for Storage

Force merging.

PR #1336, opened 2021-07-13
Add SQL Server sample connection string setting

jsimonweb:
Addressing issue #1311

### DIFF
--- a/cloud-sql/sql-server/Startup.cs
+++ b/cloud-sql/sql-server/Startup.cs
@@ -151,8 +151,8 @@ namespace CloudSql
             // 4-8 minutes, or if the pooler detects that the
             // connection with the server no longer exists.
             // [END cloud_sql_sqlserver_dotnet_ado_lifetime]
-            // [END_EXCLUDE]
             connectionString.TrustServerCertificate = true;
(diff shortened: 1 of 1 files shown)


cloud-sql/sql-server/Startup.cs
```diff
@@ -152,6 +152,7 @@ public static SqlConnectionStringBuilder GetSqlServerConnectionString()
             // connection with the server no longer exists.
             // [END cloud_sql_sqlserver_dotnet_ado_lifetime]
             // [END_EXCLUDE]
+            connectionString.TrustServerCertificate = true;
```
amanda-tarafa, line 155: I'm a little confused by all the tags here, just checking that this won't appear on samples? I don't think we should be recommending this setting by default.

amanda-tarafa (commented):
LGTM, just one question about it.

cloud-sql/sql-server/Startup.cs
jsimonweb: Updated as suggested to exclude the additional setting so that it doesn't appear in documentation code snippets.

amanda-tarafa (approved):
LGTM

PR #1337, opened 2021-07-14
test: retry CreateInstance test on Unavailable

### LINKED ISSUE #1332: Spanner: CreateInstanceTest is flaky
```
[xUnit.net 00:04:01.76]     CreateInstanceTest.TestCreateInstance [FAIL]
Waiting for the operation to finish.
  X CreateInstanceTest.TestCreateInstance [8ms]
  Error Message:
   Assert.Equal() Failure
Expected: AlreadyExists
Actual:   Unavailable
  Stack Trace:
     at CreateInstanceTest.TestCreateInstance() in /tmpfs/src/github/dotnet-docs-samples/spanner/api/Spanner.Samples.Tests/CreateInstanceTest.cs:line 34
```

olavloite:
The `CreateInstance` test should be retried if the error that is returned is `Unavailable`. `Unavailable` can be returned by any RPC, and should normally be retried automatically for idem-potent RPCs. `CreateInstance` is not idem-potent, and is therefore not retried automatically. In this test however we can safely retry it without checking the status of the backend, as the test expects an `AlreadyExists` error to be returned.

Fixes #1332

### DIFF
--- a/spanner/api/Spanner.Samples.Tests/CreateInstanceTest.cs
+++ b/spanner/api/Spanner.Samples.Tests/CreateInstanceTest.cs
@@ -12,12 +12,18 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
+using System;
+using System.Linq;
 using Grpc.Core;
+using System.Threading;
+using Google.Rpc;
 using Xunit;
 
 [Collection(nameof(SpannerFixture))]
 public class CreateInstanceTest
 {
+    private const int MaxAttempts = 10;
+    private static readonly string s_retryInfoMetadataKey = RetryInfo.Descriptor.FullName + "-bin";
     private readonly SpannerFixture _spannerFixture;
 
     public CreateInstanceTest(SpannerFixture spannerFixture)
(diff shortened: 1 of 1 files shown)


spanner/api/Spanner.Samples.Tests/CreateInstanceTest.cs
```diff
@@ -29,8 +35,38 @@ public CreateInstanceTest(SpannerFixture spannerFixture)
...
+            // Retry the test if we get a temporary Unavailable error.
+            if (StatusCode.Unavailable == exception.StatusCode && attempt < MaxAttempts)
+            {
+                var delay = ExtractRetryDelay(exception);
+                Thread.Sleep(delay.Milliseconds);
+                continue;
+            }
+            break;
+        }
         Assert.Equal(StatusCode.AlreadyExists, exception.StatusCode);
```
amanda-tarafa, line 54: It's fine as it is, but maybe this is cleaner? Just a suggestions though.
```suggestion
        // Instance already exists since it was created in the test setup so it should throw an exception.
        RpcException Attempt() => Assert.Throws<RpcException>(() => createInstanceSample.CreateInstance(_spannerFixture.ProjectId, _spannerFixture.InstanceId));
        
        for (int attempt = 0, RpcException ex = Attempt(); attempt < 10 && ex.StatusCode == StatusCode.Unavailable; attempt++, ex = Attempt())
        {
            var delay = ExtractRetryDelay(exception);
            Thread.Sleep(delay.Milliseconds);
        }
        Assert.Equal(StatusCode.AlreadyExists, exception.StatusCode);
```

amanda-tarafa (approved):
LGTM

skuruppu (approved):

spanner/api/Spanner.Samples.Tests/CreateInstanceTest.cs
olavloite: It's certainly more compact, but I personally find it harder to read and quickly understand what it does.

amanda-tarafa:
Failures are unrelated and being tracked in #1272

PR #1339, opened 2021-07-19
refactor(Storage): Retries HMAC key cleanup.

### LINKED ISSUE #1272: Storage: HMAC tests not cleaning up after themselves. HMAC keys limit reached.

amanda-tarafa:
And moves the cleanup to individual Dispose methods so that tests themselves are not delayed.
Note: HMAC keys need to be deleted inmediately because only 5 are allowed per service account.
Fixes #1272

### DIFF
--- a/storage/api/Storage.Samples.Tests/ActivateHmacKeyTest.cs
+++ b/storage/api/Storage.Samples.Tests/ActivateHmacKeyTest.cs
@@ -12,30 +12,13 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
-using System;
 using Xunit;
 
 [Collection(nameof(BucketFixture))]
-public class ActivateHmacKeyTest : IDisposable
+public class ActivateHmacKeyTest : HmacKeyManager
 {
-    private readonly BucketFixture _bucketFixture;
-
-    private string _accessId;
-    private bool _isActive = true;
-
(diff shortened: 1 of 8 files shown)


jskeet (approved):
It's okay, but given that we've got the same code in multiple tests, I wonder whether a common base class might be worth considering.

tritone (approved):
Thanks for the fix!

amanda-tarafa:
> It's okay, but given that we've got the same code in multiple tests, I wonder whether a common base class might be worth considering.

Yep, makes sense, will push that shortly.

amanda-tarafa:
@jskeet Added a second commit with a common class in case you want to take another look. Else, I'll wait on green before merging. Thanks!

jskeet:
Taking a quick look now.

jskeet (approved):
Great, thanks :)

amanda-tarafa:
Failures are unrelated and being tracked in:
- #1340

PR #1345, opened 2021-07-21
spanner: cleanup stale instances before starting new test cycle

### LINKED ISSUE #1341: tests: delete stale instances
We introduced a code sample for creating an instance with [processing units](https://github.com/GoogleCloudPlatform/dotnet-docs-samples/commit/ad91ad6e3249620b038b83ae7d1faa74ac66e08e#diff-3451c37ceaa172ce91470fb9900bc234ae60eb25739b1ec0a6059d33fe9eb8fb), which also included cleaning up the instance at the end of the test. But if the tests fail, then the instance cleanup code may not get triggered. So it would be nice to also have a method that deletes stale instances periodically, just as we do for databases.

laljikanjareeya:
Fixes #1341


skuruppu (approved):

amanda-tarafa (approved):
LGTM

PR #1346, opened 2021-07-22
tests: Delete backups before deleting database.

amanda-tarafa:
Else, databases won't be deleted with active backups (it seems).


amanda-tarafa:
Self merging as this is a trivial change and tests going green means that it acomplished its goal. I've also checked on the GCP project and databases have been deleted.

PR #1350, opened 2021-07-28
[Spanner]: samples for configurable leader placement

laljikanjareeya:
Still, it's in progress.
The samples are done few Tests are Pending.
//cc
@skuruppu

PR #1366, opened 2021-07-30
samples: samples for configurable leader options

olavloite:
Adds samples and tests for the new configurable leader options in Cloud Spanner. Also adds a number of samples for existing features that were not yet covered by samples.


laljikanjareeya:
@googlebot I consent.

spanner/api/Spanner.Samples.Tests/GetDatabaseDdlAsyncTest.cs
```diff
@@ -0,0 +1,58 @@
...
+                     ComposerInfo BYTES(MAX),
+                 ) PRIMARY KEY (SingerId)";
+        var createAlbumsTable =
+            @"CREATE TABLE Albums (
+                     SingerId INT64 NOT NULL,
+                     AlbumId INT64 NOT NULL,
+                     AlbumTitle STRING(MAX),
+                 ) PRIMARY KEY (SingerId, AlbumId),
+                 INTERLEAVE IN PARENT Singers ON DELETE CASCADE";
+        await _spannerFixture.RunWithTemporaryDatabaseAsync(_spannerFixture.InstanceId, async databaseId =>
```
olavloite, line 45: Note: This creates a temporary database, executes the test function and then deletes the test database. That might seem inefficient, but creating and dropping (empty) databases is relatively quick on Cloud Spanner nowadays. The execution time for only this test when measured from my local development machine is between 5-7 seconds excluding the time spent setting up the `SpannerFixture`. The total execution time for the test, including the initialisation that is done in `SpannerFixture`, is however around 4 minutes when it is executed multiple times (that is; the tables, backups, etc. have already been created by previous runs).
For this type of test that does not depend on any large data model or previously inserted data, it would be good to have a possibility of skipping most of the initialisation in the `SpannerFixture` in order to speed up the development of new samples and tests.

olavloite:
(The build errors are not Spanner related)

spanner/api/Spanner.Samples.Tests/GetDatabaseDdlAsyncTest.cs
```diff
@@ -0,0 +1,58 @@
...
+        {
+            var sample = new GetDatabaseDdlAsyncSample();
+            var statements =
+                await sample.GetDatabaseDdlAsync(_spannerFixture.ProjectId, _spannerFixture.InstanceId, databaseId);
+            Assert.Collection(statements,
+                // Only check the start of the statement, as there is no guarantee on exactly
+                // how Cloud Spanner will format the returned SQL string.
+                statement => Assert.StartsWith("CREATE TABLE Singers", statement),
+                statement => Assert.StartsWith("CREATE TABLE Albums", statement)
```
skuruppu, line 54: Could we also check if the default leader extra statment is there? Even though it's not explicitly stated in the spec, I feel like that's the intention of this sample. But I see that it might be tricky here given that we create a new database for every sample. Whereas in other languages, because we execute the samples in a certain order, we'd create a database with the extra statement for specifying the leader and then dump the DDL.

So you might just be able to call `CreateDatabaseWithDefaultLeaderAsyncSample()` to create the database instead of `RunWithTemporaryDatabaseAsync()`.

skuruppu (approved):

spanner/api/Spanner.Samples.Tests/GetDatabaseDdlAsyncTest.cs
olavloite: Good point. I've added an `ALTER DATABASE` statement to the temporary database definition and added asserts for it in the test.

amanda-tarafa (changes requested):
Nothing major. A few questions though:

- What's the relation betwee this PR and #1350? Is this one a precursor? Or can we close #1350?
- Mostly for @skuruppu : several of the samples here, ListDatabases for instance, seem to not be exclusively related to `leader`, for instance, their region tags have no `leader` in the name. The only inmediate relation to `leader` is that is what they print on the console. My main question is, if there's more database metadata that we'll want to print , we'll we use these same samples, or will we have new ones? If we are going to have different samples for different metadata aspects, then these here should have leader on the class name, etc. but most importantly on the region tag.

olavloite:
> * What's the relation betwee this PR and [[Spanner]: samples for configurable leader placement #1350](https://github.com/GoogleCloudPlatform/dotnet-docs-samples/pull/1350)? Is this one a precursor? Or can we close [[Spanner]: samples for configurable leader placement #1350](https://github.com/GoogleCloudPlatform/dotnet-docs-samples/pull/1350)?

This PR includes all commits from #1350, and #1350 can be closed (I actually thought it had been closed already).

spanner/api/Spanner.Samples.Tests/GetDatabaseDdlAsyncTest.cs
olavloite: We could call `CreateDatabaseWithDefaultLeaderAsyncSample()`, but I would prefer not to for the following two reasons:
1. That sample is specifically for creating a database with a default leader. The fact that it also creates two tables is not really an integral part of the sample (there's nothing against that sample just creating an empty database). If anyone would change that sample in the future to create a different data model, or no tables at all, this test would start failing. I don't think that kind of cross-dependence between unrelated tests and samples is a good thing.
2. See also [my comment here](https://github.com/GoogleCloudPlatform/dotnet-docs-samples/pull/1366#discussion_r680315614). By creating (and dropping) a single database, the test case is kept independent of all other tests and samples, and could easily also be run without all the initialization in `SpannerFixture`. Running only this test (including the creating and dropping of the database) takes approximately 5-7 seconds. The initialization of `SpannerFixture` takes 4 minutes when it has already been executed at least once, and most of the resources have already been created. That makes the development of this kind of simple samples a lot more of a hassle than it needs to be. It also sometimes makes it very difficult to determine exactly in what the state of a database and/or table is at the moment that a test is being executed. See for example [this test for one of the tagging samples](https://github.com/GoogleCloudPlatform/dotnet-docs-samples/blob/20102071d977d57b3f1ce00da77a54d97df2f6b3/spanner/api/Spanner.Samples.Tests/TransactionTagAsyncTest.cs#L34). I have absolutely no idea what the contents is of the `Venues` table at the moment that this test is being executed, as it might have been modified by any of the preceding sample tests.

spanner/api/Spanner.Samples.Tests/GetInstanceConfigAsyncTest.cs
```diff
@@ -0,0 +1,36 @@
...
+
+    public GetInstanceConfigAsyncTest(SpannerFixture spannerFixture)
+    {
+        _spannerFixture = spannerFixture;
+    }
+
+    [Fact]
+    public async Task TestGetInstanceConfigAsync()
+    {
+        var configId = "nam6";
```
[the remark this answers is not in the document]
olavloite: Done.

The config is by the way not created. The available configs are defined by Cloud Spanner: https://cloud.google.com/spanner/docs/instance-configurations#available-configurations-regional

spanner/api/Spanner.Samples.Tests/ListDatabasesTest.cs
```diff
@@ -0,0 +1,38 @@
...
+
+    public ListDatabasesTest(SpannerFixture spannerFixture)
+    {
+        _spannerFixture = spannerFixture;
+    }
+
+    [Fact]
+    public async Task TestListDatabases()
+    {
+        await _spannerFixture.RunWithTemporaryDatabaseAsync(async databaseId =>
```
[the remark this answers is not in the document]
olavloite: See also my comment above: As creating and dropping databases is relatively quick nowadays (less than 10 seconds), I would prefer to keep these tests independent of all the setup in `SpannerFixture`, as initializing `SpannerFixture` takes 4 minutes (at least in my test project) even after it has executed multiple tests.

spanner/api/Spanner.Samples.Tests/ListInstanceConfigsTest.cs
```diff
@@ -0,0 +1,35 @@
...
+    {
+        _spannerFixture = spannerFixture;
+    }
+
+    [Fact]
+    public void TestListInstanceConfigs()
+    {
+        var sample = new ListInstanceConfigsSample();
+        var configs = sample.ListInstanceConfigs(_spannerFixture.ProjectId);
+        Assert.True(configs.Any());
```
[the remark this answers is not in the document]
olavloite: Yup, Assert.NotEmpty() exists. Changed.

spanner/api/Spanner.Samples.Tests/SpannerFixture.cs
```diff
@@ -90,6 +93,7 @@ public async Task InitializeAsync()
     public Task DisposeAsync()
     {
         DeleteInstance(InstanceIdWithProcessingUnits);
+        DeleteInstance(InstanceIdWithMultiRegion);
```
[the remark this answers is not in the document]
olavloite: That would from a technical viewpoint certainly be possible, but I don't know what the possible resource limits are for the test project. @amanda-tarafa and @skuruppu Do you know if there are any limits, and if so, what they are? (I know that we keep the default test instance alive, and that it does not cause any problems with any limits, but removing the deletion of both of these would triple the number of instances that is kept alive)

spanner/api/Spanner.Samples.Tests/SpannerFixture.cs
```diff
@@ -438,4 +465,40 @@ public async Task RefillMarketingBudgetsAsync(int firstAlbumBudget, int secondAl
...
+        }
+
+        try
+        {
+            await testFunction(databaseId);
+        }
+        finally
+        {
+            // Cleanup the test database.
+            await databaseAdminClient.DropDatabaseAsync(DatabaseName.FormatProjectInstanceDatabase(ProjectId, instanceId, databaseId));
```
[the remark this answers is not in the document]
olavloite: I see the added value for the above for resources that are used for multiple tests, but I don't think that it would be possible for this kind of temporary test databases. There is a hard limit of max 100 databases per instance, and as the cleanup is only done during startup (and only deletes a limited number of resources), a small number of tests that use a temporary database could already cause a test run to hit that limit.
As noted above, the long initialization time and monolithic nature of the `SpannerFixture` is more a hassle than help for small, simple tests for features that do not involve slow processes (such as for example backups).

spanner/api/Spanner.Samples.Tests/UpdateDatabaseWithDefaultLeaderAsyncTest.cs
```diff
@@ -0,0 +1,47 @@
...
+
+    public UpdateDatabaseWithDefaultLeaderAsyncTest(SpannerFixture spannerFixture)
+    {
+        _spannerFixture = spannerFixture;
+    }
+
+    [Fact]
+    public async Task TestUpdateDatabaseWithDefaultLeaderAsync()
+    {
+        await _spannerFixture.RunWithTemporaryDatabaseAsync(_spannerFixture.InstanceIdWithMultiRegion, async databaseId =>
```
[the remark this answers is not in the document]
olavloite: (See also my comments above on the `SpannerFixture`)

The tradeoff here is that:
1. Having a fixed database on the multi-region instance would mean that for this test to actually do something, we would need to first set the default leader to X, and then let the sample change it to Y. Executing two separate `ALTER DATABASE` statements on an existing database actually takes longer (49.1 sec) than executing one `CreateDatabase` RPC with an `ALTER DATABASE` statement in the `ExtraStatements` list and then one `ALTER DATABASE` statement (30.6 sec).
2. By not using a fixed database that is tied to `SpannerFixture`, it would be easier to make this test independent of `SpannerFixture`, which again would make it quicker to run as a standalone test.

skuruppu:
Adding @hengfengli to answer some of the question for me since I'm out sick today.

hengfengli:
> Mostly for @skuruppu : several of the samples here, ListDatabases for instance, seem to not be exclusively related to leader, for instance, their region tags have no leader in the name. The only inmediate relation to leader is that is what they print on the console. My main question is, if there's more database metadata that we'll want to print , we'll we use these same samples, or will we have new ones? If we are going to have different samples for different metadata aspects, then these here should have leader on the class name, etc. but most importantly on the region tag.

@amanda-tarafa If we have more DB metadata to print, we will use these same samples. I designed this on purpose to avoid to add too many similar samples for the situation you describe.

spanner/api/Spanner.Samples.Tests/SpannerFixture.cs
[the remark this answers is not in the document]
hengfengli: I think it can hit the max limit easily if we don't delete the instances. Also, we had an issue to see a large amount of empty LCI instances before. I would not recommend to keep them alive.

spanner/api/Spanner.Samples.Tests/GetDatabaseDdlAsyncTest.cs
amanda-tarafa: Hmm, somehow I missed this comment when I reviewed earlier.

I think the best option we have here is to have different fixtures classes, a base one, for the common code and initialization (Project Id, for instance, etc..) and then specific ones for the biggest use cases. This is what we tend to do in libraries.
So maybe we have a fixture that initializes the backups, and maybe another that initializes the multi region instance, etc...

Any of that of course can be done in a separate PR, but I agree that our current fixture is getting a little bit overwhelming in terms of how long it takes it to initialize and how much (unrelated) code it is holding.

spanner/api/Spanner.Samples.Tests/GetDatabaseDdlAsyncTest.cs
amanda-tarafa: OK on point 1. (I've never been a fan of using different samples for testing another one but we started like that so, .... and I do get that it avoid some code duplication.)

On point 2, see [my comment](https://github.com/GoogleCloudPlatform/dotnet-docs-samples/pull/1366#discussion_r683605984) earlier.

spanner/api/Spanner.Samples.Tests/SpannerFixture.cs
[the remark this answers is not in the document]
amanda-tarafa: They'll be deleted on the next test run. And we can be more aggresive deleting old ones (we currently delete the 10 oldest I think). Although I prefer my suggestion of keeping them "for a while", as it is sometimes useful to inspect the resource used by a failed test, I'm not terrible bothered by deleting them here.

amanda-tarafa:
> > Mostly for @skuruppu : several of the samples here, ListDatabases for instance, seem to not be exclusively related to leader, for instance, their region tags have no leader in the name. The only inmediate relation to leader is that is what they print on the console. My main question is, if there's more database metadata that we'll want to print , we'll we use these same samples, or will we have new ones? If we are going to have different samples for different metadata aspects, then these here should have leader on the class name, etc. but most importantly on the region tag.
> 
> @amanda-tarafa If we have more DB metadata to print, we will use these same samples. I designed this on purpose to avoid adding too many similar samples for the situation you describe.

That's my preference as well, thanks!

spanner/api/Spanner.Samples.Tests/GetDatabaseDdlAsyncTest.cs
amanda-tarafa: @olavloite I think most of the outstanding comments are related to this suggestion here of splitting the fixture into more logical units and better managing initialization and deletion there. I'm happy to merge this PR as is if we can work on this after. Else, I'll probably take a closer look at a couple of things (nothing major, just want to make sure that we are consistent in our approach to resources, etc.)

olavloite: I'm certainly open to working on splitting the fixture into more logical units in one or more separate PRs. Your suggestion to split it into a base for ProjectId etc., and a couple of more feature specific fixtures sounds good. Just let me know if you want me to start with a draft PR, or if you want to start with a suggestion.

amanda-tarafa: Perfect. Before starting any work, I'm creating an issue so we can track our preferences there and reach an agreement beforehand. Thanks!

amanda-tarafa (approved):
LGTM

PR #1386, opened 2021-08-02
Intialize DB before app run in Cloud SQL samples

### LINKED ISSUE #1385: Cloud SQL samples should initialize DB before web app startup
The Cloud SQL sample app for Postgres, MySQL and SQL Server currently call the method to initialize the database (to create "votes" table) after the call to startup the sample web app, resulting in flakiness/errors like the following:

[From Cloud Console Logging]
{
  "textPayload": "    MessageText: relation \"votes\" does not exist",
...
}

jsimonweb:
Fixes #1385


enocom (approved):

amanda-tarafa:
Failures are unrelated and because we have reached quota because of all the renovate PRs.

PR #1399, opened 2021-08-03
feat: add samples for tagging

olavloite:
Adds samples for request and transaction tags.

(Build errors are not Spanner related)


skuruppu (approved):

spanner/api/Spanner.Samples/TransactionTagAsync.cs
```diff
@@ -0,0 +1,86 @@
...
+                venues.Add(venue);
+                // Update the capacity of the venue.
+                venue.Capacity /= 4;
+                var updateCommand = connection.CreateDmlCommand(
+                    "UPDATE Venues SET Capacity = @capacity WHERE VenueId = @venueId",
+                    new SpannerParameterCollection
+                    {
+                        {"capacity", SpannerDbType.Int64, venue.Capacity},
+                        {"venueId", SpannerDbType.Int64, venue.VenueId}
+                    });
```
amanda-tarafa, line 74: It feels wrong to go through all this trouble when we can just do `UPDATE Venues SET Capacity = Capacity/4`. This is the kind of sample that promotes bad practices. I know we had a few of these from before but we have fixed them. If we want to show two different command tags being set, can we do an update and an insert instead of the unnecesary select?

amanda-tarafa (changes requested):
A tiny nit and a (soft) change request. Something maybe to consider for all equivalent samples in other languages as well.

spanner/api/Spanner.Samples/RequestTag.cs
```diff
@@ -0,0 +1,58 @@
...
+// limitations under the License.
+
+// [START spanner_set_request_tag]
+
+using Google.Cloud.Spanner.Data;
+using System;
+using System.Collections.Generic;
+using System.Threading.Tasks;
+
+public class RequestTagAsyncSample
```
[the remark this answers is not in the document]
olavloite: Done

spanner/api/Spanner.Samples/TransactionTagAsync.cs
olavloite: Good point. Once again this is something that should (probably) be kept equal across all programming languages, so it's not really my call. @skuruppu Any thoughts on this?

spanner/api/Spanner.Samples/TransactionTagAsync.cs
```diff
@@ -0,0 +1,86 @@
...
+                  FROM Venues
+                  WHERE OutdoorVenue = @outdoorVenue",
+                new SpannerParameterCollection
+                {
+                    new SpannerParameter("outdoorVenue", SpannerDbType.Bool, false)
+                });
+            selectCommand.Transaction = transaction;
+            // Sets the request tag to "app=concert,env=dev,action=select".
+            // This request tag will only be set on this request.
+            selectCommand.Tag = "app=concert,env=dev,action=select";
```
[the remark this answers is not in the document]
amanda-tarafa: But aren't the apt,env tags being set already by the transaction? Why do you need to repeat them on the command? Or did I missunderstood the tagging feature?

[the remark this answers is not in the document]
olavloite: The tagging feature is a little bit complicated when it comes to the combination of transaction tags and request tags. The `ExecuteSqlRequest` that is sent to the contains a `RequestOptions` message, and `RequestOptions` has two different independent fields for request and transaction tag. So in the example above a `RequestOptions` message with the following values would be created:
- TransactionTag: `app=concert,env=dev`
- RequestTag: `action=select`

My understanding is that the request tag can be used to group or search for queries in the SPANNER_SYS.QUERY_STATS_* tables, while the transaction tag can be used for the SPANNER_SYS.TXN_STATS_* tables. I'm not sure whether the transaction tag can __also__ be used to group/search statements in the SPANNER_SYS.QUERY_STATS_* table.

skuruppu:
Adding @larkee to reply to the questions for me since I'm out sick.

spanner/api/Spanner.Samples/TransactionTagAsync.cs
larkee: Hi all, I wrote the samples for this. I was avoiding inserts as I was worried they may affect other samples which isn't, admittedly, isn't a great reason :sweat_smile:  I am happy to update the samples to avoid promoting bad practices :+1: 

Are there practices I should avoid when rewriting the samples to include an `UPDATE` and an `INSERT` statement ?

spanner/api/Spanner.Samples/TransactionTagAsync.cs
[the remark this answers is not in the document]
larkee: > My understanding is that the request tag can be used to group or search for queries in the SPANNER_SYS.QUERY_STATS_* tables, while the transaction tag can be used for the SPANNER_SYS.TXN_STATS_* tables.

This is my understanding as well. I would consider the tags independent. The transaction tag is used for the transaction stats while the request tag is used for the query stats.

> I'm not sure whether the transaction tag can also be used to group/search statements in the SPANNER_SYS.QUERY_STATS_* table.

I don't believe the transaction tag can be used like this. I will follow up to confirm though.

spanner/api/Spanner.Samples/TransactionTagAsync.cs
olavloite: Thanks for the quick response.

Regarding good practices for `UPDATE` and `INSERT` statements: I would say 'prefer parameters above literals for constant values' in the statements. So `INSERT INTO Foo (Id, Value) VALUES (@id, @value)` above `INSERT INTO Foo (Id, Value) VALUES (1, 'test')`.

olavloite: Thanks for the quick response.

Regarding good practices for `UPDATE` and `INSERT` statements: I would say 'prefer parameters above literals for constant values' in the statements. So `INSERT INTO Foo (Id, Value) VALUES (@id, @value)` above `INSERT INTO Foo (Id, Value) VALUES (1, 'test')`.

spanner/api/Spanner.Samples/TransactionTagAsync.cs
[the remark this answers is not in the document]
amanda-tarafa: Ok, it seems a little counterintuitive to add the same labels at both levels, but if that's the feature, then the sample makes sense. Thanks both for the clarification!

spanner/api/Spanner.Samples/TransactionTagAsync.cs
amanda-tarafa: +1 to what @olavloite said. And in general, don't do in a sample anything you wouldn't do in your own application (as the unnecesary select here). I do understand the worry about the inserts interfering, we have sometimes struggle with that as well, but it's better to show as real life code as possible :)! Thanks again.

I'll wait till the new sample is out then?

olavloite: The specs and the sample have been updated. PTAL.

@larkee I've had to make two small changes to the code that deviate from the exact sample code:
- Use `DIV(Capacity, 4)` instead of `Capacity / 4` as Capacity is INT64.
- Add a value for `LastUpdatedTime` in the insert statement, as it is a required column.

(The same has also been applied to the [NodeJS sample](https://github.com/googleapis/nodejs-spanner/pull/1450/files))

skuruppu (approved):

amanda-tarafa (approved):
LGTM. Thanks!!

amanda-tarafa:
Failures are unrelated (Asset has reached quota and I'm on it). Force merging.

PR #1445, opened 2021-08-10
feat(Compute): Use paginated lists.

amanda-tarafa:
Updates Google.Cloud.Compute.V1 dependency.
Closes #1438

### DIFF
--- a/compute/api/Compute.Samples.Tests/ListImagesAsyncTest.cs
+++ b/compute/api/Compute.Samples.Tests/ListImagesAsyncTest.cs
@@ -1,0 +1,34 @@
+﻿/*
+ * Copyright 2021 Google LLC
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     https://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+using System.Threading.Tasks;
+using Xunit;
+
+namespace Compute.Samples.Tests
+{
+    [Collection(nameof(ComputeFixture))]
+    public class ListImagesAsyncTest
+    {
+        private readonly ComputeFixture _fixture;
+        private readonly ListImagesAsyncSample _listSample = new ListImagesAsyncSample();
+
+        public ListImagesAsyncTest(ComputeFixture fixture) => _fixture = fixture;
+
+        [Fact]
+        public async Task ListsImages() => // Just test that it doesn't fail.
+            await _listSample.ListImagesAsync(_fixture.PublicImagesProjectId);
+    }
+}
--- a/compute/api/Compute.Samples.Tests/ListImagesPerPageAsyncTest.cs
+++ b/compute/api/Compute.Samples.Tests/ListImagesPerPageAsyncTest.cs
@@ -1,0 +1,35 @@
+﻿/*
+ * Copyright 2021 Google LLC
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     https://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+using System.Threading.Tasks;
+using Xunit;
+
+namespace Compute.Samples.Tests
+{
+    [Collection(nameof(ComputeFixture))]
+    public class ListImagesPerPageAsyncTest
+    {
+        private readonly ComputeFixture _fixture;
+        private readonly ListImagesPerPageAsyncSample _listSample = new ListImagesPerPageAsyncSample();
+
+        public ListImagesPerPageAsyncTest(ComputeFixture fixture) => _fixture = fixture;
+
+        [Fact]
+        public async Task ListsImagesPerPage() => // Just test that it doesn't fail.
+            await _listSample.ListImagesPerPageAsync(_fixture.PublicImagesProjectId);
+
+    }
+}
--- a/compute/api/Compute.Samples/ListImagesAsync.cs
+++ b/compute/api/Compute.Samples/ListImagesAsync.cs
@@ -1,0 +1,54 @@
+﻿/*
+ * Copyright 2021 Google LLC
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     https://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+// [START compute_images_list]
+
+using Google.Cloud.Compute.V1;
+using System;
+using System.Threading.Tasks;
+
+public class ListImagesAsyncSample
+{
+    public async Task ListImagesAsync(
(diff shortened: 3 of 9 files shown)


compute/api/Compute.Samples/ListZoneInstancesAsync.cs
```diff
@@ -31,30 +31,15 @@ public class ListZoneInstancesAsyncSample
...
         IList<Instance> allInstances = new List<Instance>();
 
         // Make the request to list all VM instances in the given zone in the specified project.
-        ListInstancesRequest request = new ListInstancesRequest 
-        { 
-            Project = projectId,
-            Zone = zone,
-        };
-        do
+        await foreach(var instance in client.ListAsync(projectId, zone))
```
jskeet, line 37: An alternative to await foreach here would be to `await client.ListAsync(projectId, zone).ToListAsync()` to come up with `allInstances`, then synchronously iterate over that, but I'm not bothered either way.

jskeet (approved):

compute/api/Compute.Samples/ListZoneInstancesAsync.cs
amanda-tarafa: This one gives you he chance to break in the middle of the async iteration without having fetched everything, right? Even if we are not showing that,...

jskeet: Yes, it does. (And will show results as it goes, of course.) I just twitch at manually adding to the list ;)

amanda-tarafa: Ah yes, that's for the purposes of the tests only, but I do see what you mean about that making you twitch. We have several tests that do this kind of thing across the repo. It's the alternative to testing against stdout.

amanda-tarafa:
@rsamborski I've added the two new samples, PTAL. Thanks.

compute/api/Compute.Samples/ListImagesPerPageAsync.cs
```diff
@@ -0,0 +1,55 @@
...
+        string projectId = "your-project-id")
+    {
+        // Initialize client that will be used to send requests. This client only needs to be created
+        // once, and can be reused for multiple requests.
+        ImagesClient client = await ImagesClient.CreateAsync();
+
+        // Make the request to list all non-deprecated images in a project.
+        ListImagesRequest request = new ListImagesRequest
+        {
+            Project = projectId,
```
Sita04, line 36: `MaxResults = 2,`
Kindly add MaxResults parameter, which tells the number of images to be displayed in each page.
If not added, all images will be displayed in a single page.

compute/api/Compute.Samples.Tests/ListImagesAsyncTest.cs
```diff
@@ -0,0 +1,34 @@
...
+    public class ListImagesAsyncTest
+    {
+        private readonly ComputeFixture _fixture;
+        private readonly ListImagesAsyncSample _listSample = new ListImagesAsyncSample();
+
+        public ListImagesAsyncTest(ComputeFixture fixture) => _fixture = fixture;
+
+        [Fact]
+        public async Task ListsImages() => // Just test that it doesn't fail.
+            await _listSample.ListImagesAsync(_fixture.ProjectId);
```
Sita04, line 32: Instead of using projectId here, you could also use "image families" and list images under them. 
Eg: windows-sql-cloud

compute/api/Compute.Samples/ListImagesAsync.cs
```diff
@@ -0,0 +1,48 @@
...
+        string projectId = "your-project-id")
+    {
+        // Initialize client that will be used to send requests. This client only needs to be created
+        // once, and can be reused for multiple requests.
+        ImagesClient client = await ImagesClient.CreateAsync();
+
+        // Make the request to list all non-deprecated images in a project.
+        ListImagesRequest request = new ListImagesRequest
+        {
+            Project = projectId,
```
Sita04, line 36: `MaxResults = 2,` kindly include this. MaxResults states the number of images to list on each page.
If not specified, all images will be listed in a single page.

compute/api/Compute.Samples/ListImagesPerPageAsync.cs
```diff
@@ -0,0 +1,55 @@
...
+            // Listing only non-deprecated images to reduce the size of the reply.
+            Filter = "deprecated.state != DEPRECATED"
+        };
+        // Calling AsRawResponses will return a sequence of pages.
```
Sita04, line 40:
```suggestion
      // Use the `AsRawResponses` attribute of returned iterable to have more granular control of
      // iteration over paginated results from the API. Each time you want to access the
      // next page, the library retrieves that page from the API.
```
Suggesting the comment for standardisation across languages.

compute/api/Compute.Samples/ListImagesAsync.cs
```diff
@@ -0,0 +1,48 @@
...
+            // Listing only non-deprecated images to reduce the size of the reply.
+            Filter = "deprecated.state != DEPRECATED"
+        };
+        await foreach (var image in client.ListAsync(request))
```
Sita04, line 40: // Although the `MaxResults` parameter is specified in the request, the iterable returned
      // by the `ListAsync()` method hides the pagination mechanic. The library makes multiple
      // requests to the API for you, so you can simply iterate over all the images.

Kindly add the above comment. Thanks!

Sita04 (changes requested):
Thanks Amanda!

compute/api/Compute.Samples.Tests/ListImagesAsyncTest.cs
rsamborski: I agree with Sita, using image family will give us a long list that should always fall under pagination.

compute/api/Compute.Samples.Tests/ListImagesPerPageAsyncTest.cs
```diff
@@ -0,0 +1,35 @@
...
+    public class ListImagesPerPageAsyncTest
+    {
+        private readonly ComputeFixture _fixture;
+        private readonly ListImagesPerPageAsyncSample _listSample = new ListImagesPerPageAsyncSample();
+
+        public ListImagesPerPageAsyncTest(ComputeFixture fixture) => _fixture = fixture;
+
+        [Fact]
+        public async Task ListsImagesPerPage() => // Just test that it doesn't fail.
+            await _listSample.ListImagesPerPageAsync(_fixture.ProjectId);
```
rsamborski, line 32: Same as with https://github.com/GoogleCloudPlatform/dotnet-docs-samples/pull/1445/files#r687222108

rsamborski (approved):
Thanks Amanda. LGTM once Sita's comments are implemented.

compute/api/Compute.Samples/ListImagesPerPageAsync.cs
amanda-tarafa: Done, but changed slightly for idiomacity:
- AsRawResponses is a method not a property (or attribute in Java), so it's more accurate to say `Call the AsRawResponse() method...`
- iterable => sequence
- distinguishing between the first sequence and the second. The first is the images sequence and the second is the page sequence.

I've change it to:

```
// Call the AsRawResponses() method of the returned image sequence to access the page sequece instead
// This allows you to have more granular control of iteration over paginated results from the API.
// Each time you request the next element on the page sequence, the library retrieves that page from the API.
```

amanda-tarafa (commented):
Review feedback addressed, PTAL @Sita04 , thanks!

compute/api/Compute.Samples/ListImagesPerPageAsync.cs
rsamborski: We used 10 in PHP and Python for the paginated results as this seemed to be a good number for demonstration purposes.

compute/api/Compute.Samples/ListImagesAsync.cs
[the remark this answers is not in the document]
rsamborski: I agree with 100 here, because we are not showing results page by page and just using this under the hood to generate multiple requests.

compute/api/Compute.Samples/ListImagesPerPageAsync.cs
amanda-tarafa: Yes 10 seems better, I did list sevaral of the public image sets and couldn't find any set containing more than 200 non-deprecated images. I'll change here to 10, but leave the fully automatic one on 100.

Sita04: Thanks Amanda for the explanation! I understand the rational behind the param setting. Will double check the param values in other languages as well.

compute/api/Compute.Samples/ListImagesPerPageAsync.cs
Sita04: Thank you for changing.

Sita04:
Thanks for making the changes, @amanda-tarafa! LGTM

Sita04 (approved):

amanda-tarafa:
Thanks! Failures in CI are unrelated, will force merge.

PR #1446, opened 2021-08-10
IsCompleted is checked instead of the metadata in AnalyzeIamPolicyLongrunning.

aaronlichen-hp:
IsCompleted is checked For AnalyzeIamPolicyLongrunning API instead of the metadata since this is what customers would first care about.


amanda-tarafa:
Ah, we crossed PRs. See #1454 .
I'll close this one as a duplicate basically. Thanks!

PR #1454, opened 2021-08-11
fix(Asset): Modify samples to acommodate API breaking change.

amanda-tarafa:


asset/api/Asset.Samples/AnalyzeIamPolicyLongrunningBigquery.cs
```diff
@@ -57,8 +54,14 @@ public class AnalyzeIamPolicyLongrunningBigquerySample
         var operation = client.AnalyzeIamPolicyLongrunning(request);
         // Wait for it to complete
         operation = operation.PollUntilCompleted();
-        // Return the metadata(request)
-        return operation.Metadata;
+
+        // Check whether the operation failed.
```
jskeet, line 58: No need for this - just use operation.Result and it will throw if it's failed.

asset/api/Asset.Samples/AnalyzeIamPolicyLongrunningGcs.cs
```diff
@@ -56,8 +53,14 @@ public class AnalyzeIamPolicyLongrunningGcsSample
         var operation = client.AnalyzeIamPolicyLongrunning(request);
         // Wait for it to complete
         operation = operation.PollUntilCompleted();
-        // Return the metadata(request)
-        return operation.Metadata;
+
+        // Check whether the operation failed.
```
jskeet, line 57: Ditto. (If this is something we do in a lot of samples, let's talk about that.)

jskeet (approved):

asset/api/Asset.Samples/AnalyzeIamPolicyLongrunningBigquery.cs
amanda-tarafa: Ah, yes, done.

asset/api/Asset.Samples/AnalyzeIamPolicyLongrunningGcs.cs
amanda-tarafa: Some samples do this, but to explicitly write some output, which is better than writing that output on a catch I think. But here, it makes no sense because we are simply throwing.

PR #1201, opened 2020-09-30
[Storage]: remove old samples

laljikanjareeya:
All the sample has been refactor based on new canonical style and [PR](https://github.com/GoogleCloudPlatform/dotnet-docs-samples/pull/973) has been merged.

PR #1455, opened 2021-08-13
chore(storage): Undo dependecy update

amanda-tarafa:
This is just we can avoid conflicts in #1201 which deletes this file anyway. And we don't have access to that PR's branch anymore.


amanda-tarafa:
Self merging, as this is very straightforward.

PR #1469, opened 2021-08-26
Adresses several flakyness causing issues, and general cleanup.

amanda-tarafa:


amanda-tarafa:
Note for reviewers: This PR contains many small changes to address flakyness, improvements, and general cleanup. Commits are all independent and some of them, although big, are just simple renamings.
You may want to only take a look at changes in projects you are an owner of.

Thanks.

sethvargo (approved):

appengine/flexible/Analytics/runTests.ps1
```diff
@@ -14,10 +14,10 @@
 
 Import-Module -DisableNameChecking ..\..\..\BuildTools.psm1
 
-dotnet restore
+dotnet restore --force
```
jskeet, line 17: Can't say I've ever seen this make a difference, but I wouldn't expect it to cause any problems, at least.

jskeet (approved):

appengine/flexible/Analytics/runTests.ps1
amanda-tarafa: Yep, I'm not sure either tbh, but it does says on docs that the other commands (build, test, run...) perform a restore "if necessary". We have #1278 that might be due to restore not being done properly. If this doesn't help I'll try forcing it to always fetch packages from Nuget.

PR #1488, opened 2021-09-07
Addresses some more flakiness

amanda-tarafa:


compute/api/Compute.Samples.Tests/ComputeFixture.cs
```diff
@@ -52,6 +54,9 @@ public class ComputeFixture : IDisposable, ICollectionFixture<ComputeFixture>
 
         private StorageClient StorageClient { get; } = StorageClient.Create();
 
+        public RetryRobot Assert { get; } = new RetryRobot { RetryWhenExceptions = new Type[] { typeof(XunitException) } };
+
```
jskeet, line 58: If you feel like removing one of these blank lines, that would be good - but don't worry :)

jskeet (approved):

compute/api/Compute.Samples.Tests/ComputeFixture.cs
amanda-tarafa: Done!

PR #1496, opened 2021-09-09
tests(util): Jitter the retry robot delay.

amanda-tarafa:


testutil/TestUtil.cs
```diff
@@ -73,7 +76,12 @@ public T Eventually<T>(Func<T> func)
                 catch (Exception e)
                 when (ShouldCatch(e) && i < MaxTryCount)
                 {
-                    Thread.Sleep(delayMs);
+                    int jitteredDelayMs;
+                    lock(_lock)
+                    {
+                        jitteredDelayMs = (int)(_random.NextDouble() * delayMs);
```
jskeet, line 82: This is fine if we're happy with the average being half the specified delay. You *might* want to add `delayMs/2`, so that it's between 0.5x and 1.5x `delayMs`. Having said which, the GAX jitter looks basically like this...

jskeet (approved):

testutil/TestUtil.cs
amanda-tarafa: I copied it from GAX :), where the unjittered delay is understood as MaxDelay.
But thinking about it more, and given then issues we see here because of concurrent test executions and quotas, we might want to see the unjittered delay as the minimum delay, so, I'll actually add `delayMs` so that we get 1x to 2x values.

amanda-tarafa: Ah no, I'll add `delayMs/2` as you suggested, as the default delay multiplier is 2, so if I add `delayMs` I'll be just "catching up" with the next iteration (of a concurrent one), which means that jittering would be less effective.

PR #1545, opened 2021-11-09
tests(Compute): Load tools from the correct path.

amanda-tarafa:
This is why timeout was not having any effect.


jskeet (approved):

amanda-tarafa:
Merging. The only error is a Pub/Sub Unavailable error and that's transient by definition. If it starts happening often, I'll take a closer look and possible add retries.

PR #1567, opened 2021-12-07
Addresses some more flakiness

amanda-tarafa:


asset/api/Asset.Samples.Tests/AssetFixture.cs
```diff
@@ -28,7 +29,8 @@ public class AssetFixture : IDisposable, ICollectionFixture<AssetFixture>
     public string BucketName { get; }
     public RetryRobot Retry { get; } = new RetryRobot 
     { 
-        RetryWhenExceptions = new Type[] { typeof(XunitException) },
+        ShouldRetry = ex => ex is XunitException
```
jskeet, line 32: I'd use more pattern matching here, assuming the SDK version allows you to:

```csharp
ShouldRetry = ex => ex is XunitException || ex is RpcException { StatusCode: StatusCode.DeadlineExceeded };
```

jskeet (approved):
Approving in case you can't make make use of the suggestion :)

asset/api/Asset.Samples.Tests/AssetFixture.cs
amanda-tarafa: Will try to see if it works. I have to fix a couple other more flakes anyway.

Issue #1509, opened 2021-09-29
Add samples for turbo replication (15 minute RPO)


ddelgrosso1:
Samples need to be added for turbo replication ([client library request](https://docs.google.com/document/d/1ee3LzRfyArNIR4rLEeY60Eio7DBwLy5TkmvcMPxSork/edit?usp=sharing&resourcekey=0-4CYSH1Gc0ekxa1oUntE9yg)). 

Please see NodeJS example here: https://github.com/googleapis/nodejs-storage/pull/1618

Please wait for approval from myself before merging or releasing this change so we can coordinate around release dates.

amanda-tarafa:
@shivgautam assigning to you so you can reassing appropiately.

I can take it if you don't have the bandwidth.

shivgautam:
@amanda-tarafa I'll get it done. Just got the timelines from Denis. This should be doable.

anshuldavid13:
Samples done and merged

PR #1577, opened 2021-12-21
Turbo replication changes

anshuldavid13:
PR for https://github.com/GoogleCloudPlatform/dotnet-docs-samples/issues/1509


storage/api/Storage.Samples/SetRPODefault.cs
```diff
@@ -0,0 +1,35 @@
...
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+// [START storage_set_rpo_default]
+
+using System;
+using Google.Apis.Storage.v1.Data;
+using Google.Cloud.Storage.V1;
+
+public class SetRPODefaultSample
```
amanda-tarafa, line 21:
```suggestion
public class SetRpoDefaultSample
```
Here and everywhere, including file names.

storage/api/Storage.Samples/SetRPODefault.cs
```diff
@@ -0,0 +1,35 @@
...
+{
+    public void SetRPODefault(
+        string bucketName = "your-unique-bucket-name")
```
amanda-tarafa, line 24: When it's just a few parameters, there's no need to put them on the next line.
```suggestion
    public void SetRPODefault(string bucketName = "your-unique-bucket-name")
```

storage/api/Storage.Samples/SetRPODefault.cs
```diff
@@ -0,0 +1,35 @@
...
+    {
+        var storage = StorageClient.Create();
+        var bucket = storage.GetBucket(bucketName);
+        bucket.Rpo = "DEFAULT";
+        storage.UpdateBucket(bucket);
+
+        Console.WriteLine($"Turbo replication disabled for bucket {bucketName}");
```
amanda-tarafa, line 31: Does setting it to default means that it is disabled? Just double checking this.

storage/api/Storage.Samples.Tests/SetRPODefaultTest.cs
```diff
@@ -0,0 +1,41 @@
...
+        _fixture = fixture;
+    }
+
+    [Fact]
+    public void SetRPODefault()
+    {
+        SetRPODefaultSample setRPODefaultSample = new SetRPODefaultSample();
+        GetBucketMetadataSample getBucketMetadataSample = new GetBucketMetadataSample();
+        
+        setRPODefaultSample.SetRPODefault(_fixture.BucketNameGeneric);
```
amanda-tarafa, line 34: Should you create a new bucket for this test? I don't remember what the other tests that change the metadata do, but maybe we shouldn't change the metadata for the generic bucket as it might make it useless for tests that come later.

storage/api/Storage.Samples/SetRPOAsyncTurbo.cs
```diff
@@ -0,0 +1,36 @@
...
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+// [START storage_set_rpo_async_turbo]
+
+using System;
+using Google.Apis.Storage.v1.Data;
+using Google.Cloud.Storage.V1;
+
```
amanda-tarafa, line 20:
```suggestion
```

storage/api/Storage.Samples.Tests/SetRPODefaultTest.cs
```diff
@@ -0,0 +1,41 @@
...
+    public SetRPODefaultTest(StorageFixture fixture)
+    {
+        _fixture = fixture;
+    }
+
+    [Fact]
+    public void SetRPODefault()
+    {
+        SetRPODefaultSample setRPODefaultSample = new SetRPODefaultSample();
+        GetBucketMetadataSample getBucketMetadataSample = new GetBucketMetadataSample();
```
amanda-tarafa, line 32: Maybe use the GetBucketRpo sample instead?

storage/api/Storage.Samples/CreateBucketWithTurboReplication.cs
```diff
@@ -0,0 +1,40 @@
...
+
+public class CreateBucketWithTurboReplicationSample
+{
+    public Bucket CreateBucketWithTurboReplication(
+        string projectId = "your-project-id",
+        string bucketName = "your-unique-bucket-name",
+        string location = "your-bucket-location")
+    {
+        var storage = StorageClient.Create();
+        var bucket = storage.CreateBucket(projectId, bucketName);
```
amanda-tarafa, line 29: I think you can use the CreateBucket overload that takes a Bucket and set all fields from the start, that way you don't have to make to calls (create and update) but just the create one.

storage/api/Storage.Samples.Tests/CreateBucketWithTurboReplicationTest.cs
```diff
@@ -0,0 +1,41 @@
...
+    public CreateBucketWithTurboReplicationTest(StorageFixture fixture)
+    {
+        _fixture = fixture;
+    }
+
+    [Fact]
+    public void CreateBucketWithTurboReplication()
+    {
+        CreateBucketWithTurboReplicationSample createBucketWithTurboReplication = new CreateBucketWithTurboReplicationSample();
+        var bucketName = Guid.NewGuid().ToString();
```
amanda-tarafa, line 35: Add a method on the fixture, something like GenerateBucketName, and add that bucket name into the TempBuckets (see the method CreateBucket in the fixture), so this bucket will be removed on clean up.

amanda-tarafa (changes requested):
Fine overall, a few request changes.

And some comments will apply everywhere, not just where I made it.

storage/api/Storage.Samples/SetRPODefault.cs
anshuldavid13: yes 
image

storage/api/Storage.Samples.Tests/SetRPODefaultTest.cs
anshuldavid13: I could see these tests creating fresh temporary buckets in the instance for each test run. Am I missing something?

ddelgrosso1:
@anshuldavid13 once this is approved please go ahead and merge it when you can. We are trying to get this feature / samples released this week. Thanks!

storage/api/Storage.Samples/CreateBucketWithTurboReplication.cs
```diff
@@ -0,0 +1,36 @@
...
+public class CreateBucketWithTurboReplicationSample
+{
+    public Bucket CreateBucketWithTurboReplication(
+        string projectId = "your-project-id",
+        string location = "your-bucket-location")
+    {
+        var storage = StorageClient.Create();
+        var bucket = storage.CreateBucket(projectId, new Bucket { Location = location, Rpo = "ASYNC_TURBO" });
+
+        Console.WriteLine($"Created {bucket.Name} in {bucket.Location} with RPO Setting {bucket.Rpo}.");
```
amanda-tarafa, line 30: we need to add this bucket.Name to the temp buckets for cleanup

```_fixture.TempBucketNames.Add(bucket.Name)```

File: storage/api/Storage.Samples/SetRpoAsyncTurbo.cs [hunk 10]
```diff
@@ -0,0 +1,34 @@
...
+// [START storage_set_rpo_async_turbo]
+
+using System;
+using Google.Cloud.Storage.V1;
+
+public class SetRpoAsyncTurboSample
+{
+    public void SetRpoAsyncTurbo(string bucketName = "your-unique-bucket-name")
+    {
+        // It requires a bucket with dual-region configuration for turbo replication to be enabled
```
amanda-tarafa, line 24:
```suggestion
        // Enabling turbo replication requires a bucket with dual-region configuration.
        // You can read more about dual-region configuration here: https://cloud.google.com/storage/docs/locations#location-dr
```

storage/api/Storage.Samples.Tests/SetRpoAsyncTurboTest.cs
```diff
@@ -0,0 +1,40 @@
...
+        _fixture = fixture;
+    }
+
+    [Fact]
+    public void SetRpoAsyncTurbo()
+    {
+        SetRpoAsyncTurboSample setRpoAsyncTurboSample = new SetRpoAsyncTurboSample();
+        GetRpoSample getRpoSample = new GetRpoSample();
+
+        // It requires a bucket with dual-region configuration for turbo replication to be enabled
```
amanda-tarafa, line 33:
```suggestion
        // Enabling turbo replication requires a bucket with dual-region configuration.
```

Does the generic bucket has dual-region configuration? Else, this test will always fail.

And also, this test and the one setting Default RPO may run at the same time, they'd both be changing the RPO setting for the same bucket so the last one that changes the value will "win" and the other test will fail.

In both these tests you need to create a new bucket and use that one instead of the generic one. Also, for this test in particular you need to make sure that the bucket you have created is dual-region.

storage/api/Storage.Samples/SetRpoAsyncTurbo.cs
amanda-tarafa: And change to this everywhere else.

amanda-tarafa (changes requested):
One nit, and one change we need to avoid tests failing.

amanda-tarafa (approved):
LGTM

I'll run tests and I can merge once it's green.

amanda-tarafa:
Error is an unrelated transient Monitoring error. Merging

PR #1579, opened 2021-12-21
File download/upload samples

anshuldavid13:


storage/api/Storage.Samples/UploadObjectFromMemory.cs
```diff
@@ -0,0 +1,41 @@
...
+using Google.Apis.Storage.v1.Data;
+using Google.Cloud.Storage.V1;
+using System.IO;
+
+public class UploadObjectFromMemorySample
+{
+    public string UploadObjectFromMemory(
+        string bucketName = "your-unique-bucket-name",
+        string objectName = "my-file-name",
+        string contents = "contents-to-be-uploaded")
```
amanda-tarafa, line 29:
```suggestion
        string contents = "Hello world!")
```

storage/api/Storage.Samples/UploadObjectFromMemory.cs
```diff
@@ -0,0 +1,41 @@
...
+    {
+        var storage = StorageClient.Create();
+        byte[] byteArray = Encoding.UTF8.GetBytes(contents);
+        MemoryStream stream = new MemoryStream(byteArray);
+        storage.UploadObject(bucketName, objectName, "application/octet-stream" , stream);
+
+        Console.WriteLine($" {objectName} uploaded to bucket {bucketName} with contents: {contents}");
+        return contents;
```
amanda-tarafa, line 37: Maybe don't return the contents and make this method void.

storage/api/Storage.Samples/DownloadFile.cs
```diff
@@ -13,6 +13,7 @@
 // limitations under the License.
 
 // [START storage_download_file]
+// [START storage_stream_file_download]
```
amanda-tarafa, line 16: @frankyn and @shaffeeullah Can you confirm that these two region tags may be covered by the same sample?

Please see this other issue and @jskeet's reply: https://github.com/googleapis/google-cloud-dotnet/issues/7638

storage/api/Storage.Samples/UploadFile.cs
```diff
@@ -13,6 +13,7 @@
 // limitations under the License.
 
 // [START storage_upload_file]
+// [START storage_stream_file_upload]
```
amanda-tarafa, line 16: @frankyn and @shaffeeullah Same question applies here as it did for download. See https://github.com/googleapis/google-cloud-dotnet/issues/7638

(I don't know the context of the documentation in which these samples will be shown, so I cannot be sure this is correct.)

storage/api/Storage.Samples/DownloadObjectIntoMemory.cs
```diff
@@ -0,0 +1,39 @@
...
+{
+    public Stream DownloadObjectIntoMemory(
+        string bucketName = "your-unique-bucket-name",
+        string objectName = "my-file-name")
+    {
+        var storage = StorageClient.Create();
+        Stream stream = new MemoryStream();
+        storage.DownloadObject(bucketName, objectName, stream);
+
+        Console.WriteLine($"The contents of {objectName} from bucket {bucketName} are {stream}");
```
amanda-tarafa, line 34: This won't print the contents I believe. It will just probably print "System.IO.MemoryStream"

amanda-tarafa (changes requested):
Looking good overall. I still want the Storage folk to confirm that the upload/download samples are equivalent to what they expect to be the streaming samples.

storage/api/Storage.Samples/DownloadFile.cs
shaffeeullah: These region tags may be covered by the same sample 👍🏾

storage/api/Storage.Samples/UploadFile.cs
shaffeeullah: Yep-- if this is how c# streaming works, this looks good too.

storage/api/Storage.Samples.Tests/UploadObjectFromMemoryTest.cs
```diff
@@ -0,0 +1,39 @@
...
+        _fixture = fixture;
+    }
+
+    [Fact]
+    public void UploadObjectFromMemory()
+    {
+        UploadObjectFromMemorySample uploadObjectFromMemorySample = new UploadObjectFromMemorySample();
+        ListFilesSample listFilesSample = new ListFilesSample();
+
+        uploadObjectFromMemorySample.UploadObjectFromMemory(_fixture.BucketNameGeneric, _fixture.FileName, "This is testing content");
```
amanda-tarafa, line 35: The fixture creates and object in the generic bucket with this file name. Which means that this test will always fail, or worst, it will override the existing file which is expected by some other test.

In general, the values that are defined on the fixture are to be used in tests that do not change them. Whatever your tests is going to change, the you need to create new and for your test only, and make sure that it is cleaned up afterwards. In this case you need to use a specific file name.

amanda-tarafa (changes requested):
On change request related to modifying fixture defined resources.
Please review this same thing in all other open PRs before merging.

storage/api/Storage.Samples/UploadObjectFromMemory.cs
```diff
@@ -0,0 +1,38 @@
...
+using Google.Cloud.Storage.V1;
+using System;
+using System.IO;
+using System.Text;
+
+public class UploadObjectFromMemorySample
+{
+    public void UploadObjectFromMemory(
+        string bucketName = "your-unique-bucket-name",
+        string objectName = "my-file-name",
```
amanda-tarafa, line 26: It's cofusing, as it seems to have meaning, you use "your" and "my". Maybe don't use any at all?

```suggestion
        string objectName = "your-file-name",
```

storage/api/Storage.Samples.Tests/UploadObjectFromMemoryTest.cs
```diff
@@ -0,0 +1,41 @@
...
+        _fixture = fixture;
+    }
+
+    [Fact]
+    public void UploadObjectFromMemory()
+    {
+        UploadObjectFromMemorySample uploadObjectFromMemorySample = new UploadObjectFromMemorySample();
+        ListFilesSample listFilesSample = new ListFilesSample();
+
+        var fileName = Guid.NewGuid().ToString();
```
amanda-tarafa, line 36: This is to make sure that the file is deleted so that the bucket can be deleted.
```suggestion
        var fileName = Guid.NewGuid().ToString();
        _fixture.Collect(fileName);
```

storage/api/Storage.Samples/DownloadObjectIntoMemory.cs
```diff
@@ -0,0 +1,36 @@
...
+
+using Google.Cloud.Storage.V1;
+using System;
+using System.IO;
+
+public class DownloadObjectIntoMemorySample
+{
+    public Stream DownloadObjectIntoMemory(
+        string bucketName = "your-unique-bucket-name",
+        string objectName = "my-file-name")
```
amanda-tarafa, line 25:
```suggestion
        string objectName = "your-file-name")
```

amanda-tarafa (changes requested):
Almost there as well!

amanda-tarafa (approved):
LGTM.

I'll wait for tests to go green before merging

amanda-tarafa:
A few unrelated errors, due to several tests runs running at the same time. Merging...

PR #1584, opened 2022-01-10
feat: Retail API code samples

t-karasova:


t-karasova:
Hi, 
This is a commit to add the code samples for Retail API.
To run the tests the secrets need to be added, whom I can ask to help me with this?

amanda-tarafa (changes requested):
I haven't looked in detail, but there are a few issues with these samples as they do not conform with our guidelines. Please see [here](https://github.com/googleapis/google-api-dotnet-client/blob/68ee5d3a9e8578f076aeea3d29d3657eda98aeac/Src/Support/Google.Apis/Upload/ResumableUpload.cs#L670) for the canonical sample and test set.
 One main issue is that we use xunit for testing, and these samples are using something else.

I'll also start an internal thread about these.

t-karasova:
> 

Thank you for your comment, I will rewrite the tests using XUnit, but I would like to draw your attention on the document
https://github.com/GoogleCloudPlatform/dotnet-docs-samples/blob/main/TESTING.md
here it is recommended to use MSTest

amanda-tarafa:
Ah but that's not recommending to use MSTest, that's just saying which tools you need in the path to runn **all the tests on the repo** (at some point xunit and MSTest coexisted.)
https://github.com/GoogleCloudPlatform/dotnet-docs-samples/blob/main/CONTRIBUTING.md is what should be followed when addind code to the repo.

In any case I've created #1590 to remove referecens from MSTest so this does not causes confusion in the future.

Also, since you are rewriting the tests I'll review a couple of samples as well to point out a few things that I saw earlier that wouldn't match our sample guidelines. In general write both samples and tests as close to the KMS suite as possible, including folder and solution structure etc.

Also, there are a couple of questions I'd like to ask about these samples, internally. I've sent you an email earlier in the week so if we can chat about these there that'd be great.

Thanks.

retail/api/retail.sln
```diff
@@ -0,0 +1,55 @@
+﻿
```
amanda-tarafa, line 1: We usually name solutions and projects with CamelCase as it is more idiomatic.

retail/api/grs-search/SearchWithQueryExpansion.cs
```diff
@@ -0,0 +1,119 @@
...
+// enabling the query expansion feature to let the Google Retail Search build an automatic query expansion.
+
+using Google.Cloud.Retail.V2;
+using Newtonsoft.Json;
+using Newtonsoft.Json.Serialization;
+using System;
+using System.Collections.Generic;
+using System.Linq;
+
+namespace grs_search.search
```
amanda-tarafa, line 26: We don't define namespaces in samples, to improve readability.

retail/api/grs-search/SearchWithQueryExpansion.cs
```diff
@@ -0,0 +1,119 @@
...
+{
+    public static class SearchWithQueryExpansion
+    {
+        private const string Endpoint = "retail.googleapis.com";
```
amanda-tarafa, line 30: If this is the default endpoint you dont need to specify it.

retail/api/grs-search/SearchWithQueryExpansion.cs
```diff
@@ -0,0 +1,119 @@
...
+        private static readonly string ProjectNumber = Environment.GetEnvironmentVariable("PROJECT_NUMBER");
```
amanda-tarafa, line 31: Please do not include this on the sample code but on the test code (usually in a fixture). Clients might not have/want to set the project Ids on environment variables.

All fields that a sample uses should be received as parameters for testing and for being able to call the sample method from anywhere.

retail/api/grs-search/SearchWithQueryExpansion.cs
```diff
@@ -0,0 +1,119 @@
...
+        private static readonly string DefaultSearchPlacement = $"projects/{ProjectNumber}/locations/global/catalogs/default_catalog/placements/default_search";
+
+        // Get search service client
+        private static SearchServiceClient GetSearchServiceClient()
```
amanda-tarafa, line 35: We don't use individual methods for each of the steps. We just write a method that creates the client, creates the request, executes the operation and handles the response. Please see KMS.

retail/api/grs-search/SearchWithQueryExpansion.cs
```diff
@@ -0,0 +1,119 @@
...
+        {
+            var searchServiceClientBuilder = new SearchServiceClientBuilder
+            {
+                Endpoint = Endpoint
+            };
+
+            var searchServiceClient = searchServiceClientBuilder.Build();
```
amanda-tarafa, line 42: This is not needed, the default endpoint will be set by default if calling SearchServiceClient.Create();

retail/api/grs-search/SearchWithQueryExpansion.cs
```diff
@@ -0,0 +1,119 @@
...
+            return searchServiceClient;
+        }
+
+        // Get search service request
+        private static SearchRequest GetSearchRequest(string query, SearchRequest.Types.QueryExpansionSpec.Types.Condition condition)
+        {
+            var queryExpansionSpec = new SearchRequest.Types.QueryExpansionSpec()
```
amanda-tarafa, line 49:
```suggestion
            var queryExpansionSpec = new SearchRequest.Types.QueryExpansionSpec
```

retail/api/grs-search/SearchWithQueryExpansion.cs
```diff
@@ -0,0 +1,119 @@
...
+            var jsonSerializeSettings = new JsonSerializerSettings
+            {
+                NullValueHandling = NullValueHandling.Ignore,
+                ContractResolver = new CamelCasePropertyNamesContractResolver(),
+                Formatting = Formatting.Indented
+            };
+
+            var searchRequestJson = JsonConvert.SerializeObject(searchRequest, jsonSerializeSettings);
+
+            Console.WriteLine("\nSearch. request: \n\n" + searchRequestJson);
```
amanda-tarafa, line 71: This is nor required for the code to work nor idiomatic. What is the sample trying to show when doing this?

retail/api/grs-search/SearchWithQueryExpansion.cs
```diff
@@ -0,0 +1,119 @@
...
+            return searchRequest;
+        }
+
+        // Call the Retail Search:
+        [Attributes.Example]
```
amanda-tarafa, line 76: Not sure what this does?

retail/api/grs-search/SearchWithQueryExpansion.cs
```diff
@@ -0,0 +1,119 @@
...
+        public static IEnumerable<SearchResponse> Search() // PagedEnumerable<SearchResponse, SearchResponse.Types.SearchResult>
```
amanda-tarafa, line 77: What does the comment at the end of the line mean?

retail/api/grs-search/SearchWithQueryExpansion.cs
```diff
@@ -0,0 +1,119 @@
...
+        {
+            // TRY DIFFERENT QUERY EXPANSION CONDITION HERE:
```
amanda-tarafa, line 79: Not in all caps please.

retail/api/grs-search/SearchWithQueryExpansion.cs
```diff
@@ -0,0 +1,119 @@
...
+            var condition = SearchRequest.Types.QueryExpansionSpec.Types.Condition.Auto;
+            string query = "Google Youth Hero Tee Grey";
+
+            var searchRequest = GetSearchRequest(query, condition);
+
+            var searchResponse = GetSearchServiceClient().Search(searchRequest).AsRawResponses(); //.AsRawResponses();
```
amanda-tarafa, line 85: Comment at the end of the line?

retail/api/grs-search/SearchWithQueryExpansion.cs
```diff
@@ -0,0 +1,119 @@
...
+                {
+                    results = firstSearchResponse.Results,
+                    totalSize = firstSearchResponse.TotalSize,
+                    attributionToken = firstSearchResponse.AttributionToken,
+                    nextPageToken = firstSearchResponse.NextPageToken,
+                    facets = firstSearchResponse.Facets,
+                    queryExpansionInfo = firstSearchResponse.QueryExpansionInfo
+                };
+
+                var serializedJson = JsonConvert.SerializeObject(objectToSerialize, jsonSerializeSettings);
```
amanda-tarafa, line 110: Again please don't do this.

If the sample wants to print some of the resulting items and some of the properties of the operation result please do so with code similar to:

```Console.WriteLine(response.NextPageToken)```

**Review by amanda-tarafa — CHANGES_REQUESTED** [review]
OK, I've reviewed one sample and I think most of my comments will apply to the rest as well.

I think it would be best if you close this PR and submit a new with just one or two of the rewritten samples with their tests. Once those are in good shape and merged, you can create PRs with 4 or 5 related samples and their test, so we can incrementally get it right.

Let me know.

File: retail/api/grs-search/SearchWithQueryExpansion.cs (hunk 4 above)
Comment (t-karasova) [reply retail/api/grs-search/SearchWithQueryExpansion.cs:31]: The purpose of this code samples is dual: first, they will be used as part of the documentation, here https://cloud.google.com/retail/docs/search-basic#search_product-python.

The second purpose is to run this code as the interactive tutorials, that will be run from the CloudShell. 
In this tutorials, we are asking user to create their own project and set environment variables PROJECT_NUMBER, PROJECT_ID, BUCKET_NAME, after that the user will be able to run the code samples.

For the tests we will use our project, and it's name, should be defined in the Secret Manager. I kindly ask your assistance in adding the values to the Secret Manager

File: retail/api/grs-search/SearchWithQueryExpansion.cs (hunk 9 above)
Comment (t-karasova) [reply retail/api/grs-search/SearchWithQueryExpansion.cs:76]: For the purpose of creating the interactive tutorials, we decided that for all languages the code samples should be run from CloudShell ide with the command in Terminal. 

So to run the sample with a short command "dotnet run -- SearchWithQueryExpansion" we have added retail/api/grs-search/Attributes/ExampleAttribute.cs which takes the command parameter and looks for the appropriate code sample

File: retail/api/grs-search/SearchWithQueryExpansion.cs (hunk 13 above)
Comment (t-karasova) [reply retail/api/grs-search/SearchWithQueryExpansion.cs:110]: We want to show our users the search response in the output. If we just print the raw response, it is not readable at all, so we decided to use pretty printed JSON

File: retail/api/grs-search/SearchWithQueryExpansion.cs (hunk 3 above)
Comment (t-karasova) [reply retail/api/grs-search/SearchWithQueryExpansion.cs:30]: It was a recommended by the Lei Chen - the Retail libraries dev manager to directly specify the endpoint to avoid authorization issues

File: retail/api/grs-search/SearchWithQueryExpansion.cs (hunk 6 above)
Comment (t-karasova) [reply retail/api/grs-search/SearchWithQueryExpansion.cs:42]: It was a recommended by the Lei Chen - the Retail libraries dev manager to directly specify the endpoint to avoid authorization issues

File: retail/api/grs-product/AddFulfillmentPlaces.cs [hunk 14]
```diff
@@ -0,0 +1,101 @@
...
+using Newtonsoft.Json.Serialization;
+using System;
+using System.Threading;
+
+namespace grs_product
+{
+    public static class AddFulfillmentPlaces
+    {
+        private const string Endpoint = "retail.googleapis.com";
+        private const string ProductId = "add_fulfillment_test_product_id";
```
chenlei1216, line 29: Should we use a random ID, has the same concern that product may not be deleted so the test may fail when creating the project with the same ID.

retail/api/grs-product/AddFulfillmentPlaces.cs
```diff
@@ -0,0 +1,101 @@
...
+            Thread.Sleep(120000);
+        }
+
+        [Attributes.Example]
+        public static Product PerformAddFulfillment()
+        {
+            CreateProduct.CreateRetailProductWithFulfillment(ProductId);
+            AddFulfillment(ProductName);
+            var inventoryProduct = GetProduct.GetRetailProduct(ProductName);
+            // DeleteProduct.DeleteRetailProduct(ProductName);
```
chenlei1216, line 95: Delete the product after test?

retail/api/grs-product/CrudProduct.cs
```diff
@@ -0,0 +1,228 @@
...
+using Newtonsoft.Json.Serialization;
+using System;
+
+namespace grs_product
+{
+    public static class CrudProduct
+    {
+        private const string Endpoint = "retail.googleapis.com";
+
+        private const string ProductId = "crud_product_id";
```
chenlei1216, line 29: Ditto. Should we use a random id, or ignore the creation failure?

retail/api/grs-product/RemoveFulfillmentPlaces.cs
```diff
@@ -0,0 +1,101 @@
...
+using Newtonsoft.Json.Serialization;
+using System;
+using System.Threading;
+
+namespace grs_product
+{
+    public static class RemoveFulfillmentPlaces
+    {
+        private const string Endpoint = "retail.googleapis.com";
+        private const string ProductId = "remove_fulfillment_test_product_id";
```
chenlei1216, line 29: ditto.

retail/api/grs-product/RemoveFulfillmentPlaces.cs
```diff
@@ -0,0 +1,101 @@
...
+            Thread.Sleep(120000);
+        }
+
+        [Attributes.Example]
+        public static Product PerformAddRemoveFulfillment()
+        {
+            CreateProduct.CreateRetailProductWithFulfillment(ProductId);
+            RemoveFulfillment(ProductName);
+            var inventoryProduct = GetProduct.GetRetailProduct(ProductName);
+            // DeleteProduct.DeleteRetailProduct(ProductName);
```
chenlei1216, line 95: Delete the product?

retail/api/grs-product/SetInventory.cs
```diff
@@ -0,0 +1,136 @@
...
+using Newtonsoft.Json.Serialization;
+using System;
+using System.Threading;
+
+namespace grs_product
+{
+    public static class SetInventory
+    {
+        private const string Endpoint = "retail.googleapis.com";
+        private const string ProductId = "inventory_test_product_id";
```
chenlei1216, line 29: ditto.

retail/api/grs-search/SearchWithBoostSpec.cs
```diff
@@ -0,0 +1,124 @@
...
+            Console.WriteLine("\nSearch. request: \n\n" + searchRequestJson);
+            return searchRequest;
+        }
+
+        // Call the Retail Search:
+        [Attributes.Example]
+        public static IEnumerable<SearchResponse> Search()
+        {
+            // TRY DIFFERENT CONDITIONS HERE:
+            string condition = "colorFamily: ANY(\"Blue\")";
```
chenlei1216, line 85: colorFamilies

retail/api/grs-search/SearchWithFiltering.cs
```diff
@@ -0,0 +1,113 @@
...
+            Console.WriteLine("\nSearch. request: \n\n" + searchRequestJson);
+            return searchRequest;
+        }
+
+        // Call the Retail Search:
+        [Attributes.Example]
+        public static IEnumerable<SearchResponse> Search()
+        {
+            //TRY DIFFERENT FILTER EXPRESSIONS HERE:
+            string filter = "(colorFamily: ANY(\"Black\"))";
```
chenlei1216, line 76: colorFamilies

retail/api/grs-search/SearchSimpleQuery.cs
```diff
@@ -0,0 +1,110 @@
...
+
+        // Call the Retail Search:
+        [Attributes.Example]
+        public static IEnumerable<SearchResponse> Search()
+        {
+            // TRY DIFFERENT QUERY PHRASES HERE:
+            var query = "Hoodie"; 
+
+            var searchRequest = GetSearchRequest(query);
+            var searchResponse = GetSearchServiceClient().Search(searchRequest).AsRawResponses();
```
chenlei1216, line 76: Just wondering if it's using auto pagination here with the library? Will this only pull the first page, or all pages?

retail/api/grs-search/SearchWithQueryExpansion.cs
chenlei1216: We should be able to omit the Endpoint. Feel free to try it and I think it should work.

retail/api/grs-search/SearchWithQueryExpansion.cs
chenlei1216: Should we remove the comment part?

vasilypascal: @chenlei1216 yes, we will remove it.

retail/api/grs-search/SearchWithQueryExpansion.cs
vasilypascal: Actually, this is just the result cost of the refactoring. It will be removed.

retail/api/grs-events/ImportUserEventsBigQuery.cs
```diff
@@ -0,0 +1,107 @@
...
+            var importBigQueryRequest = GetImportUserEventsBigQueryRequest();
+            var importResponse = GetUserEventsServiceClient().ImportUserEvents(importBigQueryRequest);
+
+            Console.WriteLine("\nThe operation was started: \n" + importResponse.Name);
+            Console.WriteLine("\nPlease wait till opeartion is done\n");
+
+            var importResult = importResponse.PollUntilCompleted();
+
+            Console.WriteLine("Import user events operation is done");
+            Console.WriteLine("Number of successfully imported events: " + importResult.Metadata.SuccessCount);
```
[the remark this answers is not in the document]
vasilypascal: Agree, will improve this.

Issue #1583, opened 2022-01-09
Backup Exceptions


tellmeaboutcode:
I am seeing the below exception while creating the backups.

`Status(StatusCode="InvalidArgument", Detail="Invalid ListBackupOperations request.", DebugException="Grpc.Core.Internal.CoreErrorDetailException: {"created":"@1641686430.749211566","description":"Error received from peer ipv4:142.250.188.42:443","file":"/var/local/git/grpc/src/core/lib/surface/call.cc","file_line":1067,"grpc_message":"Invalid ListBackupOperations request.","grpc_status":3}")`

```
var filter =
                $"(metadata.database={dbName}) AND " +
                "(metadata.@type:type.googleapis.com/google.spanner.admin.database.v1.CreateBackupMetadata) AND " +
                "done=false";

```

amanda-tarafa:
I've seen an uptick of these errors recently on our CI as well. I'm planning to look into this today.

tellmeaboutcode:
@amanda-tarafa Hi any update on this?

amanda-tarafa:
No, I'm sorry, not yet, I'll update here as soon as I know more.

amanda-tarafa:
OK, I can reproduce this also on my local environment and there's nothing inmmediately wrong with the sample code on this repo or with the client library itself.

The problem seems to be that the API has stopped accepting `metadata.database` as part of the filter. As a very temporary workaround, consider filtering for the Database locally?

In the meantime I'll reach out to the API team. @skuruppu FYI, I'm adding you as assignee as well in case you can take a look on the API side of things.

tellmeaboutcode:
@amanda-tarafa Thanks for working on this. Could you please explain/provide a sample code as to how to go about the temporary solution mentioned above?

amanda-tarafa:
@tellmeaboutcode we know more of what is happening, and the issue seems to be with the API requiring filter predicates to be in a particular order, so, if you modify your filter to what follows, it should work.

We are still looking at whether this API change should/can be reverted, etc. But for the time being, just rewrite your filter as a workaround.

```
$"(metadata.@type:type.googleapis.com/google.spanner.admin.database.v1.CreateBackupMetadata) AND " +
"(metadata.database={dbName}) AND " +
"(done=false)";
```

amanda-tarafa:
[...]

PR #1589, opened 2022-01-12
tests(Spanner): Deactivates test for listing backup operations.

amanda-tarafa:
Seems the API has stopped acceting metadata.database as part of the filer. We are looking into it. See #1583


jskeet (approved):

amanda-tarafa:
Errors are unrelated, actually a transient server side retriable error in the Monitoring test suite. Merging...

PR #1620, opened 2022-02-09
chore: Remove dead code from CreateSink sample

arbrown:
Remove extraneous sinkRequest references based on a different method of creating a Sink than the one being demonstrated.


amanda-tarafa (approved):
LGTM

PR #1635, opened 2022-02-17
Flaky copy file

anshuldavid13:


amanda-tarafa (approved):
I'm not certain this will fix the flakiness, as we were already using a specif  bucket, but it might. So let's try this and I'll reopen the issue if it happens again.

LGTM

Issue #1425, opened 2021-08-06
Spanner: Refactor the test fixture.


amanda-tarafa:
The existing test fixture is getting unmanageable in both the amount of time it takes to initialize, and the amount of unrelated code it's holding. Also, some decisions that were made at the beginning os Spanner test rewritting might have not been the best, in hindsight, and may be reconsider now.

The general proposal is to have several specialized fixtures, maybe all deriving from a base class.

Let's use this issue to discuss some options before starting any work.

@olavloite FYI (I can't add you as an assignee)

amanda-tarafa:
Some aspects that we might want to discuss (that I can think off right now)

- Resource deletion approach.
- Fixed resources (that we reuse for all tests run) vs. temporary resources.
- Test execution ordering.
- Frontier in which to split the current fixture.

My thoughts on these:

Resource deletion: Keeping resources after tests are done allows using them for diagnosing purposes when tests fail. But it also means that we need to delete stale resources before starting a new test run, and that if we are not very good at that, we hit quotas. Initializing test runs also takes longer because of all the deletions. I'm happy to reconsider this approach and delete temporary resources on test teardown, or maybe on fixture teardown.

Fixed vs. temporary resources: Thinking particularly of backups, but also of some DB operations. They take (or at least used to) a long time (some are still being skipped, see #1300 ), so if we can avoid creating temporary resources in those cases, that's preferable. The aim should be to keep running times as low as possible.

Test execution ordering: I'm still opposed to executing tests in a certain order, not just because of testing best practices, but also because there's no guarantee that users will execute samples in a given order. We need to make sure that samples work on their own.

Splitting the current fixture: Seems like the instance is a good place to do that, as in, each fixture works with a different DB instance? But this is just intuition.

olavloite:
In addition to the above, I would also like to add the following to the list of discussion points:

- Parallel execution of tests

### Resource deletion
While it can be useful to keep a resource around after a test run for diagnostic purposes, many of the sample tests are so simple that a failure will normally be easy to diagnose without the resource as well, especially if the single test is quick to run from a development machine. I would say that an approach that prefers deletion of resources after a run is preferable, unless the resource in question has some attributes that make it different from the most common resources, or if we know that it takes a long time to create. In the latter case, it might be more efficient to diagnose a test failure by looking at the resource that was left behind, than by rerunning the test.

[...]

PR #1661, opened 2022-03-18
feat:Spanner pg samples

Rishabh-V:
First draft of Spanner PostgreSQL dialect samples for feedback. The samples depend on v3.14.0 of C# Spanner client library which were created locally for sample creation and testing.

### DIFF
--- a/spanner/api/Spanner.Samples.Tests/CastDataTypesAsyncPostgreTest.cs
+++ b/spanner/api/Spanner.Samples.Tests/CastDataTypesAsyncPostgreTest.cs
@@ -1,0 +1,47 @@
+﻿// Copyright 2022 Google Inc.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+using System;
+using System.Threading.Tasks;
+using Xunit;
+
+[Collection(nameof(SpannerFixture))]
+public class CastDataTypesAsyncPostgreTest
+{
+    private readonly SpannerFixture _spannerFixture;
+
+    private readonly CastDataTypesAsyncPostgreSample _sample;
+
+    public CastDataTypesAsyncPostgreTest(SpannerFixture spannerFixture)
+    {
+        _spannerFixture = spannerFixture;
+        _sample = new CastDataTypesAsyncPostgreSample();
+    }
+
+    [Fact]
+    public async Task TestCastDataTypesAsyncPostgre()
+    {
+        // Act.
+        var result = await _sample.CastDataTypesAsyncPostgre(_spannerFixture.ProjectId, _spannerFixture.InstanceId, _spannerFixture.PostgreSqlDatabaseId);
+
+        //Assert.
+        Assert.Equal("1", result.String);
+        Assert.Equal(2, result.Integer);
+        Assert.Equal(3, result.Decimal);
+        Assert.Equal("34", BitConverter.ToString(result.Bytes));
+        Assert.Equal(5.00, result.Float);
+        Assert.True(result.Bool);
+        Assert.Equal("2021-11-03 09:35:01Z", result.Timestamp.ToString("u"));
+    }
+}
--- a/spanner/api/Spanner.Samples.Tests/CreateDatabaseAsyncPostgreTest.cs
+++ b/spanner/api/Spanner.Samples.Tests/CreateDatabaseAsyncPostgreTest.cs
@@ -1,0 +1,47 @@
+﻿// Copyright 2022 Google Inc.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+using Google.Cloud.Spanner.Admin.Database.V1;
+using System;
+using System.Linq;
+using System.Threading.Tasks;
+using Xunit;
+
+[Collection(nameof(SpannerFixture))]
+public class CreateDatabaseAsyncPostgreTest
+{
+    private readonly SpannerFixture _spannerFixture;
+
+    private readonly CreateDatabaseAsyncPostgreSample _sample;
+
+    public CreateDatabaseAsyncPostgreTest(SpannerFixture spannerFixture)
+    {
+        _spannerFixture = spannerFixture;
+        _sample = new CreateDatabaseAsyncPostgreSample();
+    }
+
+    [Fact]
+    public async Task TestCreateDatabaseAsyncPostgre()
+    {
+        // Arrange.
+        var databaseId = $"my-db-{DateTimeOffset.UtcNow.ToUnixTimeMilliseconds()}";
+
+        // Act.
+        await _sample.CreateDatabaseAsyncPostgre(_spannerFixture.ProjectId, _spannerFixture.InstanceId, databaseId);
+
+        // Assert.
+        var databases = _spannerFixture.GetDatabases();
+        Assert.Contains(databases, d => d.DatabaseName.DatabaseId == databaseId && d.DatabaseDialect == DatabaseDialect.Postgresql);
+    }
+}
--- a/spanner/api/Spanner.Samples.Tests/SpannerFixture.cs
+++ b/spanner/api/Spanner.Samples.Tests/SpannerFixture.cs
@@ -35,6 +35,7 @@ public class SpannerFixture : IAsyncLifetime, ICollectionFixture<SpannerFixture>
     // Allow environment variables to override the default instance and database names.
     public string InstanceId { get; } = Environment.GetEnvironmentVariable("TEST_SPANNER_INSTANCE") ?? "my-instance";
     public string DatabaseId { get; } = Environment.GetEnvironmentVariable("TEST_SPANNER_DATABASE") ?? $"my-db-{DateTimeOffset.UtcNow.ToUnixTimeMilliseconds()}";
+    public string PostgreSqlDatabaseId { get; } = Environment.GetEnvironmentVariable("TEST_SPANNER_POSTGRESQL_DATABASE") ?? $"my-db-postgre-{DateTimeOffset.UtcNow.ToUnixTimeMilliseconds()}";
     public string BackupDatabaseId { get; } = "my-test-database";
     public string BackupId { get; } = "my-test-database-backup";
     public string ToBeCancelledBackupId { get; } = $"my-backup-{DateTimeOffset.UtcNow.ToUnixTimeMilliseconds()}";
@@ -91,6 +92,7 @@ public class SpannerFixture : IAsyncLifetime, ICollectionFixture<SpannerFixture>
         await DeleteStaleInstancesAsync();
         await InitializeDatabaseAsync();
         await InitializeBackupAsync();
+        await InitializePostgreSqlDatabaseAsync();
 
         // Create encryption key for creating an encrypted database and optionally backing up and restoring an encrypted database.
         await InitializeEncryptionKeys();
@@ -383,6 +385,18 @@ public class SpannerFixture : IAsyncLifetime, ICollectionFixture<SpannerFixture>
             // It's ok backup has been in progress in another test cycle.
             Console.WriteLine($"Backup {BackupId} already in progress.");
         }
+    }
+
+    private async Task InitializePostgreSqlDatabaseAsync()
+    {
+        await CreatePostgreSqlDatabaseAsync(PostgreSqlDatabaseId);
+    }
+
+    public async Task CreatePostgreSqlDatabaseAsync(string databaseId)
+    {
+        // Create PostgreSQL database.
+        CreateDatabaseAsyncPostgreSample createDatabaseAsyncSample = new CreateDatabaseAsyncPostgreSample();
+        await createDatabaseAsyncSample.CreateDatabaseAsyncPostgre(ProjectId, InstanceId, databaseId);
     }
 
     public async Task CreateVenuesTableAndInsertDataAsync(string databaseId)
--- a/spanner/api/Spanner.Samples/CastDataTypesAsyncPostgre.cs
+++ b/spanner/api/Spanner.Samples/CastDataTypesAsyncPostgre.cs
@@ -1,0 +1,68 @@
+﻿// Copyright 2022 Google Inc.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+// [START spanner_postgresql_cast_data_type]
+
+using Google.Cloud.Spanner.Data;
+using System;
+using System.Threading.Tasks;
+
+public class CastDataTypesAsyncPostgreSample
+{
+    public async Task<CastDataType> CastDataTypesAsyncPostgre(string projectId, string instanceId, string databaseId)
+    {
+        string connectionString = $"Data Source=projects/{projectId}/instances/{instanceId}/databases/{databaseId}";
+
+        using var connection = new SpannerConnection(connectionString);
+        await connection.OpenAsync();
+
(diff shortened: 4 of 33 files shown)


spanner/api/Spanner.Samples.Tests/UpdatePostgreSqlUsingBatchDmlAsyncTest.cs
```diff
@@ -0,0 +1,37 @@
+﻿// Copyright 2022 Google Inc.
```
Rishabh-V, line 1: Please ignore this file from review. This is wrongly named and got checked-in in error.

spanner/api/Spanner.Samples/CreateDatabaseAsyncPostgre.cs
```diff
@@ -0,0 +1,94 @@
...
+using System;
+using System.Threading.Tasks;
+
+public class CreateDatabaseAsyncPostgreSample
+{
+    public async Task<Database> CreateDatabaseAsyncPostgre(string projectId, string instanceId, string databaseId)
+    {
+        DatabaseAdminClient databaseAdminClient = await new DatabaseAdminClientBuilder
+        {
+            // QuotaProject needs to be set for now while feature is in Pre-GA. May not be needed in final sample.
```
amanda-tarafa, line 28: I see it is, thanks! Since we won't merge this until the library has been released and the feature is GA, maybe remove the setting of QuotaProject. We'll run testst when we are ready to merge and if they fail we can add it back in. Just don't want to put out there something that's not needed.

spanner/api/Spanner.Samples.Tests/CastDataTypesAsyncPostgreTest.cs
```diff
@@ -0,0 +1,46 @@
...
+    }
+
+    [Fact]
+    public async Task TestCastDataTypesAsyncPostgre()
+    {
+        // Act.
+        var result = await _sample.CastDataTypesAsyncPostgre(_spannerFixture.ProjectId, _spannerFixture.InstanceId, _spannerFixture.PostgreSqlDatabaseId);
+        
+        //Assert.
+        Assert.Contains("String : 1", result);
```
amanda-tarafa, line 38: Yep, avoid these tests on a constructed string.
Either return a struct from your tests or a value, like the amount of records fetched, etc.

amanda-tarafa (changes requested):
Structure and overall looks good. Just a few comments.

spanner/api/Spanner.Samples.Tests/SpannerFixture.cs
```diff
@@ -100,6 +102,49 @@ public async Task InitializeAsync()
         }
     }
 
+    public async Task<List<string>> ListTableNamesAsync(string databaseId)
```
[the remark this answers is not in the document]
Rishabh-V: This is used in one of the samples. I will share that in PR in second batch

spanner/api/Spanner.Samples/CreateDatabaseAsyncPostgre.cs
```diff
@@ -0,0 +1,94 @@
...
+        if (completedResponse.IsFaulted)
+        {
+            Console.WriteLine($"Error while creating PostgreSQL database: {completedResponse.Exception}");
+            throw completedResponse.Exception;
+        }
+
+        // Postgres Database is created. Now, we can create the tables.
+        // Define create table statement for table #1 in PostgreSQL syntax.
+        var createSingersTable =
+        @"CREATE TABLE Singers (
```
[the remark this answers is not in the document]
Rishabh-V: Right Amanda, this is indeed a requirement. This is based on the equivalent Java sample which I got to know after discussing with the team.

spanner/api/Spanner.Samples.Tests/SpannerFixture.cs
```diff
@@ -388,6 +433,13 @@ private async Task InitializeBackupAsync()
         }
     }
 
+    private async Task InitializePostgreSqlDatabaseAsync()
```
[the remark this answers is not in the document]
Rishabh-V: Initialize Database creates a PG database as well as couple of tables namely Singers and Albums. There is another sample which depicts the ordering of NULL (Nulls First, Nulls Last etc.) that creates a different version of Singers table and inserted data is then ordered in different ways. To keep the sample consistent and inline with other languages and to avoid the chances of duplicate data being inserted, I also created a CreateEmptyPostgreSqlDatabase method. I will also share this sample in next batch.

Rishabh-V (commented):
Fixed review comments

spanner/api/Spanner.Samples/Spanner.Samples.csproj
```diff
@@ -1,11 +1,11 @@
-<Project Sdk="Microsoft.NET.Sdk">
```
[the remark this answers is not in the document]
amanda-tarafa: This is still here. Revert the changes on this whole file.

spanner/api/Spanner.Samples/CreateDatabaseAsyncPostgre.cs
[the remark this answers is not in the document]
amanda-tarafa: Acked. Thanks

spanner/api/Spanner.Samples/CreateDatabaseAsyncPostgre.cs
```diff
@@ -0,0 +1,86 @@
...
+        {
+            ParentAsInstanceName = InstanceName.FromProjectInstance(projectId, instanceId),
+            CreateStatement = $"CREATE DATABASE \"{databaseId}\"",
+            DatabaseDialect = DatabaseDialect.Postgresql
+        };
+
+        var createOperation = await databaseAdminClient.CreateDatabaseAsync(createDatabaseRequest);
+
+        // Wait until the operation has finished.
+        Console.WriteLine("Waiting for the operation to finish.");
```
amanda-tarafa, line 40: This is just a nit, but when you run this sample this text will be printed twice, once for creating the db and once for adding the tables. Consider distinguising the two as in "Waiting for the DB to be created" and "Waiting for the tables to be created."

Here and everywhere where there's something like this happening.

spanner/api/Spanner.Samples.Tests/CreateDatabaseAsyncPostgreTest.cs
```diff
@@ -0,0 +1,48 @@
...
+        // Arrange.
+        var databaseId = $"my-db-{DateTimeOffset.UtcNow.ToUnixTimeMilliseconds()}";
+
+        // Act.
+        await _sample.CreateDatabaseAsyncPostgre(_spannerFixture.ProjectId, _spannerFixture.InstanceId, databaseId);
+
+        // Assert.
+        var databases = _spannerFixture.GetDatabases();
+        Assert.Contains(databases, d => d.DatabaseName.DatabaseId == databaseId);
+        Assert.Equal(DatabaseDialect.Postgresql, databases.FirstOrDefault(j=> j.DatabaseName.DatabaseId == databaseId).DatabaseDialect);
```
amanda-tarafa, line 46:
```suggestion
        Assert.Contains(databases, d => d.DatabaseName.DatabaseId == databaseId && d.DatabaseDialect == ....);
databaseId).DatabaseDialect);
```

spanner/api/Spanner.Samples/ConnectToDatabaseAsyncPostgre.cs
```diff
@@ -0,0 +1,41 @@
...
+        var command = connection.CreateSelectCommand("SELECT 'Hello from Spanner PostgreSQL!'");
+
+        var result = await command.ExecuteScalarAsync();
+        if (result is string statement)
+        {
+            Console.WriteLine(statement);
+            return statement;
+        }
+
+        return string.Empty;
```
amanda-tarafa, line 38:
```suggestion
        var result = await command.ExecuteScalarAsync<string>();
        Console.WriteLine(result);
        return result;
```

spanner/api/Spanner.Samples/CastDataTypesAsyncPostgre.cs
```diff
@@ -0,0 +1,68 @@
...
+        public byte[] Bytes { get; set; }
+
+        public float Float { get; set; }
+
+        public bool Bool { get; set; }
+
+        public DateTime Timestamp { get; set; }
+    }
+}
+// [END spanner_postgresql_cast_data_type]
```
amanda-tarafa, line 68: By the way, you have this on all your PRs, we usually leave a blank line at the end of the file.
Not a problem though, just fix it as you go along.

spanner/api/Spanner.Samples.Tests/SpannerFixture.cs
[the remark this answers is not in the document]
amanda-tarafa: OK, so then should be included on the PR where the sample that uses it is. That way I can check the method and its usage, maybe suggest something else, etc. The same comment applies for the other methods that you are not using here but will use in another PR.

If it's not toooo much trouble, try to remove these unusued methods from this PR and add them on the subsequent ones. Else, no worries, but take it into account for next time.

spanner/api/Spanner.Samples.Tests/SpannerFixture.cs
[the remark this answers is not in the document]
amanda-tarafa: OK, so then CreateEmpty... should be on the PR that uses it if posssible.

amanda-tarafa (commented):
A few nits but otherwise looks good. Feel free to apply the changes here and create the PR containing the next batch (with relevant changes there as well acording to this PR's comments.)

spanner/api/Spanner.Samples.Tests/SpannerFixture.cs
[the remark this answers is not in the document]
Rishabh-V: Sure Amanda. Added the test cases which uses this method.

spanner/api/Spanner.Samples.Tests/SpannerFixture.cs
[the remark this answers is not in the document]
Rishabh-V: Done.

spanner/api/Spanner.Samples/Spanner.Samples.csproj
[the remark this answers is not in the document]
Rishabh-V: I reverted the entire file. However, still seeing it. I will seek your help to revert this tomorrow.

spanner/api/Spanner.Samples/CreateDatabaseAsyncPostgre.cs
Rishabh-V: Done.

spanner/api/Spanner.Samples/IdentifierCaseSensitivityAsyncPostgre.cs
```diff
@@ -0,0 +1,127 @@
...
+        // DML statements must also follow the PostgreSQL case rules.
+        var dmlCommand = connection.CreateDmlCommand("INSERT INTO Concerts (ConcertId, \"Location\", \"Time\") VALUES($1, $2, $3)",
+            new SpannerParameterCollection
+            {
+                { "p1", SpannerDbType.Int64, 2 },
+                { "p2", SpannerDbType.String, "Venue 2" },
+                { "p3", SpannerDbType.Timestamp, DateTime.UtcNow }
+            });
+
+        var count = await dmlCommand.ExecuteNonQueryAsync();
```
amanda-tarafa, line 123: This is one of the main problems with having samples that do many things at once. We are barely testing this sample.

spanner/api/Spanner.Samples/GetInformationSchemaAsyncPostgre.cs
```diff
@@ -0,0 +1,73 @@
...
+public class GetInformationSchemaAsyncPostgreSample
+{
+    public async Task<List<InformationSchema>> GetInformationSchemaAsyncPostgre(string projectId, string instanceId, string databaseId)
+    {
+        string connectionString = $"Data Source=projects/{projectId}/instances/{instanceId}/databases/{databaseId}";
+
+        using var connection = new SpannerConnection(connectionString);
+        await connection.OpenAsync();
+
+        // The Spanner INFORMATION_SCHEMA tables can be used to query the metadata of tables and columns
```
amanda-tarafa, line 29: Again, something that should be in docs and not on each of the language specific samples.

spanner/api/Spanner.Samples/GetInformationSchemaAsyncPostgre.cs
```diff
@@ -0,0 +1,73 @@
...
+        {
+            var informationSchema = new InformationSchema
+            {
+                Schema = reader.GetFieldValue<string>("table_schema"),
+                Name = reader.GetFieldValue<string>("table_name")
+            };
+
+            var userDefinedTypeCatalog = $"{reader.GetFieldValue<object>("user_defined_type_catalog")}";
+            var userDefinedTypeSchema = $"{reader.GetFieldValue<object>("user_defined_type_schema")}";
+            var userDefinedTypeName = $"{reader.GetFieldValue<object>("user_defined_type_name")}";
```
amanda-tarafa, line 52: What are the real types of these and why are you converting them to string?
I guess users would prefer to deal with something typed than with a string.

spanner/api/Spanner.Samples/GetInformationSchemaAsyncPostgre.cs
```diff
@@ -0,0 +1,73 @@
...
+
+            informationSchema.UserDefinedType =
+                string.IsNullOrWhiteSpace(userDefinedTypeName) ? "null"
```
amanda-tarafa, line 55: Really don't use "null" or "NULL", that means nothing. Use `null`. If you absolutely need to provide a string value better use "unknown" or "undefined" as that's less prone to confusion. But in my view even that should only be used on a very explicit string representation like when writing to the Console or similar, never when holding the data.

spanner/api/Spanner.Samples/ExecutePartitionedDmlAsyncPostgre.cs
```diff
@@ -0,0 +1,43 @@
...
+        // maximum of 20,000 mutations in a single read/write transaction. Large update operations can
+        // be executed using Partitioned DML. This is also supported on Spanner PostgreSQL.
+        // See https://cloud.google.com/spanner/docs/dml-partitioned for more information.
+
+        using var cmd = connection.CreateDmlCommand("DELETE FROM Singers WHERE SingerId > 3");
+
+        // The returned count is the lower bound of the number of records that was deleted.
+        long rowCount = await cmd.ExecutePartitionedUpdateAsync();
+
+        Console.WriteLine($"{rowCount} row(s) deleted...");
```
amanda-tarafa, line 39:
```suggestion
        Console.WriteLine($"At least {rowCount} row(s) deleted...");
```

amanda-tarafa (changes requested):
Apart from the specific change requests I have several PR wide comments.

1. There's a lot of information in comments inside the samples that should really be part of documentation sorrounding the sample instead. In that way, it doesn't have to be repeated across languages and it is maintained by TWs. Comments in samples should explain something particular of the language, i.e. something that is just needed for the .NET library for instance; and maybe small notes making reference to documentation, but not the whole explanation.
2. A lot of samples do many (related) things and we tend to prefer simpler samples. The current samples are harder to test and possibly harder for users to figure out (for instance the ORDER BY, the name Identifier samples, etc.)
3. Some samples create database and tables, some samples create tables and some samples asume everything they need is already created. We need consistency here and also sample simplification. If someone wants to see how to execute a command they probably don't need the whole create database => create table => execute command. If some samples are creating DBs to avoid flakyness, then those should be created in the corresponding tests but not on the samples themselves.
4. It's not clear to me that all samples and samples tests that are creating DBs are doing so because of flakyness. We are already blowing Spanner quota ocasionally, if we continut to create DBs without needing them these tests will never pass. Take a look at #1425. In addition consider creating a separate Fixture for PG.
5. See if you can reuse samples for the Spanner dialect, all of those that are the same, simply add the PG region tag on the existing one.

I know some of these issues are not because decisions you have made, but we should discus them nonetheless. Let's chat tomorrow.

spanner/api/Spanner.Samples.Tests/SpannerFixture.cs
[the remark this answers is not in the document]
Rishabh-V: `CreateEmptyPostgreSqlDatabaseAsync` and `ListTableNamesAsync` are now removed from fixture.

Rishabh-V (commented):
Fixed review comments. The following changes are now done:
1. Removed comments that should be in documentation. Have kept only short summary now.
2. Refactored tests and samples, so that  new database is not created. To avoid data insertion conflicts, distinct data specific to the test scenario is inserted in tests as needed.

amanda-tarafa (approved):
Thanks Rishabh! These look good. LGTM

I'm marking this PR as `do not merge` just because we need production changes released first.

amanda-tarafa:
@Rishabh-V  I think we might need to increase the overall timeout for Spanner samples. You can do that in the `runTests.ps1` file. Do include that change as part of this PR.

amanda-tarafa:
> I'm marking this PR as do not merge just because we need production changes released first.

Is my assumption right that we still need to release some library changes or has everything been released already.

Rishabh-V:
> Thanks Rishabh! These look good. LGTM
> 
> I'm marking this PR as `do not merge` just because we need production changes released first.

Thanks Amanda!

> @Rishabh-V I think we might need to increase the overall timeout for Spanner samples. You can do that in the `runTests.ps1` file. Do include that change as part of this PR.

Sure Amanda. Timeout is currently set to 7200 seconds.  Should I increase it to 9000 seconds?

> > I'm marking this PR as do not merge just because we need production changes released first.
> 
> Is my assumption right that we still need to release some library changes or has everything been released already.

Right Amanda. I believe PgNumeric is not released yet.

amanda-tarafa:
Let's wait for the timeout until we can actually build the PR, as right now it is failing because it's missing PgNumeric as expected.

So, nothing to do for now. Thanks!

PR #1694, opened 2022-04-15
feat(cloud-sql/mysql): update to V2 sample

jsimonweb:
- Move TCP and UNIX socket examples into separate files
- Update region tags
- Update DB_HOST env. variable to be INSTANCE_HOST
- Update DB_SOCKET_PATH env. variable to be INSTANCE_UNIX_SOCKET

### DIFF
--- a/cloud-sql/mysql/ConnectTcp.cs
+++ b/cloud-sql/mysql/ConnectTcp.cs
@@ -31,9 +31,10 @@ namespace CloudSql
                 // The Cloud SQL proxy provides encryption between the proxy and instance.
                 SslMode = MySqlSslMode.None,
 
-                // Remember - storing secrets in plain text is potentially unsafe. Consider using
-                // something like https://cloud.google.com/secret-manager/docs/overview to help keep
-                // secrets secret.
+                // Note: Saving credentials in environment variables is convenient, but not
+                // secure - consider a more secure solution such as
+                // Cloud Secret Manager (https://cloud.google.com/secret-manager) to help
+                // keep secrets safe.
                 Server = Environment.GetEnvironmentVariable("INSTANCE_HOST"),   // e.g. '127.0.0.1'
                 // Set Host to 'cloudsql' when deploying to App Engine Flexible environment
                 UserID = Environment.GetEnvironmentVariable("DB_USER"),   // e.g. 'my-db-user'
--- a/cloud-sql/mysql/ConnectTcp.cs
+++ b/cloud-sql/mysql/ConnectTcp.cs
@@ -1,0 +1,50 @@
+/*
+ * Copyright (c) 2022 Google LLC.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License"); you may not
+ * use this file except in compliance with the License. You may obtain a copy of
+ * the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS, WITHOUT
+ * WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied. See the
+ * License for the specific language governing permissions and limitations under
+ * the License.
+ */
+
+// [START cloud_sql_mysql_dotnet_ado_connect_tcp]
+using MySql.Data.MySqlClient;
+using System;
+
+namespace CloudSql
+{
+    public class MySqlTcp
+    {
+        public static MySqlConnectionStringBuilder NewMysqlTCPConnectionString()
+        {
+            // Equivalent connection string:
+            // "Uid=<DB_USER>;Pwd=<DB_PASS>;Host=<INSTANCE_HOST>;Database=<DB_NAME>;"
+            var connectionString = new MySqlConnectionStringBuilder()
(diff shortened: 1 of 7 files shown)


cloud-sql/mysql/ConnectTcp.cs
```diff
@@ -0,0 +1,50 @@
...
+        public static MySqlConnectionStringBuilder NewMysqlTCPConnectionString()
+        {
+            // Equivalent connection string:
+            // "Uid=<DB_USER>;Pwd=<DB_PASS>;Host=<INSTANCE_HOST>;Database=<DB_NAME>;"
+            var connectionString = new MySqlConnectionStringBuilder()
+            {
+                // The Cloud SQL proxy provides encryption between the proxy and instance.
+                SslMode = MySqlSslMode.None,
+
+                // Remember - storing secrets in plain text is potentially unsafe. Consider using
```
enocom, line 34: Let's update the text of this comment to match the Go sample:

```
// Note: Saving credentials in environment variables is convenient, but not
// secure - consider a more secure solution such as
// Cloud Secret Manager (https://cloud.google.com/secret-manager) to help
// keep secrets safe.
```

cloud-sql/mysql/ConnectUnix.cs
```diff
@@ -0,0 +1,50 @@
...
+        {
+            // Equivalent connection string:
+            // "Server=<dbSocketDir>/<INSTANCE_CONNECTION_NAME>;Uid=<DB_USER>;Pwd=<DB_PASS>;Database=<DB_NAME>;Protocol=unix"
+            String dbSocketDir = Environment.GetEnvironmentVariable("INSTANCE_UNIX_SOCKET") ?? "/cloudsql";
+            String instanceConnectionName = Environment.GetEnvironmentVariable("INSTANCE_CONNECTION_NAME");
+            var connectionString = new MySqlConnectionStringBuilder()
+            {
+                // The Cloud SQL proxy provides encryption between the proxy and instance.
+                SslMode = MySqlSslMode.None,
+                // Remember - storing secrets in plain text is potentially unsafe. Consider using
```
enocom, line 35: Ditto above.

cloud-sql/mysql/ConnectUnix.cs
```diff
@@ -0,0 +1,50 @@
...
+
+namespace CloudSql
+{
+    public class MySqlUnix
+    {
+        public static MySqlConnectionStringBuilder NewMysqlUnixSocketConnectionString()
+        {
+            // Equivalent connection string:
+            // "Server=<dbSocketDir>/<INSTANCE_CONNECTION_NAME>;Uid=<DB_USER>;Pwd=<DB_PASS>;Database=<DB_NAME>;Protocol=unix"
+            String dbSocketDir = Environment.GetEnvironmentVariable("INSTANCE_UNIX_SOCKET") ?? "/cloudsql";
```
enocom, line 29: Let's move these down below. Also, "INSTANCE_UNIX_SOCKET" should be the full path and so we should remove `/cloudsql`.

cloud-sql/mysql/Startup.cs
```diff
@@ -112,13 +112,13 @@ PRIMARY KEY (vote_id)
         public static MySqlConnectionStringBuilder GetMySqlConnectionString()
         {
             MySqlConnectionStringBuilder connectionString; 
-            if (Environment.GetEnvironmentVariable("DB_HOST") != null)
+            if (Environment.GetEnvironmentVariable("INSTANCE_HOST") != null)
             {
-                connectionString = NewMysqlTCPConnectionString();
+                connectionString = MySqlTcp.NewMysqlTCPConnectionString();
```
enocom, line 117: Is `return` idiomatic here instead of if/else?

enocom (changes requested):

enocom:
@jsimonweb Did you forget to push some changes?

cloud-sql/mysql/ConnectTcp.cs
jsimonweb: Done.

cloud-sql/mysql/ConnectUnix.cs
jsimonweb: Done.

cloud-sql/mysql/ConnectUnix.cs
jsimonweb: Done.

cloud-sql/mysql/Startup.cs
jsimonweb: Defer to amanda-tarafa@ or jskeet@ on this one.

enocom (approved):

cloud-sql/mysql/Startup.cs
amanda-tarafa: Since we are using `connectionString` to set some properties after it was initialized (see lines >= 127) we need to do `if...else` here.

If this method was inmediately returning after initializing `connectionString` then the most idiomatic thing to do is probably:

```chsarp
return Environment.GetEnvironmentVariable("INSTANCE_HOST") is null ? MySqlUnix. ... : MySqlTcp. ...;
```

amanda-tarafa (approved):
LGTM

Issue #1548, opened 2021-11-10
Cloud-SQL: Consider updating samples to netcoreapp3.1


amanda-tarafa:
As [netcoreapp2.1 has reached end of life](https://dotnet.microsoft.com/download/dotnet/2.1).

amanda-tarafa:
Closed by #1683

Issue #1004, opened 2020-04-01
AppEngine Flex: Cleanup and update dependencies.


amanda-tarafa:
Leftover from #988 , there's a lot of work to do, possibly redo some of the applications.

amanda-tarafa:
@jahjahbin I think we can close this one right?

Thanks for bringing the samples up to date.

PR #1701, opened 2022-04-19
Several improvements

amanda-tarafa:


sethvargo (approved):

.kokoro-windows/New-BuildEnv.ps1
```diff
@@ -22,36 +22,41 @@ get-command choco -ErrorAction Stop
 choco list -li | Tee-Object -Variable chocoList
 $chocoPackages = ($chocoList) -join ' '
 
-if (-not $chocoPackages.Contains('Microsoft .NET Core SDK - 2.1.4')) {
-    choco install -y --sxs dotnetcore-sdk --version 2.2.203
+if (-not $chocoPackages.Contains('Microsoft .NET Core SDK - 2.2.')) {
```
jskeet, line 25: Not sure we should be using 2.2 at all - that went out of support even longer ago than 2.1.

Can we get rid of it entirely?

spanner/api/Spanner.Samples.Tests/GetDatabaseDdlAsyncTest.cs
```diff
@@ -48,9 +48,9 @@ public async Task TestGetDatabaseDdlAsync()
         await _spannerFixture.RunWithTemporaryDatabaseAsync(_spannerFixture.InstanceIdWithMultiRegion, databaseId, async databaseId =>
         {
             var sample = new GetDatabaseDdlAsyncSample();
-            var statements =
+            var response =
                 await sample.GetDatabaseDdlAsync(_spannerFixture.ProjectId, _spannerFixture.InstanceIdWithMultiRegion, databaseId);
-            Assert.Collection(statements,
+            Assert.Collection(response.Statements,
```
jskeet, line 53: I don't really understand how this change works - I'd have expected it to fail before. But hey, if it passes, that's fine.

spanner/api/Spanner.Samples.Tests/ListBackupOperationsTest.cs
```diff
@@ -25,11 +24,12 @@ public ListBackupOperationsTest(SpannerFixture spannerFixture)
...
 
-    [Fact(Skip = "https://github.com/GoogleCloudPlatform/dotnet-docs-samples/issues/1583")]
+    [Fact]
     public void TestListBackupOperations()
     {
         ListBackupOperationsSample getBackupOperationsSample = new ListBackupOperationsSample();
-        var operations = getBackupOperationsSample.ListBackupOperations(_spannerFixture.ProjectId, _spannerFixture.InstanceId, _spannerFixture.BackupDatabaseId).ToList();
-        Assert.True(operations.Count >= 0);
+        var operations = getBackupOperationsSample.ListBackupOperations(_spannerFixture.ProjectId, _spannerFixture.InstanceId, _spannerFixture.BackupDatabaseId);
+        // We iterated over operation on the sample itself, let's not do that again.
```
jskeet, line 32: Is this actually testing anything useful then?

jskeet: (Same question for tests below.)

jskeet (approved):
A few questions, and I don't know enough about the Spanner tests to know whether the changes here reduce our confidence that the tests will catch problems, but otherwise fine.

.kokoro-windows/New-BuildEnv.ps1
amanda-tarafa: There are like 10 sample applications that are using 2.2. Those are all web application so moving them away from 2.2 takes time.
We have:

- #1548 
- #1004

spanner/api/Spanner.Samples.Tests/GetDatabaseDdlAsyncTest.cs
amanda-tarafa: The sample method was previously returning the statements only, I made it return the response object.

spanner/api/Spanner.Samples.Tests/ListBackupOperationsTest.cs
amanda-tarafa: Just that the method is returning. I had it at Assert.NotEmpty but it still reached quota.

I could just call the method and have the test green if the method does not fail, but I'm hoping we can do better in the future so I left everything ready.

amanda-tarafa (commented):
All questions answered.

PR #1702, opened 2022-04-21
Added importing, tests and setups

vasilypascal:
Added:
* Import Events/Products BQ Table + test
* Import Events/Product Gcs Bucket + tests
* Import Events/Products inline source + test

* setup files for Creating and Removing BQ tables and Gcs Buckets
* resources: table schema and table data


mkoshurnikov:
Can be closed because it's outdated.

PR #1707, opened 2022-04-22
List backup

anshuldavid13:
- Fixing filter for ListBackupOperations
- Removed skip for ListBackupOperationsTest
- Added ListCopyBackupOperations


spanner/api/Spanner.Samples.Tests/ListCopyBackupOperationsTest.cs
```diff
@@ -0,0 +1,36 @@
...
+        _spannerFixture = spannerFixture;
+    }
+
+    [Fact]
+    public void TestListCopyBackupOperations()
+    {
+        ListCopyBackupOperationsSample listCopyBackupOperationsSample = new ListCopyBackupOperationsSample();
+
+        var operations = listCopyBackupOperationsSample.ListCopyBackupOperations(_spannerFixture.ProjectId, _spannerFixture.InstanceId, _spannerFixture.BackupDatabaseId,_spannerFixture.BackupId).ToList();
+        Assert.True(operations.Count >= 0);
```
amanda-tarafa, line 34: Please change this Assert to what I've done in https://github.com/GoogleCloudPlatform/dotnet-docs-samples/blob/main/spanner/api/Spanner.Samples.Tests/ListBackupOperationsTest.cs . This is to alleviate the quoata errors we were getting.

amanda-tarafa (changes requested):
A small change request.

And also, your branch history is wrong, I think you need to do a rebase.

PR #1717, opened 2022-04-27
feat: support customer managed instance configurations

AlexandrTrf:
Adds samples for working with user managed instance configurations for Cloud Spanner.


amanda-tarafa:
I'll review by mid week next week. Thanks.

AlexandrTrf:
> I hadn't realized this was a draft, @AlexandrTrf can you confirm if this is ready for review or not? Thanks.

@amanda-tarafa this code is ready. But it uses Google.Cloud.Scanner libraries that have not yet been released and now this code will not be able to be buit and the tests will not be passed.

spanner/api/Spanner.Samples.Tests/SpannerFixture.cs
```diff
@@ -37,6 +37,7 @@ public class SpannerFixture : IAsyncLifetime, ICollectionFixture<SpannerFixture>
     public string DatabaseId { get; } = Environment.GetEnvironmentVariable("TEST_SPANNER_DATABASE") ?? $"my-db-{DateTimeOffset.UtcNow.ToUnixTimeMilliseconds()}";
     public string BackupDatabaseId { get; } = "my-test-database";
     public string BackupId { get; } = "my-test-database-backup";
+    public string CustomInstanceConfigId { get; } = "custom-nam3";
```
amanda-tarafa, line 40: You need to use this as a name prefix as instance configs are not database specific, so all tests running on the same project will attempt to use the same `custom-nam3` config. Whenver you need to do some crud with a config you need to append a timestamp or a GUID to this prefix, and you'll need to clean configs up on dispose the same way we clean databases, backups and other resources up.

Else all of these tests are guaranteed to be very flaky and/or run into quota issues.

spanner/api/Spanner.Samples/ListInstanceConfigs.cs
```diff
@@ -27,7 +27,7 @@ public IEnumerable<InstanceConfig> ListInstanceConfigs(string projectId)
         var instanceAdminClient = InstanceAdminClient.Create();
         var projectName = ProjectName.FromProject(projectId);
         var instanceConfigs = instanceAdminClient.ListInstanceConfigs(projectName);
-
+
```
amanda-tarafa, line 30: nit: Undo this white space change

spanner/api/Spanner.Samples/ListInstanceConfigOperations.cs
```diff
@@ -0,0 +1,42 @@
...
+{
+    public class ListInstanceConfigOperationsSample
+    {
+        public IEnumerable<Operation> ListInstanceConfigOperations(string projectId)
+        {
+            var instanceAdminClient = InstanceAdminClient.Create();
+            var projectName = ProjectName.FromProject(projectId);
+            var listInstanceConfigOperations = instanceAdminClient.ListInstanceConfigOperations(projectName);
+
+            foreach (var operation in listInstanceConfigOperations)
```
amanda-tarafa, line 33: Please only print 5 elements, the same as in other listing examples, to make sure that we are not hitting the server several times, we were reaching admin quota very easily.

See this commit: https://github.com/GoogleCloudPlatform/dotnet-docs-samples/commit/35baed2aa2a6491523cebfa9d301750088af13b0

spanner/api/Spanner.Samples.Tests/ListInstanceConfigOperationsTest.cs
```diff
@@ -0,0 +1,36 @@
...
+        {
+            _spannerFixture = spannerFixture;
+        }
+
+        [Fact]
+        public void TestListInstanceConfigOperations()
+        {
+            var sample = new ListInstanceConfigOperationsSample();
+            var configs = sample.ListInstanceConfigOperations(_spannerFixture.ProjectId);
+            Assert.NotEmpty(configs);
```
amanda-tarafa, line 33:
```suggestion
            Assert.NotNull(configs);
```
Same reason as my other comment, we were hitting admin quoata, see https://github.com/GoogleCloudPlatform/dotnet-docs-samples/commit/35baed2aa2a6491523cebfa9d301750088af13b0

amanda-tarafa (changes requested):
I've made some requests that will apply to all samples and/or tests.

AlexandrTrf:
@amanda-tarafa I have made the necessary changes. Could you do a review?

jskeet:
@AlexandrTrf: Amanda is on vacation until next week.

spanner/api/Spanner.Samples/ListInstanceConfigsAsync.cs
```diff
@@ -19,12 +19,13 @@
 using System;
 using System.Collections.Generic;
 using System.Linq;
+using System.Threading.Tasks;
 
-public class ListInstanceConfigsSample
+public class ListInstanceConfigsAsyncSample
```
amanda-tarafa, line 24: Changing to creating the client async is fine, but you cannot change the file name because the existing sample was already included in docs.

Either we keep the file name incosistent (I'm fine with that in this instance) or you coordinate with the Spanner technical writers to change the docs to start using the new file name.

spanner/api/Spanner.Samples/ListInstanceConfigsAsync.cs
```diff
@@ -19,12 +19,13 @@
...
 {
-    public IEnumerable<InstanceConfig> ListInstanceConfigs(string projectId)
+    public async Task<IEnumerable<InstanceConfig>> ListInstanceConfigsAsync(string projectId)
     {
-        var instanceAdminClient = InstanceAdminClient.Create();
+        var instanceAdminClient = await InstanceAdminClient.CreateAsync();
         var projectName = ProjectName.FromProject(projectId);
         var instanceConfigs = instanceAdminClient.ListInstanceConfigs(projectName);
```
amanda-tarafa, line 30: If we are making the method Async, I think this might have an async version. Can you please double check?

spanner/api/Spanner.Samples/UpdateInstanceConfigAsync.cs
```diff
@@ -0,0 +1,61 @@
...
+            var updateInstanceConfigOperation = await instanceAdminClient.UpdateInstanceConfigAsync(new UpdateInstanceConfigRequest
+            {
+                InstanceConfig = instanceConfig,
+                UpdateMask = new FieldMask
+                {
+                    Paths = { "display_name", "labels" }
+                }
+            });
+
+            Console.WriteLine($"Update Instance Config operation is completed  - {updateInstanceConfigOperation.IsCompleted} successfully - {!updateInstanceConfigOperation.IsFaulted}");
```
amanda-tarafa, line 52: You might want to show polling until completion. Also the text printed will be a little confusing. Maybe be more explicit or even throw.

```suggestion
            updateInstanceConfigOperation = updateInstanceConfigOperation.PollUntilCompleted();
            if (updateInstanceConfigOperation.IsFaulted)
            {
                throw updateInstanceConfigOperation.Exception;
            }
            Console.WriteLine("Update Instance Config operation completed successfully.");
```

spanner/api/Spanner.Samples/UpdateInstanceConfigAsync.cs
```diff
@@ -0,0 +1,61 @@
...
+using System;
+using System.Collections.Generic;
+using System.Threading.Tasks;
+using Google.Cloud.Spanner.Admin.Instance.V1;
+using Google.Protobuf.WellKnownTypes;
+using Grpc.Core;
+
+public class UpdateInstanceConfigAsyncSample
+{
+    public async Task UpdateInstanceConfigAsync(string projectId, string instanceConfigId)
```
amanda-tarafa, line 26: You may return the operation here and check that is successful on the test, for instance.

spanner/api/Spanner.Samples.Tests/DeleteInstanceConfigAsyncTest.cs
```diff
@@ -0,0 +1,37 @@
...
+        public DeleteInstanceConfigAsyncTest(SpannerFixture spannerFixture)
+        {
+            _spannerFixture = spannerFixture;
+        }
+
+        [Fact]
+        public async Task TestDeleteInstanceConfigAsync()
+        {
+            DeleteInstanceConfigAsyncSample deleteInstanceConfigAsyncSample = new DeleteInstanceConfigAsyncSample();
+            await deleteInstanceConfigAsyncSample.DeleteInstanceConfigAsync(_spannerFixture.ProjectId, _spannerFixture.CustomInstanceConfigId);
```
amanda-tarafa, line 34: If the delete config test runs before the update config test, then the update config test won't work, i.e. won't do anything I know it won't fail because you have a try catch but still we won't test anything.

I prefer if you create a temporary config here to use for deletion.

spanner/api/Spanner.Samples.Tests/CreateInstanceConfigAsyncTest.cs
```diff
@@ -0,0 +1,43 @@
...
+            _spannerFixture = spannerFixture;
+        }
+
+        [Fact]
+        public async Task TestCreateInstanceConfigAsync()
+        {
+            CreateInstanceConfigAsyncSample createInstanceConfigAsyncSample = new CreateInstanceConfigAsyncSample();
+
+            var instanceConfig = await createInstanceConfigAsyncSample.CreateInstanceConfigAsync(_spannerFixture.ProjectId,
+                _spannerFixture.InstanceConfigId, _spannerFixture.CustomInstanceConfigId);
```
amanda-tarafa, line 38: Similar comment here, if this test executes last, then update and delete won't be able to do anything.
I think it's fine if we just have independen configs for each test.

Or is there something I'm missing.

spanner/api/Spanner.Samples.Tests/InstanceConfigPropertiesTest.cs
```diff
@@ -0,0 +1,37 @@
...
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+using Google.Cloud.Spanner.Admin.Instance.V1;
+using Xunit;
+
+namespace Spanner.Samples.Tests
+{
+    public class InstanceConfigPropertiesTest
```
amanda-tarafa, line 20: We don't need these tests here, this is a given from Protobuf. You can assume this will always work.

amanda-tarafa (changes requested):

spanner/api/Spanner.Samples/ListInstanceConfigsAsync.cs
AlexandrTrf: Here is my suggestion. 
Can I add the System.Linq.Parallel package to the project in order to print only 5 configurations to the console?

```suggestion
                var instanceConfigs = instanceAdminClient.ListInstanceConfigsAsync(projectName);
        
        // We print the first 5 elements for demonstration purposes.
        // You can print all configs in the sequence by removing the call to Take(5).
        // The sequence will lazily fetch elements in pages as needed.
        await foreach(var instanceConfig in instanceConfigs.Take(5))
        {
            Console.WriteLine($"Available leader options for instance config {instanceConfig.InstanceConfigName.InstanceConfigId}:");
            foreach (var leader in instanceConfig.LeaderOptions)
            {
                Console.WriteLine(leader);
            }

            Console.WriteLine($"Available optional replica for instance config {instanceConfig.InstanceConfigName.InstanceConfigId}:");
            foreach (var optionalReplica in instanceConfig.OptionalReplicas)
            {
                Console.WriteLine($"Replica type - {optionalReplica.Type}, default leader location - {optionalReplica.DefaultLeaderLocation}, location - {optionalReplica.Location}");
            }
        }

        return instanceConfigs.ToEnumerable();
```

amanda-tarafa (commented):
A couple of notes on the Async change.

Also let me know if/when the rest is done to re-review all. Thanks

spanner/api/Spanner.Samples/ListInstanceConfigs.cs
```diff
@@ -19,27 +19,35 @@
...
         var projectName = ProjectName.FromProject(projectId);
-        var instanceConfigs = instanceAdminClient.ListInstanceConfigs(projectName);
-
+        var instanceConfigs = instanceAdminClient.ListInstanceConfigsAsync(projectName);
+        
         // We print the first 5 elements for demonstration purposes.
         // You can print all configs in the sequence by removing the call to Take(5).
         // The sequence will lazily fetch elements in pages as needed.
-        foreach (var instanceConfig in instanceConfigs.Take(5))
+        await foreach(var instanceConfig in instanceConfigs.Take(5))
```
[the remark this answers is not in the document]
AlexandrTrf: But otherwise I won 't have an extension method .Take for AsyncEnumerable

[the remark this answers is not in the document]
amanda-tarafa: Ah yes, that's right, sorry for missing that.

Hmmm, then I think we can keep that part of the method sync for now. It will match all out other listing samples and we don't need to add the extra dependency. Once we have sorted out our quota issues in a different way than just printing 5 elements, we can revisit all listing operations.

AlexandrTrf:
> A couple of notes on the Async change.
> 
> Also let me know if/when the rest is done to re-review all. Thanks

Please perform a review of the code

spanner/api/Spanner.Samples.Tests/ListInstanceConfigsAsyncTest.cs
```diff
@@ -12,7 +12,6 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
-using System.Linq;
```
amanda-tarafa, line 15: The file name change is not needed anymore as this sample remained sync.

spanner/api/Spanner.Samples/DeleteInstanceConfigAsync.cs
```diff
@@ -0,0 +1,43 @@
...
+public class DeleteInstanceConfigAsyncSample
+{
+
+    public async Task DeleteInstanceConfigAsync(string projectId, string instanceConfigId)
+    {
+        InstanceAdminClient instanceAdminClient = await InstanceAdminClient.CreateAsync();
+        var instanceConfigName = new InstanceConfigName(projectId, instanceConfigId);
+
+        //Check if there is a custom config with this name.
+        await instanceAdminClient.GetInstanceConfigAsync(instanceConfigName);
```
amanda-tarafa, line 31: Instead of trying to obtain the config here (this will fail if the config does not exist), I think it's better to wrap the delete operation in a try catch and look for an RpcException with status "NOT_FOUND", which means the config didn't exist.

spanner/api/Spanner.Samples/UpdateInstanceConfigAsync.cs
```diff
@@ -0,0 +1,60 @@
...
+
+public class UpdateInstanceConfigAsyncSample
+{
+    public async Task<Operation<InstanceConfig, UpdateInstanceConfigMetadata>> UpdateInstanceConfigAsync(string projectId, string instanceConfigId)
+    {
+        InstanceAdminClient instanceAdminClient = await InstanceAdminClient.CreateAsync();
+        var instanceConfigName = new InstanceConfigName(projectId, instanceConfigId);
+
+        //Check if there is a custom config with this name.
+        var instanceConfig = await instanceAdminClient.GetInstanceConfigAsync(instanceConfigName);
```
amanda-tarafa, line 32: I think you don't need to check here. And in particular if you don't have a try catch the check may fail if the config does not exist.

Also, the purpose of the `UpdateMask` is that you don't have to have the full entity to update it. You create a new (blank) one, set the values that you want to update and set the mask accordingly. I think that shows a better use of the API/Library.

You can add a try catch to the updated operation and check for RpcException wit "NOT_FOUND" which means there was no such configuration before.

amanda-tarafa (changes requested):
A couple of changes, but nothing major.

Thanks!

spanner/api/Spanner.Samples/CreateInstanceConfigAsync.cs
```diff
@@ -0,0 +1,78 @@
...
+        {
+            ParentAsProjectName = ProjectName.FromProject(projectId),
+            InstanceConfig = userInstanceConfig,
+            InstanceConfigId = customInstanceConfigId
+        });
+
+        var pollResult = await operationResult.PollUntilCompletedAsync();
+
+        if (pollResult.IsFaulted)
+            throw pollResult.Exception;
```
amanda-tarafa, line 68:
```suggestion
        if (pollResult.IsFaulted)
        {
            throw pollResult.Exception;
        }
```

spanner/api/Spanner.Samples/UpdateInstanceConfigAsync.cs
```diff
@@ -0,0 +1,60 @@
...
+            UpdateMask = new FieldMask
+            {
+                Paths = { "display_name", "labels" }
+            }
+        });
+
+        updateInstanceConfigOperation = await updateInstanceConfigOperation.PollUntilCompletedAsync();
+        
+        if (updateInstanceConfigOperation.IsFaulted)
+            throw updateInstanceConfigOperation.Exception;
```
amanda-tarafa, line 53:
```suggestion
        if (updateInstanceConfigOperation.IsFaulted)
        {
            throw updateInstanceConfigOperation.Exception;
        }
```

amanda-tarafa (commented):
Two tiny nits, but otherwise LGTM.

I'll approve once the library with the needed changes is released and we can run the tests, but otherwise his looks good. Thanks!

AlexandrTrf:
> Two tiny nits, but otherwise LGTM.
> 
> I'll approve once the library with the needed changes is released and we can run the tests, but otherwise his looks good. Thanks!

Done

amanda-tarafa:
@AlexandrTrf this looks good, I'm just waiting for the released library to contain the feature so I can run the tests etc. Once the tests are green I'll appove and merge, thanks.

amanda-tarafa:
I've started the tests after updating the PR. If they are green we are ready to merge.

spanner/api/Spanner.Samples.Tests/SpannerFixture.cs
```diff
@@ -111,13 +116,64 @@ public async Task InitializeAsync()
...
+        DeleteInstanceConfig(CreateCustomInstanceConfigId);
+        DeleteInstanceConfig(UpdateCustomInstanceConfigId);
+        DeleteInstanceConfig(DeleteCustomInstanceConfigId);
+
         try
         {
             IList<Task> cleanupTasks = new List<Task>();
-
             cleanupTasks.Add(DeleteInstanceAsync(InstanceIdWithMultiRegion));
             cleanupTasks.Add(DeleteInstanceAsync(InstanceIdWithProcessingUnits));
```
amanda-tarafa, line 130: This won't build, seems like a conflict resolution gone wrong.

amanda-tarafa (changes requested):
@anshuldavid13  The code as it is is not building. I think it's just a conflict resolution on the fixture or a copy paste mistake. The rest looks fine so make no changes to it until we've been able to run the tests.

PR #1753, opened 2022-06-07
ci: Makes sharding print more dianogsing information.

amanda-tarafa:


jskeet (approved):
As far as I understand it (I don't do PowerShell) this looks fine.

PR #1794, opened 2022-06-23
tests(Spanner): Improve cleanup as it was becoming flaky again.

amanda-tarafa:


amanda-tarafa:
@jskeet it's fine to just do sanity checking here. There's a lot of cleanup and some changes in tests, but no changes in sample code.
If tests are green we should be fine.

spanner/api/Spanner.Samples.Tests/SpannerFixture.cs
```diff
@@ -34,62 +34,63 @@ public class SpannerFixture : IAsyncLifetime, ICollectionFixture<SpannerFixture>
     public string ProjectId { get; } = Environment.GetEnvironmentVariable("GOOGLE_PROJECT_ID");
     // Allow environment variables to override the default instance and database names.
     public string InstanceId { get; } = Environment.GetEnvironmentVariable("TEST_SPANNER_INSTANCE") ?? "my-instance";
-    public string DatabaseId { get; } = Environment.GetEnvironmentVariable("TEST_SPANNER_DATABASE") ?? $"my-db-{DateTimeOffset.UtcNow.ToUnixTimeMilliseconds()}";
-    public string PostgreSqlDatabaseId { get; } = Environment.GetEnvironmentVariable("TEST_SPANNER_POSTGRESQL_DATABASE") ?? $"my-db-postgre-{DateTimeOffset.UtcNow.ToUnixTimeMilliseconds()}";
+
+    public string DatabaseId { get; } = Environment.GetEnvironmentVariable("TEST_SPANNER_DATABASE") ?? $"my-db-{Guid.NewGuid().ToString("N").Substring(8)}";
```
jskeet, line 38: Maybe worth extracting `Guid.NewGuid().ToString("N").Substring(8)` to a new method so if you want to change the logic you only have to do it in one place?

spanner/api/Spanner.Samples.Tests/SpannerFixture.cs
```diff
@@ -102,68 +103,45 @@ public async Task InitializeAsync()
         }
     }
 
-    public Task DisposeAsync()
+    public async Task DisposeAsync()
```
jskeet, line 106: This will leave around older databases that would previously have been cleaned up, e.g. if a test was aborted abruptly. Is that okay? Do we have a separate cleanup program?

jskeet (approved):
I'm somewhat skimmed this one - a couple of comments, but it's basically okay.

spanner/api/Spanner.Samples.Tests/SpannerFixture.cs
amanda-tarafa: Yes, I've added a TODO in my notes to do that next week. I'll merge as is.

spanner/api/Spanner.Samples.Tests/SpannerFixture.cs
amanda-tarafa: I deleted everything that we had by hand. We don't have a cleanup process but I can set up one if we need to.
The rationaly before was that we kept the DBs around in case we needed to check for failures, in reality we never did that and it was causing quota issues. I'm hoping the case of test abruptly crashing is rare enough that I can just delete by hand again if we need to.

amanda-tarafa (commented):
I've addressed your comments, but I'll merge as is.
I'll iterate on it next week.

PR #1803, opened 2022-06-30
refactor(Spanner): Some fixture improvements.

amanda-tarafa:


amanda-tarafa:
@jskeet this is mostly extracting ID generation to a method as you suggested in #1794

PR #1810, opened 2022-07-11
Added importing, tests and setups(copy)

mkoshurnikov:
Added:

- Import Events/Products BQ Table + test
- Import Events/Product Gcs Bucket + tests
- Import Events/Products inline source + test
- setup files for Creating and Removing BQ tables and Gcs Buckets
- resources: table schema and table data


amanda-tarafa:
This is an exact copy of #1702 so I don't think we need it. Any reason why this PR has been opened?

mkoshurnikov:
Can be closed because it will be split in 2 other PRs.

PR #1842, opened 2022-08-04
chore: Update CODEOWNERS

amanda-tarafa:


jskeet (approved):
Changes look fine, but there are errors for securitycenter and retail.

amanda-tarafa:
> Changes look fine, but there are errors for securitycenter and retail.

Yes, because those folks don't have write permissions on the repo. I've invited them so it should be fine once they accept the invite.

PR #1847, opened 2022-08-05
chore: Remove build badges from README.

amanda-tarafa:
This hasn't been working in a while due to changes in CI and internal tooling.
Fixing this requires significant amount of work and is low priority right now.


amanda-tarafa:
I'll self merge on green as this change is trivial.

PR #1855, opened 2022-08-10
Create json files in buckets with invalid product/event

mkoshurnikov:
Fixed creation a GCS bucket with `user_events_some_invalid.json` file.
Fixed creation a GCS bucket with `products_some_invalid.json` file.

### DIFF
--- a/retail/interactive-tutorial/TestResourcesSetupCleanup/CreateTestResources.cs
+++ b/retail/interactive-tutorial/TestResourcesSetupCleanup/CreateTestResources.cs
@@ -300,6 +300,9 @@ public static class CreateTestResourcesTutorial
     private static readonly string ProductFileName = "products.json";
     private static readonly string EventsFileName = "user_events.json";
 
+    private static readonly string InvalidProductFileName = "products_some_invalid.json";
+    private static readonly string InvalidEventsFileName = "user_events_some_invalid.json";
+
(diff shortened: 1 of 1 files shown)


retail/interactive-tutorial/TestResourcesSetupCleanup/CreateTestResources.cs
```diff
@@ -323,10 +329,12 @@ public static void PerformCreationOfTestResources()
         // Create a GCS bucket with products.json file.
         var createdProductsBucket = CreateTestResources.CreateBucket(productsBucketName);
         CreateTestResources.UploadBlob(createdProductsBucket.Name, productFilePath, ProductFileName);
+        CreateTestResources.UploadBlob(createdProductsBucket.Name, invalidProductFilePath, InvalidProductFileName);
```
amanda-tarafa, line 332: Will this and the other Upload attempt fail? If the fails the script won't execute completely.

mkoshurnikov: No, there is just another another file in the bucket that has 1 invalid product. After products import in console output we have message about number of successfully imported products and number of failures during the importing. The same with  events.

amanda-tarafa (approved):
LGTM

PR #1859, opened 2022-08-11
feat: Exactly once delivery subscriber sample.

Rishabh-V:
This draft PR demonstrates the exactly once subscriber sample using the new `SubscriptionHandler` . The library changes are yet not released so the code will not compile and hence tests would fail. Added the do not merge label. This is done so that the code can be reviewed and fixes can be made. Thanks.

### DIFF
--- a/pubsub/api/Pubsub.Samples.Tests/CreateSubscriptionWithExactlyOnceDeliveryTest.cs
+++ b/pubsub/api/Pubsub.Samples.Tests/CreateSubscriptionWithExactlyOnceDeliveryTest.cs
@@ -1,0 +1,42 @@
+﻿// Copyright 2022 Google Inc.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+using Xunit;
+
+[Collection(nameof(PubsubFixture))]
+public class CreateSubscriptionWithExactlyOnceDeliveryTest
+{
+    private readonly PubsubFixture _pubsubFixture;
+    private readonly CreateSubscriptionWithExactlyOnceDeliverySample _createSubscriptionWithExactlyOnceDeliverySample;
+
+    public CreateSubscriptionWithExactlyOnceDeliveryTest(PubsubFixture pubsubFixture)
+    {
+        _pubsubFixture = pubsubFixture;
+        _createSubscriptionWithExactlyOnceDeliverySample = new CreateSubscriptionWithExactlyOnceDeliverySample();
+    }
+
+    [Fact]
+    public void CreateSubscriptionWithExactlyOnceDelivery()
+    {
+        string randomName = _pubsubFixture.RandomName();
+        string topicId = $"testTopicForExactlyOnceDeliverySubscriptionCreation{randomName}";
+        string subscriptionId = $"testSubscriptionForExactlyOnceDelivery{randomName}";
+
+        _pubsubFixture.CreateTopic(topicId);
+        _pubsubFixture.TempSubscriptionIds.Add(subscriptionId);
+        _createSubscriptionWithExactlyOnceDeliverySample.CreateSubscriptionWithExactlyOnceDelivery(_pubsubFixture.ProjectId, topicId, subscriptionId);
+        var subscription = _pubsubFixture.GetSubscription(subscriptionId);
+        Assert.True(subscription.EnableExactlyOnceDelivery);
+    }
+}
--- a/pubsub/api/Pubsub.Samples.Tests/PubsubFixture.cs
+++ b/pubsub/api/Pubsub.Samples.Tests/PubsubFixture.cs
@@ -94,9 +94,8 @@ public class PubsubFixture : IDisposable, ICollectionFixture<PubsubFixture>
     public Topic CreateTopic(string topicId)
     {
         var createTopicSampleObject = new CreateTopicSample();
-        var topic = createTopicSampleObject.CreateTopic(ProjectId, topicId);
         TempTopicIds.Add(topicId);
-        return topic;
+        return createTopicSampleObject.CreateTopic(ProjectId, topicId);
     }
 
     public Topic CreateTopicWithSchema(string topicId, string schemaId, Encoding encoding)
@@ -113,6 +112,13 @@ public class PubsubFixture : IDisposable, ICollectionFixture<PubsubFixture>
         var subscription = createSubscriptionSampleObject.CreateSubscription(ProjectId, topicId, subscriptionId);
         TempSubscriptionIds.Add(subscriptionId);
         return subscription;
+    }
+
+    public Subscription CreateExactlyOnceDeliverySubscription(string topicId, string subscriptionId)
+    {
+        var createSubscriptionWithExactlyOnceDeliverySample = new CreateSubscriptionWithExactlyOnceDeliverySample();
+        TempSubscriptionIds.Add(subscriptionId);
+        return createSubscriptionWithExactlyOnceDeliverySample.CreateSubscriptionWithExactlyOnceDelivery(ProjectId, topicId, subscriptionId);
     }
 
     public Schema CreateProtoSchema(string schemaId)
--- a/pubsub/api/Pubsub.Samples/CreateSubscriptionWithExactlyOnceDelivery.cs
+++ b/pubsub/api/Pubsub.Samples/CreateSubscriptionWithExactlyOnceDelivery.cs
@@ -1,0 +1,49 @@
+﻿// Copyright 2022 Google Inc.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+// [START pubsub_create_subscription_with_exactly_once_delivery]
+
+using Google.Cloud.PubSub.V1;
+using Grpc.Core;
+
+public class CreateSubscriptionWithExactlyOnceDeliverySample
+{
+    public Subscription CreateSubscriptionWithExactlyOnceDelivery(string projectId, string topicId, string subscriptionId)
+    {
+        SubscriberServiceApiClient subscriber = SubscriberServiceApiClient.Create();
+        TopicName topicName = TopicName.FromProjectTopic(projectId, topicId);
+        SubscriptionName subscriptionName = SubscriptionName.FromProjectSubscription(projectId, subscriptionId);
+
+        var subscriptionRequest = new Subscription
+        {
+            SubscriptionName = subscriptionName,
+            TopicAsTopicName = topicName,
+            EnableExactlyOnceDelivery = true
+        };
+
+        Subscription subscription = null;
+
+        try
+        {
+            subscription = subscriber.CreateSubscription(subscriptionRequest);
+        }
+        catch (RpcException e) when (e.Status.StatusCode == StatusCode.AlreadyExists)
+        {
+            // Already exists.  That's fine.
+        }
+
+        return subscription;
+    }
+}
+// [END pubsub_create_subscription_with_exactly_once_delivery]
--- a/pubsub/api/Pubsub.Samples/ExactlyOnceDeliverySubscriberAsync.cs
+++ b/pubsub/api/Pubsub.Samples/ExactlyOnceDeliverySubscriberAsync.cs
@@ -1,0 +1,92 @@
+﻿// Copyright 2022 Google Inc.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+// [START pubsub_subscriber_exactly_once]
+
+using Google.Cloud.PubSub.V1;
+using System;
+using System.Collections.Concurrent;
+using System.Collections.Generic;
+using System.Linq;
+using System.Threading;
+using System.Threading.Tasks;
+using static Google.Cloud.PubSub.V1.SubscriberClient;
+
+public class ExactlyOnceDeliverySubscriberAsyncSample
+{
+    public async Task<IEnumerable<string>> ExactlyOnceDeliverySubscriberAsync(string projectId, string subscriptionId)
+    {
+        // subscriptionId should be the ID of an exactly-once delivery subscription.
+        SubscriptionName subscriptionName = SubscriptionName.FromProjectSubscription(projectId, subscriptionId);
+        SubscriberClient subscriber = await SubscriberClient.CreateAsync(subscriptionName);
+        // To get the status of ACKnowledge (ACK) or Not ACKnowledge (NACK) request in exactly once delivery subscriptions,
+        // create a subscription handler that inherits from Google.Cloud.PubSub.V1.SubscriptionHandler. 
+        // For more information see Google.Cloud.PubSub.V1.SubscriptionHandler reference docs here:
+        // https://cloud.google.com/dotnet/docs/reference/Google.Cloud.PubSub.V1/latest/Google.Cloud.PubSub.V1.SubscriptionHandler?hl=en
+        var subscriptionHandler = new SampleSubscriptionHandler();
+        Task subscriptionTask = subscriber.StartAsync(subscriptionHandler);
+        // The subscriber will be running until it is stopped.
+        await Task.Delay(5000);
+        await subscriber.StopAsync(CancellationToken.None);
+        // Let's make sure that the start task finished successfully after the call to stop.
+        await subscriptionTask;
+        return subscriptionHandler.SuccessfulAckedIds;
+    }
+
+    // Sample handler to handle messages and ACK/NACK responses.
+    public class SampleSubscriptionHandler : SubscriptionHandler
+    {
+        public ConcurrentBag<string> SuccessfulAckedIds { get; } = new ConcurrentBag<string>();
+
+        /// <summary>
+        /// The function that processes received messages. It should be thread-safe.
+        /// Return <see cref="Reply.Ack"/> to ACKnowledge the message (meaning it won't be received again).
+        /// Return <see cref="Reply.Nack"/> to Not ACKnowledge the message (meaning it will be received again).
+        /// From the point of view of message acknowledgement, throwing an exception is equivalent to returning <see cref="Reply.Nack"/>.
+        /// </summary>
+        public override async Task<Reply> HandleMessage(PubsubMessage message, CancellationToken cancellationToken)
+        {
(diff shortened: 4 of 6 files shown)


pubsub/api/Pubsub.Samples.Tests/PubsubFixture.cs
```diff
@@ -115,6 +115,32 @@ public Subscription CreateSubscription(string topicId, string subscriptionId)
         return subscription;
     }
 
+    public Subscription CreateExactlyOnceDeliverySubscription(string projectId, string topicId, string subscriptionId)
```
amanda-tarafa, line 118: Double check with the Pub/Sub team, we might need a sample for showing how to create an exactly once subscription, in which case, this method would become a sample.

amanda-tarafa (changes requested):

pubsub/api/Pubsub.Samples/ExactlyOnceDeliverySubscriberAsync.cs
```diff
@@ -0,0 +1,87 @@
...
+        // To get the status of ACKnowledge (ACK) or Not ACKnowledge (NACK) request in exactly once delivery subscriptions,
+        // create a new handler that implements the SubscriptionHandler. The handler is backward compatible
+        // and can be used with non-exactly once delivery subscriptions as well.
+        var subscriptionHandler = new SampleSubscriptionHandler();
+        Task startTask = subscriber.StartAsync(subscriptionHandler);
+        // Run for 5 seconds.
+        await Task.Delay(5000);
+        await subscriber.StopAsync(CancellationToken.None);
+        // Let's make sure that the start task finished successfully after the call to stop.
+        // subscriptionHandler's HandleAckResponses will have the response for every ACKnowledge request that is successful or failed.
```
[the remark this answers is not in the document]
Rishabh-V: Hey Amanda, The method HandleAckResponses has a parameter of type IReadOnlyList<AckNackResponse> which will contain the successful or permanently failed AckIds. There may be a few messages with temporary errors as well, but they won't appear in the list unless they succeed or fail permanently. If have isn't appropriate, please suggest a suitable comment to convey this.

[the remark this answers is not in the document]
amanda-tarafa: Yes, OK, but still the method won't "have" anything, possibly "receive" instead?
And also, it is not clear why this comment is in this place? Will the method be called only after the susbscriber has stopped? Can it be called several times while the subscriber is listening, etc. but making sure that the start task is finished guarantees that it will be called for a last time?

pubsub/api/Pubsub.Samples/ExactlyOnceDeliverySubscriberAsync.cs
```diff
@@ -0,0 +1,87 @@
...
+        await startTask;
+        return subscriptionHandler.SuccessfulIds.Count;
+    }
+
+    // Sample handler to handle messages and ACK/NACK responses.
```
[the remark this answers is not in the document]
Rishabh-V: Right Amanda, I can do it. I will need to override the HandleNackResponses method and its implementation will be identical to HandleAckResponses just the messages will change from ACK to NACK. Currently the MessageHandler is just requesting ACK, so the handler will need to request NACK based on a field and hence I will also add a constructor to handler which accepts a boolean flag to request ack or nack, which will make the sample very long with similar code.  Please let me know if it is fine and I am happy to do it.

pubsub/api/Pubsub.Samples/ExactlyOnceDeliverySubscriberAsync.cs
```diff
@@ -0,0 +1,87 @@
...
+    internal sealed class SampleSubscriptionHandler : SubscriptionHandler
```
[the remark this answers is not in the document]
Rishabh-V: Sure.

pubsub/api/Pubsub.Samples/ExactlyOnceDeliverySubscriberAsync.cs
```diff
@@ -0,0 +1,87 @@
...
+    {
+        public ConcurrentBag<string> SuccessfulIds { get; } = new ConcurrentBag<string>();
```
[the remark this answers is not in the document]
Rishabh-V: Sure.

pubsub/api/Pubsub.Samples/ExactlyOnceDeliverySubscriberAsync.cs
```diff
@@ -0,0 +1,87 @@
...
+
+        /// <inheritdoc/>
```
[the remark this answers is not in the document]
Rishabh-V: Sure Amanda. Will remove the doc tag.

pubsub/api/Pubsub.Samples/ExactlyOnceDeliverySubscriberAsync.cs
```diff
@@ -0,0 +1,87 @@
...
+                        Console.WriteLine($"MessageId {response.MessageId} failed to acknowledge due to a permission denied error.");
+                        break;
+                    case AcknowledgementStatus.FailedPrecondition:
+                        Console.WriteLine($"MessageId {response.MessageId} failed to acknowledge due to a failed precondition.");
+                        break;
+                    case AcknowledgementStatus.InvalidAckId:
+                        Console.WriteLine($"MessageId {response.MessageId} failed to acknowledge due an invalid or expired AckId.");
+                        break;
+                    default:
+                        Console.WriteLine($"MessageId {response.MessageId} failed to acknowledge due to an unknown reason.");
```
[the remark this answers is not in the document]
Rishabh-V: In this case, we will have definite acknowledgement failure with the status of AcknowledgementStatus.Other. Please suggest if the message should be updated.

pubsub/api/Pubsub.Samples/ExactlyOnceDeliverySubscriberAsync.cs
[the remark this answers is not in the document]
Rishabh-V: Sure. Will update it with "receive". 
Right Amanda. The handler can be called several times while the subscriber is listening. Essentially, everytime we get a response from the Ack/Nack request this handler will be called.

Rishabh-V (commented):
Many thanks Amanda for a quick and very helpful review. I shall push the fixes in the next commit. I have updated the response to your comments. Just couple of confirmation questions. Please help me with your thoughts on those and I will make the appropriate changes and push. Thank you.

pubsub/api/Pubsub.Samples/ExactlyOnceDeliverySubscriberAsync.cs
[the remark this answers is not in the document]
amanda-tarafa: Ok, so maybe move this comment then to the handler itself? The code of it will part of the sample so that's fine.

pubsub/api/Pubsub.Samples/ExactlyOnceDeliverySubscriberAsync.cs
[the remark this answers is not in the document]
amanda-tarafa: I was thinking maybe you act as if you receive two distinct messages, say "Hello" and "Goodbye" and you ack one and nack the other. And then you can show how to handle both types of responses?

pubsub/api/Pubsub.Samples/ExactlyOnceDeliverySubscriberAsync.cs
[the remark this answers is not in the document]
amanda-tarafa: So, be explicit about AcknowledgementStatus.Other (same as for the other statuses) keeping your original output and on the default add my output.
You can do `var status = (AcknowledgementStatus) 10000;` and that's different than `AcknowledgementStatus.Other`.

amanda-tarafa (changes requested):

pubsub/api/Pubsub.Samples/ExactlyOnceDeliverySubscriberAsync.cs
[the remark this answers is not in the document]
Rishabh-V: As discussed offline, we will keep it simple.

pubsub/api/Pubsub.Samples/ExactlyOnceDeliverySubscriberAsync.cs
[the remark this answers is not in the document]
Rishabh-V: Sure.

pubsub/api/Pubsub.Samples/ExactlyOnceDeliverySubscriberAsync.cs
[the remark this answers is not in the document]
Rishabh-V: Sure.

Rishabh-V (commented):
Code review comments addressed.

pubsub/api/Pubsub.Samples/ExactlyOnceDeliverySubscriberAsync.cs
```diff
@@ -0,0 +1,90 @@
...
+    {
+        // subscriptionId should be an exactly once delivery subscription.
+        SubscriptionName subscriptionName = SubscriptionName.FromProjectSubscription(projectId, subscriptionId);
+        SubscriberClient subscriber = await SubscriberClient.CreateAsync(subscriptionName);
+        // To get the status of ACKnowledge (ACK) or Not ACKnowledge (NACK) request in exactly once delivery subscriptions,
+        // create a subscription handler that inherits from Google.Cloud.PubSub.V1.SubscriptionHandler. 
+        // For more information see Google.Cloud.PubSub.V1.SubscriptionHandler reference docs here:
+        // TODO: [add link to reference docs when we have them].
+        var subscriptionHandler = new SampleSubscriptionHandler();
+        Task startTask = subscriber.StartAsync(subscriptionHandler);
```
jskeet, line 38: I'd call this `subscriptionTask` - it's not just for starting the subscriber, it will only complete when the subscriber has finished (as I understand it).

jskeet (commented):
One small nit, but basically looks fine in terms of diagnostics.

What it doesn't show is what clients are expected to do with permanent failures - but I'm not sure what the expectation of the sample is in that respect.

Rishabh-V (commented):
Thanks Jon.  The sample is in line with corresponding Java sample. 

I have also added a sample to create subscription with exactly once delivery.

pubsub/api/Pubsub.Samples.Tests/PubsubFixture.cs
```diff
@@ -115,6 +115,14 @@ public Subscription CreateSubscription(string topicId, string subscriptionId)
         return subscription;
     }
 
+    public Subscription CreateExactlyOnceDeliverySubscription(string topicId, string subscriptionId)
+    {
+        var createSubscriptionWithExactlyOnceDeliverySample = new CreateSubscriptionWithExactlyOnceDeliverySample();
+        var subscription = createSubscriptionWithExactlyOnceDeliverySample.CreateSubscriptionWithExactlyOnceDelivery(ProjectId, topicId, subscriptionId);
+        TempSubscriptionIds.Add(subscriptionId);
```
amanda-tarafa, line 122: Add the subscription ID to the temp list before creating the subscription. If creation fails client side, it might have still sucedded server side and we'd want to delete the subscription.

pubsub/api/Pubsub.Samples/CreateSubscriptionWithExactlyOnceDelivery.cs
```diff
@@ -0,0 +1,48 @@
...
+
+        try
+        {
+            subscription = subscriber.CreateSubscription(subscriptionRequest);
+        }
+        catch (RpcException e) when (e.Status.StatusCode == StatusCode.AlreadyExists)
+        {
+            // Already exists.  That's fine.
+        }
+        return subscription;
```
amanda-tarafa, line 45:
```suggestion
        try
        {
            return subscriber.CreateSubscription(subscriptionRequest);
        }
        catch (RpcException e) when (e.Status.StatusCode == StatusCode.AlreadyExists)
        {
            // Already exists.  That's fine.
        }
```

pubsub/api/Pubsub.Samples.Tests/CreateSubscriptionWithExactlyOnceDeliveryTest.cs
```diff
@@ -0,0 +1,42 @@
...
+    [Fact]
+    public void CreateSubscriptionWithExactlyOnceDelivery()
+    {
+        string randomName = _pubsubFixture.RandomName();
+        string topicId = $"testTopicForExactlyOnceDeliverySubscriptionCreation{randomName}";
+        string subscriptionId = $"testSubscriptionForExactlyOnceDelivery{randomName}";
+
+        _pubsubFixture.CreateTopic(topicId);
+        _createSubscriptionWithExactlyOnceDeliverySample.CreateSubscriptionWithExactlyOnceDelivery(_pubsubFixture.ProjectId, topicId, subscriptionId);
+        _pubsubFixture.TempSubscriptionIds.Add(subscriptionId);
```
amanda-tarafa, line 38: Put this before actually creating the subscription.

Also, confirm that CreateTopic in the fixture adds the topic ID to the temp list before actually creating the topic.

pubsub/api/Pubsub.Samples/ExactlyOnceDeliverySubscriberAsync.cs
```diff
@@ -0,0 +1,90 @@
...
+using System.Linq;
+using System.Threading;
+using System.Threading.Tasks;
+using static Google.Cloud.PubSub.V1.SubscriberClient;
+
+public class ExactlyOnceDeliverySubscriberAsyncSample
+{
+    public async Task<IEnumerable<string>> ExactlyOnceDeliverySubscriberAsync(string projectId, string subscriptionId)
+    {
+        // subscriptionId should be an exactly once delivery subscription.
```
amanda-tarafa, line 30:
```suggestion
        // subscriptionId should be the ID of an exactly-once delivery subscription.
```

pubsub/api/Pubsub.Samples/ExactlyOnceDeliverySubscriberAsync.cs
```diff
@@ -0,0 +1,90 @@
...
+        public ConcurrentBag<string> SuccessfulAckedIds { get; } = new ConcurrentBag<string>();
+
+        public override async Task<Reply> HandleMessage(PubsubMessage message, CancellationToken cancellationToken)
+        {
+            string text = System.Text.Encoding.UTF8.GetString(message.Data.ToArray());
+            Console.WriteLine($"Message {message.MessageId}: {text}");
+            return await Task.FromResult(Reply.Ack);
+        }
+
+        // This method will receive the response for every ACKnowledge request that is successful or failed.
```
amanda-tarafa, line 59: Maybe in summary tags?
And consider doing the same for the HandleMessage method.

pubsub/api/Pubsub.Samples/ExactlyOnceDeliverySubscriberAsync.cs
```diff
@@ -0,0 +1,90 @@
...
+        public override void HandleAckResponses(IReadOnlyList<AckNackResponse> responses)
+        {
+            foreach (var response in responses)
+            {
+                switch (response.Status)
```
amanda-tarafa, line 64: Consider changing this to a switch expression whose result is the message to print, and then after the switch expression you just print the message.
I think it will be a lot clearer like that.
```suggestion
                string result = response.Status switch
                {
                    AkcnowledgementStatus.Success => $"MessageId {response.MessageId} succesfully aknowledged.",
                    AcknowledgementStatus.PermissionDenied => $"...",
                    ...
                };
                Console.WriteLine(result);
```

amanda-tarafa (changes requested):

pubsub/api/Pubsub.Samples.Tests/PubsubFixture.cs
Rishabh-V: Done.

pubsub/api/Pubsub.Samples/ExactlyOnceDeliverySubscriberAsync.cs
Rishabh-V: This will indeed be a lot clearer. I however need to add the successful message ID to the list in the Success case and that is where I got stuck as adding to the list is a statement not an expression but very happy to change this if there is a way to do both in the switch expression.

pubsub/api/Pubsub.Samples.Tests/CreateSubscriptionWithExactlyOnceDeliveryTest.cs
Rishabh-V: Done. Also updated the code for CreateTopic.

pubsub/api/Pubsub.Samples/ExactlyOnceDeliverySubscriberAsync.cs
```diff
@@ -0,0 +1,94 @@
...
+{
+    public async Task<IEnumerable<string>> ExactlyOnceDeliverySubscriberAsync(string projectId, string subscriptionId)
+    {
+        // subscriptionId should be the ID of an exactly-once delivery subscription.
+        SubscriptionName subscriptionName = SubscriptionName.FromProjectSubscription(projectId, subscriptionId);
+        SubscriberClient subscriber = await SubscriberClient.CreateAsync(subscriptionName);
+        // To get the status of ACKnowledge (ACK) or Not ACKnowledge (NACK) request in exactly once delivery subscriptions,
+        // create a subscription handler that inherits from Google.Cloud.PubSub.V1.SubscriptionHandler. 
+        // For more information see Google.Cloud.PubSub.V1.SubscriptionHandler reference docs here:
+        // TODO: [add link to reference docs when we have them].
```
amanda-tarafa, line 36: We have ref docs already

pubsub/api/Pubsub.Samples/ExactlyOnceDeliverySubscriberAsync.cs
```diff
@@ -0,0 +1,94 @@
...
+    public class SampleSubscriptionHandler : SubscriptionHandler
+    {
+        public ConcurrentBag<string> SuccessfulAckedIds { get; } = new ConcurrentBag<string>();
+
+        /// <summary>
+        /// The handler function that is passed all received messages.
+        /// This function may be called on multiple threads concurrently. Return <see cref="Reply.Ack"/> from this function
+        /// to ACKnowledge this message (implying it won't be received again); or return <see cref="Reply.Nack"/> to Not
+        /// ACKnowledge this message (implying it will be received again). If this function throws any Exception, then
+        /// it behaves as if it returned <see cref="Reply.Nack"/>.
```
amanda-tarafa, line 57:
```suggestion
        /// The function that processes received messages. It should be thread-safe.
        /// Return <see cref="Reply.Ack"/> to ACKnowledge the message (meaning it won't be received again).
        /// Return <see cref="Reply.Nack"/> to Not ACKnowledge the message (meaning it will be received again).
        /// From the point of view of message akcnowledgement, throwing an exception is equivalent to returning see cref="Reply.Nack"/>.
```

pubsub/api/Pubsub.Samples/ExactlyOnceDeliverySubscriberAsync.cs
```diff
@@ -0,0 +1,94 @@
...
+        /// </summary>
+        public override async Task<Reply> HandleMessage(PubsubMessage message, CancellationToken cancellationToken)
+        {
+            string text = System.Text.Encoding.UTF8.GetString(message.Data.ToArray());
+            Console.WriteLine($"Message {message.MessageId}: {text}");
+            return await Task.FromResult(Reply.Ack);
+        }
+
+        /// <summary>
+        /// This method will receive the response for every ACKnowledge request that is successful or failed.
```
amanda-tarafa, line 67:
```suggestion
        /// This method will receive responses for all ACKnowledge requests.
```

pubsub/api/Pubsub.Samples/ExactlyOnceDeliverySubscriberAsync.cs
```diff
@@ -0,0 +1,94 @@
...
+        /// </summary>
+        /// <param name="responses">The read-only list containing the response of every ACKnowledge request that is successful or failed.</param>
```
amanda-tarafa, line 69:
```suggestion
```

amanda-tarafa (changes requested):

Rishabh-V (commented):
Thanks Amanda. I have reverted the csproj changes (need to check why they are still showing up) 
Have updated the comments as suggested by you. Will update the reference docs link on Monday and request re-review. The build and tests will fail for now as csproj changes are reverted.

Rishabh-V:
csproj file is reverted but is still showing up. I will remove it.

PR #1871, opened 2022-08-25
fix(functions): add 2nd-gen region tags

ace-n:


functions/helloworld/HelloGcs/Function.cs
```diff
@@ -13,6 +13,7 @@
 // limitations under the License.
 
 // [START functions_helloworld_storage]
+// [START functions_cloudevent_storage]
```
amanda-tarafa, line 16: Are you certain these region tags will be picked up properly like this? I mean, shouldn't the new tags be either fully inside or fully outside the old tags? Otherwise LGTM, but if you can confirm that this works as is, that'd be great.

ace-n: I don't think it matters, but I went ahead and fixed it just in case.

amanda-tarafa (approved):
Thanks!

PR #1889, opened 2022-08-31
fix(PubSub): Stop using deprecated code.

amanda-tarafa:


pubsub/api/Pubsub.Samples/PublishMessageWithRetrySettingsAsync.cs
```diff
@@ -32,19 +32,20 @@ public async Task PublishMessageWithRetrySettingsAsync(string projectId, string
...
-               clientCreationSettings: new PublisherClient.ClientCreationSettings(
-                   publisherServiceApiSettings: new PublisherServiceApiSettings
-                   {
-                       PublishSettings = CallSettings.FromRetry(RetrySettings.FromExponentialBackoff(
+        var publisher = await new PublisherClientBuilder
+        {
+            TopicName = topicName,
+            ApiSettings = new PublisherServiceApiSettings
+            {
+                PublishSettings = CallSettings.FromRetry(RetrySettings.FromExponentialBackoff(
```
jskeet, line 40: It might be worth separating out the RetrySettings construction into a separate statement (and variable) beforehand. That can happen later though.

jskeet (approved):

pubsub/api/Pubsub.Samples/PublishMessageWithRetrySettingsAsync.cs
amanda-tarafa: Yes, I didn't  want to make many changes here, and there's definetely a mix of settings being created in-line or beforehand across different samples that we should look into. But yes, I don't have the time for that right now. I'm adding a low priority TODO for myself but also @Rishabh-V as FYI, you might be able to look into this at some point.

Rishabh-V: Sure Amanda. Thanks.

PR #1905, opened 2022-09-09
fix(Live Stream): delete inputs and channels more than 2 hours old

### LINKED ISSUE #1903: [Livestream]: tests are flaky.
Quota is being reached on channels, so maybe channels 2 hours old need to be deleted, currently channels 24hrs old are being deleted.

@irataxy this is just so we don't forget.

irataxy:
Fixes #1903


amanda-tarafa (approved):
LGTM

PR #1912, opened 2022-09-15
Update user env setup

mkoshurnikov:
Updated user_environment_setup.sh, so we can run script more than once.


chenlei1216 (approved):

amanda-tarafa (approved):
LGTM

PR #1935, opened 2022-10-04
Update output in SearchWithFacetSpecSample.cs

mkoshurnikov:


retail/interactive-tutorial/RetailSearch.Samples/SearchWithFacetSpecSample.cs
```diff
@@ -104,6 +104,13 @@ public IEnumerable<SearchResponse> Search(string projectId)
                 Console.WriteLine(item);
                 Console.WriteLine();
             }
+            Console.WriteLine($"Facets:");
+            var facets = firstPage.Facets.AsEnumerable().First().Values;
+            foreach (var item in facets)
+            {
+                Console.WriteLine(item);
+            }
+            Console.WriteLine();
```
amanda-tarafa, line 113: To print all facet values you should do as follows:
Your code only prints the values of the first facet, and the `AsEnumerable` call is not needed there. Even if on this particular sample you only get one facet back, it's best to show users how to deal with the general case.

```suggestion
            Console.WriteLine("Facets:");
            foreach (var facet in firstPage.Facets)
            {
                Console.WriteLine($"Facet key: {facet.Key}");
                Console.WriteLine("Facet values: ");
                foreach (var facetValue in facet.Values)
                {
                    Console.WriteLine(facetValue);
                }
            }
            Console.WriteLine();
```

amanda-tarafa (changes requested):

retail/interactive-tutorial/RetailSearch.Samples/SearchWithFacetSpecSample.cs
mkoshurnikov: sure, I changed that a little bit

amanda-tarafa (approved):
Thanks! Will run tests and merge on green.

PR #1945, opened 2022-10-18
chore: adding standardized region tags

meredithslota:
Adding in region tags with standard product prefix for Sample Browser.


meredithslota:
For region tag mapping: [go/firestore-region-tags-normalization-2020q4](https://goto.google.com/firestore-region-tags-normalization-2020q4)

amanda-tarafa (approved):
LGTM

PR #1950, opened 2022-10-21
feat:PgJsonb samples

Rishabh-V:
The PR will not compile as PgJsonb changes are not released yet. The idea is to get the sample code reviewed. Thanks.

### DIFF
--- a/spanner/api/Spanner.Samples.Tests/QueryJsonbDataUsingParameterAsyncPostgreTest.cs
+++ b/spanner/api/Spanner.Samples.Tests/QueryJsonbDataUsingParameterAsyncPostgreTest.cs
@@ -12,6 +12,7 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
+using Newtonsoft.Json;
 using System.Threading.Tasks;
 using Xunit;
 
@@ -40,7 +41,20 @@ public class QueryJsonbDataUsingParameterAsyncPostgreTest
         // Act - Query the VenueInformation table.
         var venues = await _sample.QueryJsonbDataUsingParameterAsyncPostgre(_spannerFixture.ProjectId, _spannerFixture.InstanceId, _spannerFixture.PostgreSqlDatabaseId);
 
-        // Assert that VenueId 19 exists in the queried data from the table.
-        Assert.Contains(venues, v => v.VenueId == 19);
+        // Assert that venues collection is not empty and all the venues in the collection have a rating greater than 2.
+        Assert.NotEmpty(venues);
+        Assert.All(venues, v => AssertRating(v.Details, 2));
+    }
+
+    // Asserts that the rating of the given venue details is greater than the threshold value.
+    private static void AssertRating(string details, int threshold)
+    {
+        var venueDetails = JsonConvert.DeserializeObject<Details>(details);
+        Assert.True(venueDetails.Rating > threshold);
+    }
+
+    private struct Details
+    {
+        public int Rating { get; set; }
     }
 }
--- a/spanner/api/Spanner.Samples.Tests/QueryJsonbDataUsingParameterAsyncPostgreTest.cs
+++ b/spanner/api/Spanner.Samples.Tests/QueryJsonbDataUsingParameterAsyncPostgreTest.cs
@@ -1,0 +1,46 @@
+﻿// Copyright 2022 Google Inc.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+using System.Threading.Tasks;
+using Xunit;
+
+[Collection(nameof(SpannerFixture))]
+public class QueryJsonbDataUsingParameterAsyncPostgreTest
+{
+    private readonly SpannerFixture _spannerFixture;
+
+    private readonly QueryJsonbDataUsingParameterAsyncPostgreSample _sample;
+
+    private readonly UpdateDataWithJsonbAsyncPostgreSample _updateDataSample;
+
+    public QueryJsonbDataUsingParameterAsyncPostgreTest(SpannerFixture spannerFixture)
+    {
+        _spannerFixture = spannerFixture;
+        _sample = new QueryJsonbDataUsingParameterAsyncPostgreSample();
+        _updateDataSample = new UpdateDataWithJsonbAsyncPostgreSample();
+    }
+
+    [Fact]
+    public async Task TestQueryJsonbDataUsingParameterAsyncPostgre()
+    {
+        // Arrange - Update data.
+        await _updateDataSample.UpdateDataWithJsonbAsyncPostgre(_spannerFixture.ProjectId, _spannerFixture.InstanceId, _spannerFixture.PostgreSqlDatabaseId);
+
+        // Act - Query the VenueInformation table.
+        var venues = await _sample.QueryJsonbDataUsingParameterAsyncPostgre(_spannerFixture.ProjectId, _spannerFixture.InstanceId, _spannerFixture.PostgreSqlDatabaseId);
+
+        // Assert that VenueId 19 exists in the queried data from the table.
+        Assert.Contains(venues, v => v.VenueId == 19);
+    }
+}
--- a/spanner/api/Spanner.Samples.Tests/SpannerFixture.cs
+++ b/spanner/api/Spanner.Samples.Tests/SpannerFixture.cs
@@ -271,6 +271,7 @@ public class SpannerFixture : IAsyncLifetime, ICollectionFixture<SpannerFixture>
     {
         CreateDatabaseAsyncPostgreSample createDatabaseAsyncSample = new CreateDatabaseAsyncPostgreSample();
         await createDatabaseAsyncSample.CreateDatabaseAsyncPostgre(ProjectId, InstanceId, PostgreSqlDatabaseId);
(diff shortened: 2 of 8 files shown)


amanda-tarafa:
@Rishabh-V I'll review when I'm back, on the week of 31st October, thanks!

spanner/api/Spanner.Samples.Tests/SpannerFixture.cs
```diff
@@ -533,4 +534,48 @@ public IEnumerable<Database> GetDatabases()
         var databases = DatabaseAdminClient.ListDatabases(instanceName);
         return databases;
     }
+
+    private async Task CreateVenueTablesAndInsertDataAsyncPostgre()
```
amanda-tarafa, line 538: This maybe should be done in the individual tests that use them.

The resources that we create in the fixture are usually going to be shared by several tests. This doesn't seem to be the case here.

spanner/api/Spanner.Samples.Tests/SpannerFixture.cs
```diff
@@ -533,4 +534,48 @@ public IEnumerable<Database> GetDatabases()
...
+            using var cmd = PgSpannerConnection.CreateInsertCommand("VenueInformation", new SpannerParameterCollection
+            {
+                { "VenueId", SpannerDbType.Int64, id },
+                { "VenueName", SpannerDbType.String, $"Venue {id}" }
+            });
+            return cmd.ExecuteNonQueryAsync();
+        }));
+    }
+
+    private async Task CreateTablesAsyncPostgre(string[] createTableStatements)
```
amanda-tarafa, line 573: You can leave this as a public helper method to call from the tests, etc.

spanner/api/Spanner.Samples/QueryJsonbDataUsingParameterAsyncPostgre.cs
```diff
@@ -0,0 +1,53 @@
...
+
+public class QueryJsonbDataUsingParameterAsyncPostgreSample
+{
+    public async Task<List<VenueInformation>> QueryJsonbDataUsingParameterAsyncPostgre(string projectId, string instanceId, string databaseId)
+    {
+        string connectionString = $"Data Source=projects/{projectId}/instances/{instanceId}/databases/{databaseId}";
+        using var connection = new SpannerConnection(connectionString);
+        // Get all the venues with a rating greater than 2.
+        using var command = connection.CreateSelectCommand(
+            "SELECT venueid, details FROM VenueInformation WHERE CAST(details ->> 'rating' AS INTEGER) > $1",
```
amanda-tarafa, line 29: Is it clear from the context of this sample that details is a JSON? If not, a note here showing the structure of details could be useful.

spanner/api/Spanner.Samples/UpdateDataWithJsonbAsyncPostgre.cs
```diff
@@ -0,0 +1,102 @@
...
+        string connectionString = $"Data Source=projects/{projectId}/instances/{instanceId}/databases/{databaseId}";
+        using var connection = new SpannerConnection(connectionString);
+        await connection.OpenAsync();
+
+        await Task.WhenAll(venueInformationList.Select(venue =>
+        {
+            // Update rows in the Venues table.
+            using var cmd = connection.CreateUpdateCommand("VenueInformation", new SpannerParameterCollection
+            {
+                    { "VenueId", SpannerDbType.Int64, venue.VenueId },
```
amanda-tarafa, line 88: nit: this indent is wrong

spanner/api/Spanner.Samples.Tests/QueryJsonbDataUsingParameterAsyncPostgreTest.cs
```diff
@@ -0,0 +1,46 @@
...
+    [Fact]
+    public async Task TestQueryJsonbDataUsingParameterAsyncPostgre()
+    {
+        // Arrange - Update data.
+        await _updateDataSample.UpdateDataWithJsonbAsyncPostgre(_spannerFixture.ProjectId, _spannerFixture.InstanceId, _spannerFixture.PostgreSqlDatabaseId);
+
+        // Act - Query the VenueInformation table.
+        var venues = await _sample.QueryJsonbDataUsingParameterAsyncPostgre(_spannerFixture.ProjectId, _spannerFixture.InstanceId, _spannerFixture.PostgreSqlDatabaseId);
+
+        // Assert that VenueId 19 exists in the queried data from the table.
```
amanda-tarafa, line 43: I'd prefer if we checked that the list is not empty and all elements on the list have rating > 2.

amanda-tarafa (changes requested):
Looking good, a couple of comments only.

spanner/api/Spanner.Samples.Tests/SpannerFixture.cs
Rishabh-V: I will move the VenueDetails table creation to the AddJsonbColumn test. 

However, Table VenueInformation is used by QueryJsonbData and UpdateDataWithJsonb samples. If we move this method to individual tests, we may need to add a condition to run the method only if they don't exist as either of these samples may execute first and probably that code may be duplicated.  

Please let me know if it is fine to keep VenueInformation table creation and data insertion in fixture or if you have other suggestions. Thanks.

spanner/api/Spanner.Samples.Tests/SpannerFixture.cs
Rishabh-V: Sure.

spanner/api/Spanner.Samples/QueryJsonbDataUsingParameterAsyncPostgre.cs
Rishabh-V: Sure. I will add the following: Details is a column of type Jsonb. The data persisted in this column has the following structure. 
Details = [{
                    "name": "string",
                    "available": true,
                    "rating": int  // This field is optional. 
                }]
Please let me know if any changes are needed. Thanks.

spanner/api/Spanner.Samples.Tests/QueryJsonbDataUsingParameterAsyncPostgreTest.cs
Rishabh-V: Sure. Will change this.

Rishabh-V (commented):
Thanks Amanda. Couple of queries to make the correct changes. I will push the fixes in next commit.

spanner/api/Spanner.Samples.Tests/SpannerFixture.cs
amanda-tarafa: Yes, whatever is shared can stay in the fixture, of course, that's its main purpose.

Just remove from here what's used by a single test only.
One of the main reasons for these is that, during development, if you are adding or modifying just a few tests, and you want to run those few tests only, then you have to wait for all these tables creation to be done as the fixture is initialized regardless of whether you run one test or 20. With the shared resources, that's fine, but with the one-offs it just adds to waiting timing during dev, and this fixture is already heavy.

spanner/api/Spanner.Samples/QueryJsonbDataUsingParameterAsyncPostgre.cs
amanda-tarafa: Yep, that seems right, only thing is, I believe that Details does not necessarily have the same structure for all rows, right? So instead of saying "The data persisted in this column has the following structure" I would say "Some of the data persisted in this column ...."

spanner/api/Spanner.Samples/UpdateDataWithJsonbAsyncPostgre.cs
```diff
@@ -0,0 +1,102 @@
...
+                Details = JsonConvert.SerializeObject(new
+                {
+                    rating = 9,
+                    open = true,
+                })
+            },
+            new VenueInformation
+            {
+                VenueId = 4,
+                // PostgreSQL JSONB sorts first by key length and then lexicographically with equivalent key length
```
amanda-tarafa, line 43: I don't understand what this comment is about.
Does key here mean field?
Also, I don't think what's described in this comment is represented on the example below, which makes the comment a little out of place.

spanner/api/Spanner.Samples/AddJsonbColumnAsyncPostgre.cs
```diff
@@ -0,0 +1,34 @@
...
+{
+    public async Task AddJsonbColumnAsyncPostgre(string projectId, string instanceId, string databaseId)
+    {
+        string connectionString = $"Data Source=projects/{projectId}/instances/{instanceId}/databases/{databaseId}";
+        string alterStatement = $"ALTER TABLE VenueDetails ADD COLUMN Details JSONB";
+
+        using var connection = new SpannerConnection(connectionString);
+        using var ddlCmd = connection.CreateDdlCommand(alterStatement);
+        await ddlCmd.ExecuteNonQueryAsync();
+        Console.WriteLine($"Added the jsonb column named Details to VenueDetails table.");
```
amanda-tarafa, line 31: nit: should this be
```suggestion
        Console.WriteLine($"Added the JSONB column named Details to VenueDetails table.");
```

spanner/api/Spanner.Samples/UpdateDataWithJsonbAsyncPostgre.cs
Rishabh-V: Sure Amanda. Does this sound better? "PostgreSQL JSONB takes the last value in the case of duplicate fields".

spanner/api/Spanner.Samples/AddJsonbColumnAsyncPostgre.cs
Rishabh-V: Sure. Will change this.

spanner/api/Spanner.Samples/UpdateDataWithJsonbAsyncPostgre.cs
amanda-tarafa: We have discussed this offline and agreed on a couple changes around this comment.

amanda-tarafa (approved):
LGTM, thanks!

Issue #1937, opened 2022-10-10
[Retail]: RetailSearch.Samples.Tests.SearchWithQueryExpansionTest.TestSearchWithQueryExpansion is failing


amanda-tarafa:
Two different errors in diferent runs

```
Failed RetailSearch.Samples.Tests.SearchWithQueryExpansionTest.TestSearchWithQueryExpansion [3 s]
  Error Message:
   Assert.Contains() Failure
Not found: (filter expression)
In value:  SearchResponse []
  Stack Trace:
     at RetailSearch.Samples.Tests.SearchWithQueryExpansionTest.TestSearchWithQueryExpansion() in T:\src\github\dotnet-docs-samples\retail\interactive-tutorial\RetailSearch.Samples.Tests\SearchWithQueryExpansionTest.cs:line 34
```

```
Failed RetailSearch.Samples.Tests.SearchWithQueryExpansionTest.TestSearchWithQueryExpansion [250 ms]
  Error Message:
   System.ArgumentOutOfRangeException : Index was out of range. Must be non-negative and less than the size of the collection. (Parameter 'index')
  Stack Trace:
     at System.Collections.Generic.List`1.get_Item(Int32 index)
   at RetailSearch.Samples.Tests.SearchWithQueryExpansionTest.TestSearchWithQueryExpansion() in /tmpfs/src/github/dotnet-docs-samples/retail/interactive-tutorial/RetailSearch.Samples.Tests/SearchWithQueryExpansionTest.cs:line 31
```

chenlei1216:
This issue has been fixed on the backend and I cannot reproduce it anymore.

amanda-tarafa:
This continues to happen:

On nightly CI, consistently, since October 20.

Just now on presubmit:

Here: https://source.cloud.google.com/results/invocations/3af94d8b-6179-4928-8466-52fb8bec07e5/log
```
Failed RetailSearch.Samples.Tests.SearchWithQueryExpansionTest.TestSearchWithQueryExpansion [815 ms]
  Error Message:
   System.ArgumentOutOfRangeException : Index was out of range. Must be non-negative and less than the size of the collection. (Parameter 'index')
  Stack Trace:
     at System.Collections.Generic.List`1.get_Item(Int32 index)
[...]

PR #1962, opened 2022-11-03
chore: Remove dependencies on coverlet.collector as we don't use it.

amanda-tarafa:


amanda-tarafa:
Failures are unrelated and being tracked in #1937 . Merging...

PR #1965, opened 2022-11-04
fix(Retail): Fix SearchWithQueryExpansionTest

chenlei1216:
With the test query and page size 10 our backend only returns 1 page of result currently. Fix the test case to only check the first page.


amanda-tarafa (approved):
LGTM

PR #1971, opened 2022-11-08
chore(firestore): update region tags to use product prefix

meredithslota:
Note: these are not used, according to my tracking sheet, so let's see if SnippetBot agrees. :) 

Context: [go/firestore-region-tags-normalization-2020q4](https://goto.google.com/firestore-region-tags-normalization-2020q4)


amanda-tarafa (approved):
LGTM. And SnippetBot agrees!

PR #1983, opened 2022-11-30
fix(tasks): add main method + clarify arg structure

ace-n:


cloudtasks/api/TasksSample/TasksSample.csproj
```diff
@@ -1,7 +1,7 @@
 <Project Sdk="Microsoft.NET.Sdk">
   <PropertyGroup>
     <TargetFramework>net6.0</TargetFramework>
-
+    <GenerateProgramFile>false</GenerateProgramFile>
```
ace-n, line 4: Added based on [this blog post](https://andrewlock.net/fixing-the-error-program-has-more-than-one-entry-point-defined-for-console-apps-containing-xunit-tests/).

cloudtasks/api/TasksSample/CreateTask.cs
```diff
@@ -50,5 +51,11 @@ class CreateAppEngineTask
         Console.WriteLine($"Created Task {response.Name}");
         return response.Name;
     }
+
+    public static void Main(string[] args)
+    {
+        var snippet = new CreateAppEngineTask();
+        snippet.CreateTask();
+    }
```
amanda-tarafa, line 59: We have been streadily moving away from Main methods in C#, we don't need them as we can have parameters with default values.
Any particular reason this is needed here?

cloudtasks/api/TasksSample/CreateTask.cs
```diff
@@ -22,6 +22,7 @@
 class CreateAppEngineTask
 {
     public string CreateTask(
+        // TODO<developer>: replace these values
```
amanda-tarafa, line 25:
```suggestion
        // TODO<developer>: Call this method by passing correct values for the following parameters or change the parameters' default values.
```

ace-n: Added!

cloudtasks/api/TasksSample/CreateTask.cs
ace-n: Clarified internally - this has since been reverted.

cloudtasks/api/TasksSample/TasksSample.csproj
ace-n: This has since been reverted

amanda-tarafa (approved):
LGTM, thanks!

PR #2014, opened 2023-01-10
chore: Update all samples to Functions Framework v2

jskeet:
(Initially to FF v2.0.0-beta01, but we can get it ready that way, then quickly swap out the dependencies after FF v2 has gone GA.)

Note that the C# code changes are basically just using file-scoped namespaces, which reduce the indent level (particularly handy for samples).


amanda-tarafa (approved):
LGTM

jskeet:
(Will leave this as-is until FF2 has been released.)

jskeet:
I've now updated this to v2

PR #2078, opened 2023-02-27
chore(storage): remove nested region tag

meredithslota:
Addressing this TODO: remove the storage_set_bucket_public_iam tag later as it is already declared in MakeBucketPublic.cs

### DIFF
--- a/storage/api/Storage.Samples/MakePublic.cs
+++ b/storage/api/Storage.Samples/MakePublic.cs
@@ -12,10 +12,7 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
-//TODO: remove the storage_set_bucket_public_iam tag later as it is already declared in MakeBucketPublic.cs
-
 // [START storage_make_public]
-// [START storage_set_bucket_public_iam]
 
 using Google.Apis.Storage.v1.Data;
 using Google.Cloud.Storage.V1;
@@ -38,5 +35,4 @@ public class MakePublicSample
     }
 }
 
-// [END storage_set_bucket_public_iam]
 // [END storage_make_public]


meredithslota:
The region tag that SnippetBot is concerned about also exists here: https://github.com/GoogleCloudPlatform/dotnet-docs-samples/blob/main/storage/api/Storage.Samples/MakeBucketPublic.cs so this is removal of a duplicate tag, not removal of the tag entirely.

amanda-tarafa:
The thing is that the documentation is linking to the file from which you are trying to remove the region tag. So you first need to update the docs to reference the other file, and then we can delete this one because SnippetBot will stop complaining.

See here: https://cloud.google.com/storage/docs/access-control/making-data-public#storage-make-object-public-csharp and then click on the "View on GitHub" button.

meredithslota:
Gotcha. This CL is pending: https://critique.corp.google.com/cl/512737328

amanda-tarafa (approved):

amanda-tarafa:
CL has been merged, we now need to wait for doc publication to catch up. I'll merge when that happens. Thanks!

amanda-tarafa:
@meredithslota not sure why the page is taking so long in using the new sample, let's give it another day and see what happens.

amanda-tarafa:
The documentation page has not yet been updated, even though the CL was merged a while back. SnippetBot is complaining and rightly so, at the moment, if we were to delete this region tag from this file, we would break documentation.

amanda-tarafa:
I've looked closer and the original cl only handled part of the changes needed. @meredithslota  I've sent another cl for review that completes the work. Once that's in, and the change propagates to docs, we'll be able to merge this.

amanda-tarafa:
SnippetBot is finally green. Merging.

PR #2103, opened 2023-03-25
chore(functions/firebase/FirebaseFirestore): Upgrade Google.Events.Protobuf to make sample gen2 compatible

kenneth-rosario:


functions/v2/firebase/FirebaseFirestore/Function.cs
```diff
@@ -0,0 +1,76 @@
...
+using Google.Protobuf.WellKnownTypes;
+using Microsoft.Extensions.Logging;
+using System.Collections.Generic;
+using System.Linq;
+using System.Threading;
+using System.Threading.Tasks;
+
+namespace FirebaseFirestore
+{
+    public class Function : ICloudEventFunction
```
jskeet, line 28: When 1.3.0 is out, you shouldn't have to do this. You should be able to use `ICloudEventFunction<DocumentEventData>`, even with binary data.

functions/v2/firebase/Firebase.Tests/FirebaseFirestoreTest.cs
```diff
@@ -0,0 +1,131 @@
...
+        }
+
+        [Fact]
+        public async Task MissingFieldsIgnored()
+        {
+            // There really would normally be Value or OldValue, but it's simpler to test
+            // a completely empty payload.
+            var data = new DocumentEventData();
+
+            ByteString eventDataByteString;
```
jskeet, line 102: You should be able to use just `var wrappedData = Any.Pack(data);` - although it's not clear to me why there's an `Any` involved in the first place.

functions/v2/firebase/FirebaseFirestore/Function.cs
```diff
@@ -0,0 +1,76 @@
...
+    {
+        private readonly ILogger _logger;
+
+        public Function(ILogger<Function> logger) =>
+            _logger = logger;
+
+        public Task HandleAsync(CloudEvent cloudEvent, CancellationToken cancellationToken)
+        {
+            Any anyWrapper = Any.Parser.ParseFrom((byte[])cloudEvent.Data);
```
jskeet, line 37: Why is the CloudEvent data wrapped in an Any? I'd expect it to be just the data itself, and a content type of application/protobuf. It may be worth pinging me internally about this...

functions/v2/firebase/Firebase.Tests/FirebaseFirestoreTest.cs
```diff
@@ -0,0 +1,131 @@
...
+                    Name = documentName,
+                    Fields =
+                    {
+                        { "highScore", new Value { IntegerValue = 20 } },
+                        { "name", new Value { StringValue = "Function Tester" } }
+                    }
+                }
+            };
+
+            ByteString eventDataByteString;
```
jskeet, line 55: As later on - just use `var wrappedData = Any.Pack(data);`

jskeet (commented):
Fundamentally confused by the use of an Any here. Please contact me internally about this.

functions/v2/firebase/FirebaseFirestore/Function.cs
jskeet: Note: this is being discussed/changed internally. The PR will need to change - I would *hope* that you'll be able to just use `ICloudEventFunction<DocumentEventData>` though.

functions/v2/firebase/FirebaseFirestore/Function.cs
kenneth-rosario: Updated version and tested. Everything worked as expected. Reducing PR to just upgrading to version 1.3.0 on existing sample

jskeet (approved):
Hooray - that's the sort of change I'm very happy to review :)

kenneth-rosario:
Added additional region tag to pair dotnet gen2 compatible sample with other firestore gen 2 compatible samples in other languages.

jskeet (approved):

functions/firebase/FirebaseFirestore/Function.cs
```diff
@@ -12,6 +12,7 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
+// [START functions_cloudevent_firebase_firestore]
 // [START functions_firebase_firestore]
```
amanda-tarafa, line 16: Will you update all the docs to reference this sample by the new region tag?
Once you've done that, we should remove the old region tag. Let me know and I'll create an issue so we don't forget about that.

amanda-tarafa (approved):
LGTM

functions/firebase/FirebaseFirestore/Function.cs
kenneth-rosario: We can leave the old tag for gen 1 docs and use the new one for gen 2 docs which reference all samples from all languages with the new tag.

kenneth-rosario:
I will need help merging this because I do not have write access to the repository.

functions/firebase/FirebaseFirestore/Function.cs
amanda-tarafa: SGTM

PR #2117, opened 2023-04-05
chore: update bigquery quickstart to fetch project id from env variable

hemanshv:
updated the big query sample to fetch project id from env variable.


jskeet (approved):
This change looks okay to me - I don't have background as to the rest of the sample.

Rishabh-V (approved):
LGTM

PR #2118, opened 2023-04-06
DLP: Added sample for Inspect custom regex

jainamtrivedi-crest:
- Added sample for [Inspect Custom Regex](https://cloud.google.com/dlp/docs/creating-custom-infotypes-regex#dlp-inspect-custom-regex-java) and unit tests

### DIFF
--- a/dlp/api/Snippets.Tests/InspectDataWithCustomRegexTests.cs
+++ b/dlp/api/Snippets.Tests/InspectDataWithCustomRegexTests.cs
@@ -1,0 +1,38 @@
+﻿// Copyright 2023 Google Inc.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License
+
+using Google.Cloud.Dlp.V2;
+using Xunit;
+
+namespace GoogleCloudSamples
+{
+    public class InspectDataWithCustomRegexTests : IClassFixture<DlpTestFixture>
+    {
+        private DlpTestFixture _fixture;
+        public InspectDataWithCustomRegexTests(DlpTestFixture fixture) => _fixture = fixture;
+
+        [Fact]
+        public void TestInspectDataWithCustomRegex()
+        {
+            string mrn = "444-5-22222";
+            string textToInspect = $"Patients MRN {mrn}";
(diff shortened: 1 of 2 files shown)


dlp/api/Snippets/InspectDataWithCustomRegex.cs
```diff
@@ -0,0 +1,79 @@
...
+using Google.Cloud.Dlp.V2;
+
+public class InspectDataWithCustomRegex
+{
+    public static InspectContentResponse InspectDataCustomRegex(string projectId, string text, string customRegex, InfoType infoType = null)
+    {
+        // Instantiate a client.
+        var dlp = DlpServiceClient.Create();
+
+        // Construct the byte content item.
```
ivanmed, line 28: I think there is no need for byteContentItem here, you can just assign the 'text' parameter directly to ContentItem.Value, something like:
var contentItem = new ContentItem { Value = text };

jainamtrivedi-crest: Removed the byte content item. In JAVA code sample, they have used byte content item; So, I used it here as well

ivanmed (approved):

dlp/api/Snippets.Tests/InspectDataWithCustomRegexTests.cs
```diff
@@ -0,0 +1,38 @@
...
+        private DlpTestFixture _fixture;
+        public InspectDataWithCustomRegexTests(DlpTestFixture fixture) => _fixture = fixture;
+
+        [Fact]
+        public void TestInspectDataWithCustomRegex()
+        {
+            string mrn = "444-5-22222";
+            string textToInspect = $"Patients MRN {mrn}";
+            string customRegexPattern = "[1-9]{3}-[1-9]{1}-[1-9]{5}";
+            InfoType infoType = new() { Name = "C_MRN" };
```
amanda-tarafa, line 31: tiny nit: Be consistent, you have been using: 
```suggestion
            var infoType = new InfoType { Name = "C_MRN" };
```

dlp/api/Snippets.Tests/InspectDataWithCustomRegexTests.cs
```diff
@@ -0,0 +1,38 @@
...
+            var result = InspectDataWithCustomRegex.InspectDataCustomRegex(_fixture.ProjectId, textToInspect, customRegexPattern, infoType);
+            var findings = result.Result.Findings;
+            Assert.Single(findings);
+            Assert.Equal(mrn, findings[0].Quote);
```
amanda-tarafa, line 35:
```suggestion
            var finding = Assert.Single(findings);
            Assert.Equal(mrn, finding.Quote);
```

amanda-tarafa (changes requested):
A couple of changed but otherwise looks good.

dlp/api/Snippets.Tests/InspectDataWithCustomRegexTests.cs
jainamtrivedi-crest: Changed.

amanda-tarafa (approved):
Thanks!

PR #2119, opened 2023-04-06
DLP: Added sample for deidentifying content with replacement

jainamtrivedi-crest:
- Added sample for [Deidentify content using replace config](https://cloud.google.com/dlp/docs/deidentify-sensitive-data.md#dlp_deidentify_replace-java) and unit tests

### DIFF
--- a/dlp/api/Snippets.Tests/DeidentifyWithReplacementTests.cs
+++ b/dlp/api/Snippets.Tests/DeidentifyWithReplacementTests.cs
@@ -1,0 +1,33 @@
+﻿// Copyright 2023 Google Inc.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+using Xunit;
+
+namespace GoogleCloudSamples
+{
+    public class DeidentifyWithReplacementTests : IClassFixture<DlpTestFixture>
+    {
+        private DlpTestFixture _fixture;
+
+        public DeidentifyWithReplacementTests(DlpTestFixture fixture) => _fixture = fixture;
+        [Fact]
+        public void TestDeidentifyWithReplace()
+        {
+            var text = "My name is Alicia Abernathy, and my email address is aabernathy@example.com and abc@gmail.com.";
+            var expected = "My name is Alicia Abernathy, and my email address is [email-address] and [email-address].";
+            var result = DeidentifyWithReplacement.Deidentify(_fixture.ProjectId, text);
+            Assert.Equal(expected, result.Item.Value);
+        }
+    }
+}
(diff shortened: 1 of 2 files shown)


dlp/api/Snippets.Tests/DeidentifyWithReplacementTests.cs
```diff
@@ -0,0 +1,37 @@
...
+        {
+            Fixture = fixture;
+        }
+
+        [Fact]
+        public void TestDeidentifyWithReplace()
+        {
+            var text = "My name is Alicia Abernathy, and my email address is aabernathy@example.com and abc@gmail.com.";
+            var result = DeidentifyWithReplacement.Deidentify(Fixture.ProjectId, text);
+            Assert.NotEqual(text, result.Item.Value);
```
soumya92, line 33: Can we Assert.Equal() here? The output should always be the same, and it will better demonstrate what the api result will be without needing to run the sample.

dlp/api/Snippets/DeidentifyWithReplacement.cs
```diff
@@ -0,0 +1,82 @@
...
+        // Construct the deidentify config using replace value config.
+        var deidentifyConfig = new DeidentifyConfig
+        {
+            InfoTypeTransformations = new InfoTypeTransformations
+            {
+                Transformations =
+                {
+                    new InfoTypeTransformations.Types.InfoTypeTransformation
+                    {
+                        InfoTypes = { infoTypes ?? new InfoType[] { new InfoType { Name = "EMAIL_ADDRESS" } } },
```
soumya92, line 54: Can we do this upfront (`infoTypes ??= ...`) instead of repeating this logic twice? That would make it easier to change the default value in the future and avoid mistakes where we update one location but not the other.

soumya92 (changes requested):

dlp/api/Snippets.Tests/DeidentifyWithReplacementTests.cs
jainamtrivedi-crest: Added expected text.

dlp/api/Snippets/DeidentifyWithReplacement.cs
jainamtrivedi-crest: Moved infotypes declaration to the top.

soumya92 (approved):

dlp/api/Snippets.Tests/DeidentifyWithReplacementTests.cs
```diff
@@ -0,0 +1,35 @@
...
+        public DeidentifyWithReplacementTests(DlpTestFixture fixture) => _fixture = fixture;
+        [Fact]
+        public void TestDeidentifyWithReplace()
+        {
+            var text = "My name is Alicia Abernathy, and my email address is aabernathy@example.com and abc@gmail.com.";
+            var expected = "My name is Alicia Abernathy, and my email address is [email-address] and [email-address].";
+            var result = DeidentifyWithReplacement.Deidentify(_fixture.ProjectId, text);
+            Assert.NotEqual(text, result.Item.Value);
+            Assert.Contains("[email-address]", result.Item.Value);
+            Assert.Equal(expected, result.Item.Value);
```
amanda-tarafa, line 32: These are all redundant, you really only need:

```suggestion
            Assert.Equal(expected, result.Item.Value);
```

amanda-tarafa (changes requested):

dlp/api/Snippets.Tests/DeidentifyWithReplacementTests.cs
jainamtrivedi-crest: Done.

amanda-tarafa (approved):
Thanks!

PR #2128, opened 2023-04-11
chore: updated runTests.ps1 file of bigquery Quickstart

hemanshv:
updated quick start to use the Google_Cloud_Project env variable directly.
made the corresponding changes in  runTest.ps1 as well. 
The test is running fine locally.

### DIFF
--- a/bigquery/api/SimpleApp/runTests.ps1
+++ b/bigquery/api/SimpleApp/runTests.ps1
@@ -11,12 +11,6 @@
 # WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied. See the
 # License for the specific language governing permissions and limitations under
 # the License.
-Import-Module ..\..\..\BuildTools.psm1 -DisableNameChecking
 
-
-BackupAndEdit-TextFile @(".\Program.cs") `
-    @{"YOUR-PROJECT-ID" = $env:GOOGLE_PROJECT_ID} `
-{
-    dotnet restore --force
-    dotnet test --no-restore --test-adapter-path:. --logger:junit 2>&1 | %{ "$_" }
-}
+dotnet restore --force
+dotnet test --no-restore --test-adapter-path:. --logger:junit 2>&1 | %{ "$_" }


Rishabh-V (commented):
I am not sure if there is any environment in which the test for this samples run, but if there is then without the new `GOOGLE_CLOUD_PROJECT` environment variable the tests would fail. This may be fine if we work and get this variable set. However, it does introduce inconsistencies in ProjectID environment variable used here and in other samples.

We have removed the `BackupAndEdit` command. However, one way could have been to use it to replace the environment variable with `GOOGLE_PROJECT_ID`. That way, we would have made changes needed for the BQ team and also kept our tests running with the same environment variable. Please discuss with Amanda and get an agreement. Thanks.

hemanshv:
Hi @amanda-tarafa ,
Please share your thoughts on the same.

Thanks,

amanda-tarafa:
Why is the environment variable change necessary? As @Rishabh-V mentioned, the environment variable we use through samples is `GOOGLE_PROJECT_ID`. Why should we add a new environment variable?
Let's discuss internally please.

amanda-tarafa:
Now that we are using an env variable, removing the replacement of the placeholder is a good change. It would have been better in #2117, but that's fine.

hemanshv:
Hi @Rishabh-V and @amanda-tarafa ,
I've udpated the PR. Please have a look and provide your approval.

Thanks,

bigquery/api/SimpleApp/runTests.ps1
```diff
@@ -11,12 +11,6 @@
...
 
-
-BackupAndEdit-TextFile @(".\Program.cs") `
-    @{"YOUR-PROJECT-ID" = $env:GOOGLE_PROJECT_ID} `
-{
-    dotnet restore --force
-    dotnet test --no-restore --test-adapter-path:. --logger:junit 2>&1 | %{ "$_" }
-}
+dotnet restore --force
+dotnet test --no-restore --test-adapter-path:. --logger:junit 2>&1 | %{ "$_" }
```
Rishabh-V, line 16: nit: We have removed the newline from the end of the file.

Rishabh-V (commented):
A nit but this looks fine to me. Please request review from Amanda after addressing/responding to it. Thanks.

bigquery/api/SimpleApp/runTests.ps1
hemanshv: Done

amanda-tarafa (approved):
Please rebase all changes into a single commit before merging. Thanks.

Rishabh-V (approved):

PR #2131, opened 2023-04-11
ci: Fix script so that CI runs in Windows again

### LINKED ISSUE #2101: Windows integration tests are not being run always
There seems to be a problem with how the diff command is written, and in some cases, it's not yielding folders to run tests for.

amanda-tarafa:
Fixes #2101


amanda-tarafa:
@jskeet some changes in Kokoro meant that the default git work dir was broken for Windows. Specifying it explicitly works well for both Linux and Windows and protects us against similar changes in the future.

jskeet (approved):

PR #2158, opened 2023-04-20
DLP: Added sample for inspect image with all infotypes

jainamtrivedi-crest:
- Added sample for [Inspect Image With All Infotypes](https://cloud.google.com/dlp/docs/inspecting-images.md#dlp-inspect-image-all-infotypes-java) with test cases

### DIFF
--- a/dlp/api/Snippets/InspectImageForSensitiveDataWithInfoTypes.cs
+++ b/dlp/api/Snippets/InspectImageForSensitiveDataWithInfoTypes.cs
@@ -1,0 +1,72 @@
+﻿// Copyright 2023 Google Inc.
+//
+// Licensed under the Apache License, Version 2.0 (the "License"); you may not
+// use this file except in compliance with the License. You may obtain a copy of
+// the License at
+//
+// http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS, WITHOUT
+// WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied. See the
+// License for the specific language governing permissions and limitations under
+// the License.
+
+// [START dlp_inspect_image_all_infotypes]
+
+using System;
+using System.IO;
+using System.Linq;
+using Google.Api.Gax.ResourceNames;
+using Google.Cloud.Dlp.V2;
+using Google.Protobuf;
+
+public class InspectImageForSensitiveDataWithInfoTypes
+{
+    public static InspectContentResponse InspectImage(
+        string projectId,
+        string filePath)
+    {
+        // Instantiate dlp client.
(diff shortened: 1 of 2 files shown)


dlp/api/Snippets/InspectImageForSensitiveDataWithInfoTypes.cs
```diff
@@ -0,0 +1,68 @@
...
+
+        // Call the API.
+        var response = dlp.InspectContent(request);
+
+        // Inspect the response.
+        var resultFindings = response.Result.Findings;
+
+        Console.WriteLine($"Findings: {resultFindings.Count}");
+
+        foreach (var f in resultFindings)
```
soumya92, line 57: Better output formatting?

Instead of producing output like:

```
Findings:
    Quote: q1
    InfoType: i1
    Likelihood: l1
    Quote: q2
    InfoType: i2
    Likelihood: l2
    Quote: q3
    InfoType: i3
    Likelihood: l3
```

can we add some spacing between entries? And maybe only indent quote and likelihood, so that the output looks more like:

```
Findings:

InfoType: i1
    Quote: q1
    Likelihood: l1

InfoType: i2
    Quote: q2
    Likelihood: l2

InfoType: i3
    Quote: q3
    Likelihood: l3
```


Also since this is an image, could you add co-ordinates as well to demonstrate that we provide that information? That would be `f.location.contentLocations[*].imageLocation.boundingBoxes[*]`. (There can be multiple results for a single infotype, hence arrays).

Ideally we could flatten them and present them as one more item in each finding, making it something like:

```
Findings:

InfoType: i1
    Quote: q1
    ImageLocations: [(left1,top1,+width1,+height1), ...] 
    Likelihood: l1

...
```

jainamtrivedi-crest: Regarding the output format, All the code samples follow same output format (pre-written); So, Just to be consistent about it, I have kept the output format like that.

jainamtrivedi-crest: image
this kind of output of ImageLocations you were asking?

soumya92: Sure, that format works too.

I don't know where we got the format from, but going forward we should make the output easier for users to understand. The lack of spacing makes it difficult to understand that VeryLikely is associated with PHONE_NUMBER and not PERSON_NAME...

jainamtrivedi-crest: Okay. Let me change the format for this example.

jainamtrivedi-crest: Done.

amanda-tarafa (approved):

PR #2169, opened 2023-05-02
DLP: Fixed test case in redact_image

jainamtrivedi-crest:
- Added `[Fact]` attribute to get recognized as test case by xUnit.


annadata (dismissed):
Thanks.

jainamtrivedi-crest:
As per the @annadata 's suggestion, I have created the new image file and named it as per the test case name.

PR #2174, opened 2023-05-05
feat(appengine): Delete unused samples

jahjahbin:
Delete GAE Flex samples which are not referenced by any GCP docs: `F#`, `SocialAuth`, and `WebSocketSignalR`.

Please note, I also added `WebSocket` project to the solution, because it somehow was not there before.


jahjahbin:
> Here is the summary of changes.
> 
> You are about to delete 2 region tags.
> This comment is generated by [snippet-bot](https://github.com/apps/snippet-bot). If you find problems with this result, please file an issue at: https://github.com/googleapis/repo-automation-bots/issues. To update this comment, add `snippet-bot:force-run` label or use the checkbox below:
> 
> * [ ]  Refresh this comment

I investigated this -- I don't see any GCP docs pointing to these tags. We're good.

jinglundong (approved):
We had a similar discussion about keeping the old code samples. Our conclusion was to differentiate old and new samples by branches, right? I'm wondering if we should create a branch to preserve these 3 samples (still delete them from the main branch as being done in this PR). I understand that they are in the commit history (of main), but a separate branch is more discoverable.

jahjahbin:
> We had a similar discussion about keeping the old code samples. Our conclusion was to differentiate old and new samples by branches, right? I'm wondering if we should create a branch to preserve these 3 samples (still delete them from the main branch as being done in this PR). I understand that they are in the commit history (of main), but a separate branch is more discoverable.

I believe the branch we created recently will suffice: `flex-dotnet3-and-earlier`

jinglundong:
> > We had a similar discussion about keeping the old code samples. Our conclusion was to differentiate old and new samples by branches, right? I'm wondering if we should create a branch to preserve these 3 samples (still delete them from the main branch as being done in this PR). I understand that they are in the commit history (of main), but a separate branch is more discoverable.
> 
> I believe the branch we created recently will suffice: `flex-dotnet3-and-earlier`

You are right. `flex-dotnet3-and-earlier` should be sufficient.

amanda-tarafa:
> I investigated this -- I don't see any GCP docs pointing to these tags. We're good.

Yes, if you attempt to delete a region tag in use, snippet-bot check will fail, and that's a prerequisite for being able to merge to PR. The comment from snippet-bot will be different as well highlighting the in-use regio tags that you are attempting to delete.

> I believe the branch we created recently will suffice: flex-dotnet3-and-earlier

Yes, these samples are also in that branch of course.

amanda-tarafa (approved):

PR #2123, opened 2023-04-06
DLP: Added sample for deidentify table using condition and infotypes

jainamtrivedi-crest:
- Added sample for [Deidentify Table Using Condition and Infotypes](https://cloud.google.com/dlp/docs/examples-deid-tables.md#transform_findings_only_when_specific_conditions_are_met_on_another_field) and unit tests

PR #2180, opened 2023-05-08
DLP: Added sample for deidentify data using dictionary replacements

jainamtrivedi-crest:
- Added sample for [Deidentify Data Using Dictionary Replacements](https://cloud.google.com/dlp/docs/transformations-reference#dictionary-replacement) and test cases

### DIFF
--- a/dlp/api/Snippets/DeidentifyDataUsingDictionaryReplacement.cs
+++ b/dlp/api/Snippets/DeidentifyDataUsingDictionaryReplacement.cs
@@ -1,0 +1,91 @@
+﻿// Copyright 2023 Google Inc.
+//
+// Licensed under the Apache License, Version 2.0 (the "License"); you may not
+// use this file except in compliance with the License. You may obtain a copy of
+// the License at
+//
+// http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS, WITHOUT
+// WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied. See the
+// License for the specific language governing permissions and limitations under
+// the License.
+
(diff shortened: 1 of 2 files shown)


dlp/api/Snippets/DeidentifyDataUsingDictionaryReplacement.cs
```diff
@@ -0,0 +1,91 @@
...
+using System.Collections.Generic;
+using Google.Api.Gax.ResourceNames;
+using Google.Cloud.Dlp.V2;
+
+public class DeidentifyDataUsingDictionaryReplacement
+{
+    public static DeidentifyContentResponse Deidentify(
+        string projectId,
+        string text,
+        IEnumerable<InfoType> infoTypes = null,
```
soumya92, line 27: Other samples accept infotype names and then map them in here, can we do the same for consistency?

soumya92 (changes requested):

dlp/api/Snippets/DeidentifyDataUsingDictionaryReplacement.cs
jainamtrivedi-crest: The samples which we are creating now follows this process. And this process is widely used in the samples apart from passing just names and iterating over them to create the `InfoType`. 

These are few of the samples where we have used this method:
- https://github.com/GoogleCloudPlatform/dotnet-docs-samples/pull/2160 
- https://github.com/GoogleCloudPlatform/dotnet-docs-samples/pull/2121
- https://github.com/GoogleCloudPlatform/dotnet-docs-samples/pull/2123

soumya92: Okay, might be a difference between languages then.

soumya92 (approved):

amanda-tarafa (approved):

PR #2196, opened 2023-05-17
feat(appengine): Update Metadata sample to .NET 6

jahjahbin:


appengine/flexible/Metadata/Metadata.Sample/Program.cs
```diff
@@ -0,0 +1,53 @@
...
+            // [END gae_flex_metadata]
+        }
+        catch (System.Net.Http.HttpRequestException e)
+        {
+            response = "<not set>";
+        }
+        await context.Response.WriteAsync(string.Format(
+            "Public IP address: {0}",
+	    response));
+    }
```
amanda-tarafa, line 46: I'm wondering if we shouldn't do any of this, in this sample and the others that make an outbound call. Basically, if we should always call `response.EnsureSuccessStatusCode()` so that the sample method throws if there was an error . And tests would fail in that case, whereas they are not failing now because what we are really testing for is that the call `context.Response.WriteAsync` succeeds.

Of course that would mean that all of these samples that need to be actually running in App Engine Flex will fail in local or any other environment. What do you think?

amanda-tarafa (approved):
LGTM, my comment is not blocking and we can make related changes in a separate PR for all the samples that have already been updated, if we decide to make such changes, I'm not certain at the moment.

You'll need a review from App Engine Flex team.

appengine/flexible/Metadata/Metadata.Sample/Program.cs
jahjahbin: I'm open to changing it, but is there a particular reason we'd want to change it? I'm a bit hesitant to iterate on the sample code because it's just supposed to be a sample to start with, anyway.

amanda-tarafa: Yes you are right. We should be testing that the sample as a whole works, and that's what we are doing at the moment. It's not for samples to test for actual product features, it's definetely a plus if they do (and we have caught some API issues that way), but product features should be covered by product tests regardless.
Let's leave it as, we can always re-evaluate later.

jahjahbin: Ok, sounds good. I pinged a serverless-runtimes reviewer this morning, so just waiting on that now.

jinglundong: Adding my 2 cents. I agree with the general ideas. The tests in this repo are responsible for guarding the regressions in this repo. Detecting service and server-side regressions is out-of-scope in my opinion. We have unit tests and e2e tests in our CI/CD to prevent regressions and outages. This repo is not the right place to do so, because it's too late in the release process.

jinglundong (approved):

PR #2197, opened 2023-05-22
feat(appengine): Delete unused Redis sample

jahjahbin:
I feel bad deleting all this work, but it is *not used* in the GCP docs.

I was told the snippet-bot will block the PR on the region tag deletion -- not sure how to handle that when it comes up...


jinglundong (approved):

amanda-tarafa:
> I feel bad deleting all this work, but it is _not used_ in the GCP docs.
> 
> I was told the snippet-bot will block the PR on the region tag deletion -- not sure how to handle that when it comes up...

The code remains in the repo history so it's not fully lost.

SnippetBot only complains if the tag is used in docs, else it will let you delete it.

amanda-tarafa (approved):

PR #2201, opened 2023-05-23
chore(endpoints): Update region tag with product prefix

arbrown:


arbrown:
Freezing references to snippet in cl/534553109. Please do not merge until that is submitted (or let me take care of it.) Will un-freeze references once this PR has merged.

endpoints/getting-started/src/IO.Swagger/app.yaml
```diff
@@ -1,7 +1,7 @@
 runtime: aspnetcore
 env: flex
 
-# [START configuration]
+# [START endpoints_configuration]
```
amanda-tarafa, line 4: The region tag "configuration" is being used so you cannot just change it (that's why SnippetBot is complaining about). You first need to add the new tag on a PR, once that's merged, you need to update the docs that use the wrong tag to use the new tag. Once that CL is submitted, you need to create a new PR that deletes the old tag.

amanda-tarafa (changes requested):

endpoints/getting-started/src/IO.Swagger/app.yaml
arbrown: Hi Amanda,

I approached this from the other angle and submitted a CL to freeze references to the snippet to a known CL and will change references once this is merged.  Having both in for now is fine though.

Thanks!

amanda-tarafa (approved):
SnippetBot wouldn't let me merge, it does not know about freezing the CL ;)

PR #2203, opened 2023-05-24
feat(appengine): Update Pubsub sample to .NET 6

jahjahbin:
One of the last two samples to migrate. Please let me know if the environment variable approach here is fine, or if appsettings.json would be simpler.


appengine/flexible/Pubsub/Pubsub.Sample/Program.cs
```diff
@@ -0,0 +1,125 @@
...
+        builder.Services.AddSingleton(PublisherClient.Create(topic));
+        builder.Services.AddMvc();
+        App = builder.Build();
+        App.UseStaticFiles();
+        App.MapControllerRoute(
+            name: "default",
+            pattern: "{controller=Home}/{action=Index}/{id?}");
+    }
+
+    static TopicName CreateTopicAndSubscription(PubsubOptions options)
```
jahjahbin, line 50: Note, this and the next method are both lifted from Startup.cs. I could refactor if it makes sense.

appengine/flexible/Pubsub/Pubsub.Sample/app.yaml
```diff
@@ -0,0 +1,13 @@
+﻿runtime: aspnetcore
+env: flex
+
+runtime_config:
+  operating_system: ubuntu22
+
+env_variables:
+  TEST_PROJECT_ID: your-project-id
+  TEST_VERIFICATION_TOKEN: your-token
```
jinglundong, line 9: In the previous .NET 2 snippet shown at https://cloud.google.com/appengine/docs/flexible/writing-and-responding-to-pub-sub-messages?tab=.net#setting_environment_variables, these env variables are set in the `appsettings.json` file. Do we need to update the user guide to match this code sample?

jinglundong (commented):
How do we add this .NET 6 sample to the user guide https://screenshot.googleplex.com/BF3XJK347AsD5aJ? Are we going to replace the .NET 2 sample in place or add another (nested) tab?

jahjahbin:
> How do we add this .NET 6 sample to the [user guide](https://cloud.google.com/appengine/docs/flexible/writing-and-responding-to-pub-sub-messages?tab=.net#clone_the_sample_app)? Are we going to replace the .NET 2 sample in place or add another (nested) tab?

> user guide

This doesn't seem like a comment on any code in this PR, just a general question. I think the plan is to add a nested tab.

appengine/flexible/Pubsub/Pubsub.Sample/app.yaml
jahjahbin: Sure. I would prefer using the appsettings personally, but I think this repo has a convention to just put things in envars instead of using the appsettings file. I believe the docs change will require more work because of this choice.

jinglundong: Sure, either works for me. I just want to bring up a point to keep the docs and samples in sync. .NET 2 user guide uses `appsettings.json`; .NET 6 user guide uses `app.yaml`. The corresponding descriptions or instructions need to reflect that.

jinglundong:
> > How do we add this .NET 6 sample to the [user guide](https://cloud.google.com/appengine/docs/flexible/writing-and-responding-to-pub-sub-messages?tab=.net#clone_the_sample_app)? Are we going to replace the .NET 2 sample in place or add another (nested) tab?
> 
> > user guide
> 
> This doesn't seem like a comment on any code in this PR, just a general question. I think the plan is to add a nested tab.

Right, this is a general question. Having a nested tab sounds like a good plan. Just need to work with the Docs team to differentiate gaps in the configs like `appsettings.json` and `app.yaml` (that we are discussing in the other thread).

jinglundong (approved):

appengine/flexible/Pubsub/Pubsub.Sample/Program.cs
```diff
@@ -0,0 +1,132 @@
...
+        try
+        {
+            subscriber.ModifyPushConfig(subscriptionName, pushConfig);
+        }
+        catch (Grpc.Core.RpcException e)
+        when (e.Status.StatusCode == Grpc.Core.StatusCode.PermissionDenied)
+        {
+            Console.WriteLine("Could not modify subscription {0}", subscriptionName);
+        }
+        return;
```
amanda-tarafa, line 124:
```suggestion

```

appengine/flexible/Pubsub/Pubsub.Sample/Program.cs
amanda-tarafa: The error handling in particular seems a little off. You know better what the goal of this sample is, but I guess these methods can be made to be cleaner (and clearer).

appengine/flexible/Pubsub/Pubsub.Sample/Controllers/HomeController.cs
```diff
@@ -63,7 +63,7 @@ public IActionResult Index(MessageForm messageForm)
                     Data = ByteString.CopyFromUtf8(messageForm.Message)
                 };
                 pubsubMessage.Attributes["token"] = _options.VerificationToken;
-                _publisher.PublishAsync(pubsubMessage);
+                _publisher.PublishAsync(pubsubMessage).Wait();
```
amanda-tarafa, line 66:
```suggestion
                await _publisher.PublishAsync(pubsubMessage);

```

And then the method signature needs to change to:

```
public async Task<IActionResult> IndexAsync(...)
```

appengine/flexible/Pubsub/Pubsub.Sample/app.yaml
amanda-tarafa: There are two separate aspects here:

- How we test the sample, and we set environment variables for that in runTests.ps1
- How users deploy the application to AppEngine Flex, and we are setting env variable for that in app.yaml

But, the good thing of using .NET Configuration mechanisms in Program.cs, is that the code does not need to change whether the values are specified through env vars or `appsettings.json` and this is common knowledge (or should more or less be) amongst .NET developers.

So, the point is that on the sample documentation, you can say something like "Env variables with required values are being added in app.yaml for deployment, but you can instead use appsettings.json."

amanda-tarafa (changes requested):
Main change is for the `async` bit.

Refactoring and cleaning the setup methods is a nice to have.

jinglundong (approved):

appengine/flexible/Pubsub/Pubsub.Sample/Program.cs
jahjahbin: I ended up actually *deleting* these, because the public docs have you create these resources anyway (so no need to do them in code too, I feel like): https://cloud.google.com/appengine/docs/flexible/writing-and-responding-to-pub-sub-messages?tab=.net

appengine/flexible/Pubsub/Pubsub.Sample/app.yaml
jahjahbin: We'll point to the vars listed in app.yaml then.

appengine/flexible/Pubsub/Pubsub.Sample/Program.cs
```diff
@@ -0,0 +1,132 @@
...
+    {
+        var subscriber = SubscriberServiceApiClient.Create();
+        try
+        {
+            subscriber.CreateSubscription(subscriptionName, topicName,
+                pushConfig, 20);
+            return;
+        }
+        catch (Grpc.Core.RpcException e)
+        when (e.Status.StatusCode == Grpc.Core.StatusCode.AlreadyExists ||
```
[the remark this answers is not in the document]
jahjahbin: I think it was probably intended as a way to make the creation process smoother? Not sure.

appengine/flexible/Pubsub/Pubsub.Sample/Program.cs
jahjahbin: Done, in a different way.

appengine/flexible/Pubsub/Pubsub.Sample/Controllers/HomeController.cs
jahjahbin: Done.

PR #2204, opened 2023-05-24
feat(appengine): Delete AntiForgery sample

jahjahbin:
Context: I saw [this blog](https://medium.com/google-cloud/antiforgery-tokens-asp-net-core-and-google-cloud-7ac6a5c7842b) which provides a motivating example for the sample. I gave it a shot on GAE Flex without doing any of that Storage or KMS stuff, but keeping the `ValidateAntiForgeryToken` attribute and `AddDataProtection()`. It _still_ worked. I saw the custom home page, entered some text, clicked submit (http post) and got the response with no errors.

To me, this sample is ok to remove -- perhaps it was inspired by trying to work around a shortcoming in GAE Standard? (yes, even though the folder is `flexible`).

I'll work on removing the GCP docs for this.


amanda-tarafa:
FYI @hemanshv and @Rishabh-V:

https://github.com/GoogleCloudPlatform/dotnet-docs-samples/pull/2204#issue-1724775528 is maybe relevant for the discussion around the future of https://github.com/hemanshv/google-cloud-dotnet-powerpack

jinglundong (approved):
Sure, we can delete this sample and the doc if they aren't being supported. There's a chance that this sample was used for Flex Managed VM or Flex Compat mode which had access to the bundled Standard APIs. These two modes have been deprecated and turned down before 2020.

PR #2205, opened 2023-05-25
DLP: Added sample for inspect big query table with SCC integration

jainamtrivedi-crest:
- Added sample for [Inspect Big Query Data and Send to SCC](https://cloud.google.com/dlp/docs/sending-results-to-scc.md#bigquery-data) with test cases

### DIFF
--- a/dlp/api/Snippets.Tests/InspectBigQueryWithSCCIntegrationTests.cs
+++ b/dlp/api/Snippets.Tests/InspectBigQueryWithSCCIntegrationTests.cs
@@ -1,0 +1,32 @@
+// Copyright 2023 Google Inc.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License
+
+using System.IO;
+using Xunit;
+
+namespace GoogleCloudSamples
+{
+    public class InspectBigQueryWithSCCIntegrationTests : IClassFixture<DlpTestFixture>
+    {
+        private DlpTestFixture _fixture;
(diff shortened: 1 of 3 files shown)


jainamtrivedi-crest:
We will need `Moq` framework for mocking the data in the test cases.

annadata (approved):

amanda-tarafa:
Please fix conflicts, tests can't be run before that happens. Thanks

amanda-tarafa:
Tests timed out, @jainamtrivedi-crest can you increase the time out again please?

annadata:
> > Tests timed out, @jainamtrivedi-crest can you increase the time out again please?
> 
> @amanda-tarafa I have increased the timeout to 1800. Thanks.

@jainamtrivedi-crest This sample is related to inspect jobs and they can take time to get queued/run. Can we move to mocks in the tests?

jainamtrivedi-crest:
TEst

jainamtrivedi-crest:
@annadata I have removed mocking approach and asserted the console output. Please check.

dlp/api/Snippets.Tests/InspectBigQueryWithSCCIntegrationTests.cs
```diff
@@ -0,0 +1,36 @@
...
+    public class InspectBigQueryWithSCCIntegrationTests : IClassFixture<DlpTestFixture>
+    {
+        private DlpTestFixture _fixture;
+        public InspectBigQueryWithSCCIntegrationTests(DlpTestFixture fixture) => _fixture = fixture;
+
+        [Fact]
+        public void TestCreate()
+        {
+            var stringWriter = new StringWriter();
+            System.Console.SetOut(stringWriter);
```
amanda-tarafa, line 29: If you do this, you change the standard out for the whole process, i.e. for all tests coming afterwards. It's also prone to tests overlapping and failing if several tests do the same thing.
There's no need to tests the stdoutput and we have moved away from that in the .NET repo.
If you want to assert that there has been no error, in this case, you just call the sample method, if it succeeds, there's no error, else it will throw an exception and the test will fail.

And if you are doing this in other tests, please change those as well.

amanda-tarafa (changes requested):

dlp/api/Snippets.Tests/InspectBigQueryWithSCCIntegrationTests.cs
jainamtrivedi-crest: Thank you for this context and I have incorporated your suggestion.

amanda-tarafa (approved):

PR #2207, opened 2023-05-25
DLP: Added sample of inspect data using hybrid job trigger

jainamtrivedi-crest:
- Added sample of [Inspect Data Using Hybrid Job Trigger](https://cloud.google.com/dlp/docs/how-to-hybrid-jobs#formatting) and test cases

### DIFF
--- a/dlp/api/Snippets.Tests/SendDataToTheHybridJobTriggerTests.cs
+++ b/dlp/api/Snippets.Tests/SendDataToTheHybridJobTriggerTests.cs
@@ -1,0 +1,93 @@
+// Copyright 2023 Google Inc.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License
+
+using Google.Cloud.Dlp.V2;
+using Xunit;
+using System.IO;
+using Google.Api.Gax.ResourceNames;
+
+namespace GoogleCloudSamples
+{
+    public class SendDataToTheHybridJobTriggerTests : IClassFixture<DlpTestFixture>
+    {
+        private DlpTestFixture _fixture;
+        public SendDataToTheHybridJobTriggerTests(DlpTestFixture fixture) => _fixture = fixture;
+
+        [Fact]
+        public void TestHybridJob()
+        {
+            var dlp = DlpServiceClient.Create();
+
+            // Create the Job Trigger.
+            var request = new CreateJobTriggerRequest
+            {
+                ParentAsLocationName = new LocationName(_fixture.ProjectId, "global"),
+                JobTrigger = new JobTrigger
+                {
+                    Triggers =
+                    {
+                        new JobTrigger.Types.Trigger
+                        {
+                            Manual = new Manual()
+                        }
+                    },
+                    InspectJob = new InspectJobConfig
+                    {
+                        Actions =
+                        {
+                            new Action
+                            {
+                                JobNotificationEmails = new Action.Types.JobNotificationEmails(),
+                                PublishToStackdriver = new Action.Types.PublishToStackdriver(),
+                            }
+                        },
+                        InspectConfig = new InspectConfig
+                        {
+                            InfoTypes =
+                            {
+                                new InfoType[]
+                                {
+                                    new InfoType { Name = "EMAIL_ADDRESS" }
+                                }
+                            },
+                            MinLikelihood = Likelihood.Possible,
+                            IncludeQuote = true,
+                        },
+                        StorageConfig = new StorageConfig
+                        {
+                            HybridOptions = new HybridOptions
+                            {
+                                Description = "Hybrid job trigger for data from the comments field of a table that contains customer appointment bookings",
+                            }
+                        }
+                    }
+                }
+            };
+            var hybridJobTrigger = dlp.CreateJobTrigger(request);
+
+            var text = "My email is test@example.org and name is Ariel.";
+
+            DlpJob result = SendDataToTheHybridJobTrigger.SendToHybridJobTrigger(_fixture.ProjectId, hybridJobTrigger.JobTriggerName.JobTriggerId, text);
+
+            // Delete the activated job.
+            JobsDelete.DeleteJob(result.Name);
+
+            // Delete the created job trigger.
+            dlp.DeleteJobTrigger(new DeleteJobTriggerRequest
+            {
+                Name = hybridJobTrigger.Name
+            });
+        }
+    }
+}
--- a/dlp/api/Snippets/SendDataToTheHybridJobTrigger.cs
+++ b/dlp/api/Snippets/SendDataToTheHybridJobTrigger.cs
@@ -1,0 +1,99 @@
+// Copyright 2023 Google Inc.
+//
+// Licensed under the Apache License, Version 2.0 (the "License"); you may not
+// use this file except in compliance with the License. You may obtain a copy of
+// the License at
+//
+// http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS, WITHOUT
+// WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied. See the
+// License for the specific language governing permissions and limitations under
+// the License.
+
+// [START dlp_inspect_send_data_to_hybrid_job_trigger]
+
+using System;
+using Google.Api.Gax.ResourceNames;
+using Google.Api.Gax;
+using Google.Cloud.Dlp.V2;
+using Grpc.Core;
+
+public class SendDataToTheHybridJobTrigger
+{
+    public static DlpJob SendToHybridJobTrigger(
+       string projectId,
+       string jobTriggerId,
+       string text = null)
+    {
+        // Instantiate the dlp client.
+        var dlp = DlpServiceClient.Create();
+
+        // Construct the hybrid finding details which will be used as metadata with the content.
+        // Refer to this for more information: https://cloud.google.com/dlp/docs/reference/rpc/google.privacy.dlp.v2#google.privacy.dlp.v2.Container
+        var findingDetails = new HybridFindingDetails
+        {
+            ContainerDetails = new Container
+            {
+                FullPath = "10.0.0.2:logs1:aap1",
+                RelativePath = "app1",
+                RootPath = "10.0.0.2:logs1",
+                Type = "System Logs"
+            }
+        };
+
+        // Construct the hybrid content item using the finding details and text to be inspected.
+        var hybridContentItem = new HybridContentItem
+        {
+            Item = new ContentItem { Value = text ?? "My email is ariel@example.org and name is Ariel." },
+            FindingDetails = findingDetails
+        };
+
+        var jobTriggerName = new JobTriggerName(projectId, jobTriggerId);
+
+        // Construct the request to activate the Job Trigger.
+        var activate = new ActivateJobTriggerRequest
+        {
+            JobTriggerName = jobTriggerName
+        };
+
+        DlpJob triggerJob = null;
+
+        try
+        {
+            // Call the API to activate the trigger.
+            triggerJob = dlp.ActivateJobTrigger(activate);
+        }
+        catch (RpcException)
+        {
+            ListDlpJobsRequest listJobsRequest = new ListDlpJobsRequest
+            {
+                ParentAsLocationName = new LocationName(projectId, "global"),
+                Filter = $"trigger_name={jobTriggerName}"
+            };
+
+            PagedEnumerable<ListDlpJobsResponse, DlpJob> res = dlp.ListDlpJobs(listJobsRequest);
+            foreach (DlpJob j in res)
+            {
+                triggerJob = j;
+            }
+        }
+
+        // Construct the request using hybrid content item.
+        var request = new HybridInspectJobTriggerRequest
+        {
+            HybridItem = hybridContentItem,
+            JobTriggerName = jobTriggerName
+        };
+
+        // Call the API.
+        HybridInspectResponse _ = dlp.HybridInspectJobTrigger(request);
+
+        Console.WriteLine($"Hybrid job created successfully. Job name: {triggerJob.Name}");
+
+        return triggerJob;
+    }
+}
+
+// [END dlp_inspect_send_data_to_hybrid_job_trigger]


annadata (approved):

dlp/api/Snippets/SendDataToTheHybridJobTrigger.cs
```diff
@@ -0,0 +1,97 @@
...
+
+        // Check the activated job status to fetch the inspect details.
+        do
+        {
+            triggerJob = dlp.GetDlpJob(new GetDlpJobRequest
+            {
+                Name = triggerJob.Name
+            });
+            Thread.Sleep(5000);
+        } while (triggerJob.InspectDetails.Result.ProcessedBytes <= 0);
```
amanda-tarafa, line 81: Can this really be less than zero? If not, it's better to only check for it being zero to avoind confusing users.

dlp/api/Snippets/SendDataToTheHybridJobTrigger.cs
```diff
@@ -0,0 +1,97 @@
...
+        {
+            HybridItem = hybridContentItem,
+            JobTriggerName = new JobTriggerName(projectId, jobId),
+            Name = jobPath,
+        };
+
+        // Call the API.
+        HybridInspectResponse response = dlp.HybridInspectJobTrigger(request);
+
+        // Check the activated job status to fetch the inspect details.
```
amanda-tarafa, line 73: Wouldn't this be demonstrated as well without the iteration?

I think the sample as it is its given the impression that for the status etc, to be valid, you need the job to have processes some bytes? If you just get the job once and print it status etc. wouldn't that be demonstration enough of how to use the library?

dlp/api/Snippets.Tests/SendDataToTheHybridJobTriggerTests.cs
```diff
@@ -0,0 +1,94 @@
...
+                            MinLikelihood = Likelihood.Possible,
+                            IncludeQuote = true,
+                        },
+                        StorageConfig = new StorageConfig
+                        {
+                            HybridOptions = new HybridOptions
+                            {
+                                Description = "Hybrid job trigger for data from the comments field of a table that contains customer appointment bookings",
+                            },
+
```
amanda-tarafa, line 72:
```suggestion
```

dlp/api/Snippets.Tests/SendDataToTheHybridJobTriggerTests.cs
```diff
@@ -0,0 +1,94 @@
...
+                        }
+                    }
+                }
+            };
+
+            var hybridJobTrigger = dlp.CreateJobTrigger(request);
+            var text = "My email is ariel@example.org and name is Ariel.";
+            var result = SendDataToTheHybridJobTrigger.SendToHybridJobTrigger(_fixture.ProjectId, hybridJobTrigger.JobTriggerName.JobTriggerId, hybridJobTrigger.Name, text);
+            Assert.Equal(DlpJob.Types.JobState.Active, result.State);
```
amanda-tarafa, line 81: This might cause flakiness, as maybe the job has completed the first time you query it? Or is that not possible.

dlp/api/Snippets.Tests/SendDataToTheHybridJobTriggerTests.cs
```diff
@@ -0,0 +1,94 @@
...
+            Assert.Contains(result.InspectDetails.Result.InfoTypeStats, f => f.InfoType.Name == "EMAIL_ADDRESS");
+
+            // Delete the created job trigger.
+            dlp.DeleteJobTrigger(new DeleteJobTriggerRequest
+            {
+                Name = hybridJobTrigger.Name
+            });
+
+            // Delete the activated job.
+            JobsDelete.DeleteJob(result.Name);
```
amanda-tarafa, line 91: This need to be on a finally on a try catch block. Else, if the test fails we won't delete the job.

dlp/api/Snippets.Tests/SendDataToTheHybridJobTriggerTests.cs
```diff
@@ -0,0 +1,94 @@
...
+        private DlpTestFixture _fixture;
+        public SendDataToTheHybridJobTriggerTests(DlpTestFixture fixture) => _fixture = fixture;
+
+        [Fact]
+        public void TestHybridJob()
+        {
+            var dlp = DlpServiceClient.Create();
+
+            // Create the Job Trigger.
+            var request = new CreateJobTriggerRequest
```
amanda-tarafa, line 32: Why is this operation included in the test, but no on the sample?

amanda-tarafa (changes requested):
Apart from the specific comments, would it be possible to split this sample into several? We tend to prefer samples that focus on a single RPC

dlp/api/Snippets/SendDataToTheHybridJobTrigger.cs
jainamtrivedi-crest: The response of  `dlp.HybridInspectJobTrigger(request);` is `{ }`. And if we just get the job and does not wait for the job to get completed; It will not show the different findings that job has found. Without waiting, the job state would be `Pending` or `Running` which will not help much while asserting in the test cases. That is why we have to wait (iterations) for the job to get completed

dlp/api/Snippets.Tests/SendDataToTheHybridJobTriggerTests.cs
jainamtrivedi-crest: No, The job created inside the hybrid job will always have state as `Active` and also we are waiting for the job to get completed in sample as well. I don't think this will cause flakiness

dlp/api/Snippets.Tests/SendDataToTheHybridJobTriggerTests.cs
jainamtrivedi-crest: The sample is just for Inspecting the data using hybrid job trigger and for that we have to pass the hybrid job trigger id and hybrid job trigger name. So its better to keep the sample as per the use case and create the hybrid job trigger inside the test case.

dlp/api/Snippets/SendDataToTheHybridJobTrigger.cs
```diff
@@ -0,0 +1,97 @@
...
+       string projectId,
+       string jobId,
+       string jobPath,
+       string text = null)
+    {
+        // Instantiate the dlp client.
+        var dlp = DlpServiceClient.Create();
+
+        // Construct the hybrid finding details which will be used as metadata with the content.
+        // Refer to this for more information: https://cloud.google.com/dlp/docs/reference/rest/v2/Container
```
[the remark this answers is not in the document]
jainamtrivedi-crest: Done

dlp/api/Snippets/SendDataToTheHybridJobTrigger.cs
jainamtrivedi-crest: Changed the condition. Thanks.

jainamtrivedi-crest:
> Apart from the specific comments, would it be possible to split this sample into several? We tend to prefer samples that focus on a single RPC

We have used 2nd RPC call to get the status and findings of the Job from the text that we want to inspect. IMO, without actual job results, the sample will just demonstrate how to use the API and will not show the findings.

dlp/api/Snippets/SendDataToTheHybridJobTrigger.cs
```diff
@@ -0,0 +1,103 @@
...
+        // Construct the hybrid content item using the finding details and text to be inspected.
+        var hybridContentItem = new HybridContentItem
+        {
+            Item = new ContentItem { Value = text ?? "My email is ariel@example.org and name is Ariel." },
+            FindingDetails = findingDetails
+        };
+
+        var jobTriggerName = new JobTriggerName(projectId, jobId);
+
+        // Activate the Job Trigger.
```
amanda-tarafa, line 55:
```suggestion
        // Construct the request to activate the Job Trigger.
```

dlp/api/Snippets/SendDataToTheHybridJobTrigger.cs
```diff
@@ -0,0 +1,103 @@
...
+        var activate = new ActivateJobTriggerRequest
+        {
+            JobTriggerName = jobTriggerName,
+            Name = jobPath
+        };
+
+        // Call the API to activate the trigger.
+        DlpJob triggerJob = dlp.ActivateJobTrigger(activate);
+
+        // Construct the request using hybrid content item.
```
amanda-tarafa, line 65:
```suggestion
        // Construct the request to inspect the job trigger using hybrid content item.
```

dlp/api/Snippets/SendDataToTheHybridJobTrigger.cs
amanda-tarafa: Then check the job is in a done state, instead of checking if it has processed some bytes. It's clearer.

amanda-tarafa: As for waiting on the sample so that you can get your tests to pass, that shoouldn't be necessary. You can wait in the test itself.

If you want to show the ser how to wait for the job to be completed, then that's fine, but do not add code to a sample for the sake of the tests.

dlp/api/Snippets/SendDataToTheHybridJobTrigger.cs
```diff
@@ -0,0 +1,103 @@
...
+        {
+            triggerJob = dlp.GetDlpJob(new GetDlpJobRequest
+            {
+                Name = triggerJob.Name
+            });
+            if (triggerJob.InspectDetails.Result.ProcessedBytes == 0)
+            {
+                Thread.Sleep(5000);
+            }
+        } while (triggerJob.InspectDetails.Result.ProcessedBytes == 0);
```
amanda-tarafa, line 87:
```suggestion
        while (true)
        {
            triggerJob = dlp.GetDlpJob(triggerJob.Name);
            if (triggerJob.State == JobState.Done)
            {
                break;
            }
            // Waiting 5 seconds on each iteration is for demonstration purposes only.
            // You may want to use a more robust retry strategy, with backoff for instance.
            Thread.Sleep(5000);
        }
```

dlp/api/Snippets/SendDataToTheHybridJobTrigger.cs
```diff
@@ -0,0 +1,103 @@
...
+
+        // Parse the response.
```
amanda-tarafa, line 89:
```suggestion
        // Inspect the response.
```

dlp/api/Snippets.Tests/SendDataToTheHybridJobTriggerTests.cs
amanda-tarafa: Well, because you are now using mocks. When I reviewed, you weren't.

And on the sample you are not currently waiting for the job to be completed (that's what I'm proposing in my suggestion), you are waiting for the job to process some amount of bytes.

amanda-tarafa (changes requested):
I don't see why we have to use mocks here. This is no different than other samples in the repo that require a resource (job) to be created on setup and deleted on tear down.
Unless you have a very strong reason for that, I'd prefer to rever this test to true integration.

dlp/api/Snippets/SendDataToTheHybridJobTrigger.cs
jainamtrivedi-crest: The job inside hybrid job trigger does not contain the status `Done`. Whenever it finishes the job the status will be `Active`

dlp/api/Snippets.Tests/SendDataToTheHybridJobTriggerTests.cs
jainamtrivedi-crest: The hybrid job does not have status done; So to check whether the job has completed or we are waiting for the job to process the bytes.

dlp/api/Snippets/SendDataToTheHybridJobTrigger.cs
jainamtrivedi-crest: State of the hybrid job will not change to `Done`.

dlp/api/Snippets/SendDataToTheHybridJobTrigger.cs
jainamtrivedi-crest: Its not for the sake of tests, we are mocking the response anyways. To provide more information about the inspection details, we have waited for the job to process the bytes and showing the results.

dlp/api/Snippets/SendDataToTheHybridJobTrigger.cs
jainamtrivedi-crest: Done

dlp/api/Snippets/SendDataToTheHybridJobTrigger.cs
jainamtrivedi-crest: Done

dlp/api/Snippets/SendDataToTheHybridJobTrigger.cs
jainamtrivedi-crest: Done

jainamtrivedi-crest:
> I don't see why we have to use mocks here. This is no different than other samples in the repo that require a resource (job) to be created on setup and deleted on tear down. Unless you have a very strong reason for that, I'd prefer to rever this test to true integration.

There were some of the intermittent failure where job was not getting completed within the expected time and CI was failing. Because of that we have decided to go with the mocking approach. This was already discussed with @annadata that we will be using mocking for few of the samples where inspection is taking time.

annadata:
@jainamtrivedi-crest 
Based on discussions we have decided to not use mocking as much as possible in C# DLP samples.
Please update the sample code to gracefully detect "The job is already active" scenario  and send out for review again.

jainamtrivedi-crest:
> @jainamtrivedi-crest Based on discussions we have decided to not use mocking as much as possible in C# DLP samples. Please update the sample code to gracefully detect "The job is already active" scenario and send out for review again.

@annadata @amanda-tarafa I have removed the mock and handled the already active scenario plus all the tests are passing, could you please re-review this one. Thanks.

dlp/api/Snippets.Tests/SendDataToTheHybridJobTriggerTests.cs
```diff
@@ -0,0 +1,93 @@
...
+            DlpJob result = SendDataToTheHybridJobTrigger.SendToHybridJobTrigger(_fixture.ProjectId, hybridJobTrigger.JobTriggerName.JobTriggerId, text);
+
+            // Delete the created job trigger.
+            dlp.DeleteJobTrigger(new DeleteJobTriggerRequest
+            {
+                Name = hybridJobTrigger.Name
+            });
+
+            // Delete the activated job.
+            JobsDelete.DeleteJob(result.Name);
```
annadata, line 90: I think we should we invert the order. Delete the job and then delete the job trigger.

annadata (approved):
LGTM.

dlp/api/Snippets.Tests/SendDataToTheHybridJobTriggerTests.cs
jainamtrivedi-crest: Done

amanda-tarafa (approved):

PR #2216, opened 2023-05-28
chore: Using temporary databases for all sample tests that modify the database.

anuragsrivstv:
There changes are to make Spanner sample tests independent to each other. These includes :

- Changes to use temporary databases for all Google Standard SQL sample tests that modify the database.

- Changes to use temporary databases for all PostgreSQL sample tests that modify the database .


anuragsrivstv:
Hi @amanda-tarafa  Please let me know what do you think of this PR.  Does the changes seems fine to you ? Thanks!

amanda-tarafa (commented):
I've review the first few files, I've left comments there that will apply to some of the others.

I'm wondering if it makes sense to split this PR in 4 or 5, so that it's not more than 5 changes each.
Also, reviewing the fixture as it is is really hard, maybe check your eol configuration?

anuragsrivstv:
Will raise another PR instead of this.

PR #2218, opened 2023-05-30
feat: Add support for editorconfig

Rishabh-V:
To support consistent styling across the samples, there are two files in this PR:

1.  .editorconfig - This is very similar to the [.editorconfig used in google-cloud-dotnet](https://github.com/googleapis/google-cloud-dotnet/blob/main/.editorconfig). This one however has rules for public, internal and  private method names, properties etc and the list of common diagnostic listed at the bottom with the  severity set to suggestion. When all the current violations are fixed, we can change the severity of important diagnostics to warning/errors as needed.
2.  Directory.Build.Props - By adding a new property to this file placed in the root folder of the source, the property is added to every project in the subfolders. This can be used to enforce the code style in build using `<EnforceCodeStyleInBuild>` property. When this property is set to `true` and there are code violations for the .editorconfig diagnostic rules with the severity of error, the build will fail. When set to false, the violations will not fail the build.  The property can be overridden at the project/subfolder level. This is currently set to false. We can enable it when all the existing issues are fixed.

I am also sharing a document describing these. Thanks.

### DIFF
--- a/.editorconfig
+++ b/.editorconfig
@@ -1,0 +1,281 @@
+# To learn more about .editorconfig see https://aka.ms/editorconfigdocs
+# top-most EditorConfig file
+root = true
+###############################
+# Core EditorConfig Options   #
+###############################
+# All files
+[*]
+insert_final_newline = true
+indent_style = space
+indent_size = 4
+trim_trailing_whitespace = true
+
+# Generated code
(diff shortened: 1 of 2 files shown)


amanda-tarafa (approved):
Thanks, I'll take a look at the doc and we should start fixing what we find, until we can enforce style.

PR #2276, opened 2023-07-12
feat(functions): create new samples for typed function signature

garethgeorge:
Creates new samples for GCF nodejs strongly typed function signature. Samples added are

 * greeting - a typed hello world (accepting JSON)
 * googlechatbot - a typed function that matches the functionality documented in https://developers.google.com/chat/quickstart/gcf-app


functions/typed/GoogleChatBot/Function.cs
```diff
@@ -0,0 +1,119 @@
...
+/// for dependency injection, and adding middleware to the eventual application pipeline.
+/// In this case, we simply add an IHttpRequestReader and IHttpRequestWriter that will handle JSON encoded content for a typed function.
+/// </summary>
+public class Startup : FunctionsStartup
+{
+    // Provide implementations for IOperationSingleton, and IOperationScoped.
+    // The implementation is the same for both interfaces (the Operation class)
+    public override void ConfigureServices(WebHostBuilderContext context, IServiceCollection services) =>
+        services
+                .AddSingleton(typeof(IHttpRequestReader<JObject>), new CustomJsonReader<JObject>())
```
jskeet, line 45: Nit: unindent by four spaces.

functions/typed/GoogleChatBot/Function.cs
```diff
@@ -0,0 +1,119 @@
...
+
+/// <summary>
+/// The startup class can be used to perform additional configuration, including
+/// adding application configuration sources, reconfiguring logging, providing services
+/// for dependency injection, and adding middleware to the eventual application pipeline.
+/// In this case, we simply add an IHttpRequestReader and IHttpRequestWriter that will handle JSON encoded content for a typed function.
+/// </summary>
+public class Startup : FunctionsStartup
+{
+    // Provide implementations for IOperationSingleton, and IOperationScoped.
```
jskeet, line 41: This comment doesn't match the code.

functions/typed/GoogleChatBot/Function.cs
```diff
@@ -0,0 +1,119 @@
...
+using System.Threading;
+using System.Threading.Tasks;
+
+namespace GoogleChatBot;
+
+/// <summary>
+/// The startup class can be used to perform additional configuration, including
+/// adding application configuration sources, reconfiguring logging, providing services
+/// for dependency injection, and adding middleware to the eventual application pipeline.
+/// In this case, we simply add an IHttpRequestReader and IHttpRequestWriter that will handle JSON encoded content for a typed function.
```
jskeet, line 37: I'd wrap this comment line after "that will".

functions/typed/GoogleChatBot/Function.cs
```diff
@@ -0,0 +1,119 @@
...
+    // Provide implementations for IOperationSingleton, and IOperationScoped.
+    // The implementation is the same for both interfaces (the Operation class)
+    public override void ConfigureServices(WebHostBuilderContext context, IServiceCollection services) =>
+        services
+                .AddSingleton(typeof(IHttpRequestReader<JObject>), new CustomJsonReader<JObject>())
+                .AddSingleton(typeof(IHttpResponseWriter<JObject>), new CustomJsonWriter<JObject>());
+
+}
+
+/// <summary>
```
jskeet, line 50: I'd split this comment into multiple lines - possibly just before "Google Chat Bot"

functions/typed/GoogleChatBot/Function.cs
```diff
@@ -0,0 +1,119 @@
...
+/// </summary>
+[FunctionsStartup(typeof(Startup))]
+public class TypedFunction : ITypedFunction<JObject, JObject>
+{
+    public Task<JObject> HandleAsync(JObject request, CancellationToken cancellationToken)
+    {
+        string displayName = (string) request.SelectToken("$.message.sender.displayName");
+        string imageUrl = (string) request.SelectToken("$.message.sender.imageUrl");
+
+        return Task.FromResult(JObject.FromObject(new
```
jskeet, line 61: I'd write this as:

```csharp
var response = new
{
};
return Task.FromResult(JObject.FromObject(response));
```

functions/typed/GoogleChatBot/Function.cs
```diff
@@ -0,0 +1,119 @@
...
+                    card = new
+                    {
+                        name = "Avatar Card",
+                        header = new
+                        {
+                            title = string.Format("Hello {0}!", displayName),
+                        },
+                        sections = new[] {
+                            new {
+                                widgets = new dynamic [] {
```
jskeet, line 75: I'd suggest using `new object[]` rather than `new dynamic[]`

functions/typed/GoogleChatBot/Function.cs
```diff
@@ -0,0 +1,119 @@
...
+        {
+            cardsV2 = new[] {
+                new {
+                    cardId = "avatarCard",
+                    card = new
+                    {
+                        name = "Avatar Card",
+                        header = new
+                        {
+                            title = string.Format("Hello {0}!", displayName),
```
jskeet, line 71: I'd write:

```csharp
header = new { title = $"Hello {displayName}!" }
```

Keeping single-property objects on a single line reduces the cruft. Same for `textParagraph` and `image` later.

functions/typed/Greeting/Function.cs
```diff
@@ -0,0 +1,48 @@
...
+using System.Threading.Tasks;
+
+namespace Greeting;
+
+public class TypedFunction : ITypedFunction<GreetingRequest, GreetingResponse>
+{
+    public Task<GreetingResponse> HandleAsync(GreetingRequest request, CancellationToken cancellationToken) =>
+        Task.FromResult(new GreetingResponse
+        {
+            Message = string.Format("Hello {0} {1}!", request.FirstName, request.LastName)
```
jskeet, line 29: This can be simplified to

```csharp
Task.FromResult(new GreetingResponse { Message = $"Hello {request.FirstName, request.LastName}" });
```

functions/typed/GoogleChatBot/Function.cs
```diff
@@ -0,0 +1,119 @@
...
+// [START functions_typed_googlechatbot]
+using Google.Cloud.Functions.Framework;
+using Google.Cloud.Functions.Hosting;
+using Microsoft.AspNetCore.Hosting;
+using Microsoft.AspNetCore.Http;
+using Microsoft.Extensions.DependencyInjection;
+using Newtonsoft.Json;
+using Newtonsoft.Json.Linq;
+using System;
+using System.Dynamic;
```
jskeet, line 24: I doubt that we need System.Dynamic here - I suspect quite a few of these aren't required, in fact. (If you don't have an IDE that will show you which are actually needed, I'm happy to pull the code and see what Visual Studio suggests.)

functions/typed/Greeting/Function.cs
```diff
@@ -0,0 +1,48 @@
...
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+// [START functions_typed_greeting]
+using Google.Cloud.Functions.Framework;
+using System.Dynamic;
```
jskeet, line 17: Again, I don't think System.Dynamic should be needed.

functions/typed/Greeting/Function.cs
```diff
@@ -0,0 +1,48 @@
...
+    public string FirstName { get; set; }
+
+    [JsonPropertyName("last_name")]
+    public string LastName { get; set; }
+}
+
+public class GreetingResponse
+{
+    [JsonPropertyName("message")]
+    public dynamic Message { get; set; }
```
jskeet, line 45: I'd change this to `string`. There's no need for dynamic typing here.

functions/typed/Typed.Tests/TypedGoogleChatBotTest.cs
```diff
@@ -0,0 +1,88 @@
...
+                    ""imageUrl"": ""example.com/avatar.png""
+                }
+            }
+        }", Encoding.UTF8);
+
+        await ExecuteHttpRequestAsync(request, async response =>
+        {
+            Assert.Equal(HttpStatusCode.OK, response.StatusCode);
+            JToken actual = JToken.Parse(await response.Content.ReadAsStringAsync());
+            Assert.True(JToken.DeepEquals(actual, expected), string.Format("JSON response {0} did not match expected {1}", actual, expected));
```
jskeet, line 85: Again, use an interpolated string literal instead of string.Format.

jskeet (commented):
Mostly nits - but it does strike me as unfortunate that the chat bot sample requires such deeply nested JSON. It means that the vast majority of the sample code isn't actually useful for learning about typed functions - it's just cruft. Presumably this is the same for other languages - I'd urge you to consider changing the sample to one which has a higher signal to noise ratio.

garethgeorge:
> Mostly nits - but it does strike me as unfortunate that the chat bot sample requires such deeply nested JSON. It means that the vast majority of the sample code isn't actually useful for learning about typed functions - it's just cruft. Presumably this is the same for other languages - I'd urge you to consider changing the sample to one which has a higher signal to noise ratio.

Agree re: the google chat bot sample but we did want to provide something that would be a good example of working with a real world JSON API that's easy to get started with (google chat API takes very little setup) and it's perhaps useful to illustrate how customers can deal with unstructured JSON with more reusable boilerplate than they have today. I think we'll be able to include more useful functional examples when we publish type libraries generated from protobufs (which are the real intent of the feature). In the mean time I was hoping this would be a useful place to show how alternative JSON libraries i.e. Newtonsoft can be used with the .NET implementation.

I think our intent is to highlight the greeting example as a starting point and to include the googlechatbot sample in a separate guide that goes a bit deeper.

WDYT?

jskeet:
I'm on vacation until next Friday - I'll have a look then.

jskeet:
Personally I'd split this into three samples instead of two:

- The greeting sample as a very simple one
- A ChatBot sample using classes (i.e. an extended version of the greeting one)
- A separate but *simple* one showing the use of Newtonsoft.Json - possibly even make it just a dynamic version of the greeting sample (i.e. sample JSON, but without the classes)

jskeet:
Yes, I think it would be worth dropping the Newtonsoft.Json aspect from these samples.

I would suggest adding an example in https://github.com/GoogleCloudPlatform/functions-framework-dotnet/tree/main/examples and documenting it in https://github.com/GoogleCloudPlatform/functions-framework-dotnet/blob/main/docs/examples.md

garethgeorge:
Hi Jon, after starting on writing up a version that explicitly creates classes for the full structure of the ChatApp request I think you're right that the nesting structure is too deep to make a good sample for the statically typed languages -- the number of auxiliary types needed make for a cluttered sample.

I'm going to go ahead and drop the chatapp sample for dotnet and will think about whether we can publish an advanced example that uses types published in a type registry when we schedule that work. I'll send a separate PR to add the newtonsoft sample to the github.com/GoogleCloudPltaform/functions-framework-dotnet repo.

functions/typed/GoogleChatBot/Function.cs
garethgeorge: Good catch, thanks! Got rid of System.Dynamic and a few others.

garethgeorge:
Went ahead and applied the remaining code review feedback on Greeting, thanks!

garethgeorge:
Sent https://github.com/GoogleCloudPlatform/functions-framework-dotnet/pull/385 adding the more advanced example to the functions framework repo.

jskeet:
I'll look on Monday - I'm away until then.

functions/typed/typed.sln
```diff
@@ -0,0 +1,26 @@
+﻿
+Microsoft Visual Studio Solution File, Format Version 12.00
+# Visual Studio Version 17
+VisualStudioVersion = 17.0.31903.59
+MinimumVisualStudioVersion = 10.0.40219.1
+Project("{FAE04EC0-301F-11D3-BF4B-00C04F79EFBC}") = "Greeting", "Greeting\Greeting.csproj", "{5D604C14-3A3E-48F7-83A4-20EE191F0707}"
```
jskeet, line 6: This should include the test project as well.

functions/typed/Typed.Tests/Typed.Tests.csproj
```diff
@@ -0,0 +1,22 @@
+<Project Sdk="Microsoft.NET.Sdk">
+
+  <PropertyGroup>
+    <TargetFramework>net6.0</TargetFramework>
+
```
jskeet, line 5: Nit: remove blank line.

functions/typed/Typed.Tests/TypedGreetingTest.cs
```diff
@@ -0,0 +1,48 @@
...
+{
+    [Fact]
+    public async Task GreetingTest()
+    {
+        var request = new HttpRequestMessage(HttpMethod.Post, "example.com");
+        request.Content = new StringContent(@"{
+            ""first_name"": ""Jane"",
+            ""last_name"": ""Doe""
+        }
+        ", Encoding.UTF8);
```
jskeet, line 37: Unwrap to the previous line?

functions/typed/Typed.Tests/TypedGreetingTest.cs
```diff
@@ -0,0 +1,48 @@
...
+
+        await ExecuteHttpRequestAsync(request, async response =>
+        {
+            JToken actual = JToken.Parse(await response.Content.ReadAsStringAsync());
+            JToken expected = JToken.Parse(@"{
```
jskeet, line 42: I'd probably put all of this on one line, personally.

functions/typed/Typed.Tests/TypedGreetingTest.cs
```diff
@@ -0,0 +1,48 @@
...
+                ""message"": ""Hello Jane Doe!""
+            }");
+            Assert.True(JToken.DeepEquals(actual, expected), string.Format("JSON response {0} did not match expected {1}", actual, expected));
```
jskeet, line 45: I suggest using interpolated strings:

```csharp
Assert.True(JToken.DeepEquals(actual, expected), $"JSON response {actual} did not match expected {expected}");
```

functions/typed/Typed.Tests/TypedGreetingTest.cs
garethgeorge: Admittedly the JSON comes in under the line length limit in either style I feel that the line breaks generally keep the JSON a bit more readable.

functions/typed/Typed.Tests/TypedGreetingTest.cs
garethgeorge: Thanks! I tend to forget that .NET supports these. Done.

functions/typed/Greeting/Function.cs
```diff
@@ -0,0 +1,47 @@
...
+
+// [START functions_typed_greeting]
+using Google.Cloud.Functions.Framework;
+using System.Text.Json.Serialization;
+using System.Threading;
+using System.Threading.Tasks;
+
+namespace Greeting;
+
+public class TypedFunction : ITypedFunction<GreetingRequest, GreetingResponse>
```
jskeet, line 23: I'm not sure whether this has changed or I just didn't spot it before. Our templates all create functions just called Function, and that's what all the other samples use. If you really, really want to use the class name TypedFunction, please rename the file as well - but I think it would be better to reset the name to Function.

functions/typed/Typed.Tests/TypedGreetingTest.cs
```diff
@@ -0,0 +1,47 @@
...
+using System.Threading.Tasks;
+using Newtonsoft.Json;
+using Newtonsoft.Json.Linq;
+using Xunit;
+using Greeting;
+using System.Text;
+
+namespace Typed.Tests;
+
+public class TypedGreetingTest : FunctionTestBase<TypedFunction>
```
jskeet, line 27: I'd rename this to GreetingTest - renaming the file too - otherwise it sounds like there's a class or namespace called "TypedGreeting".

(I think the namespace and project name is fine though.)

jskeet (approved):
Tiny tweaks - I'm approving so that you can merge and then I can do the renames if you'd prefer. Let me know.

garethgeorge:
Thanks Jon, addressed final feedback.

jskeet (approved):
Thanks, this looks good to me now.

@amanda-tarafa Anything else we need to consider before merging? (Would you like to merge, in terms of doing the right thing with squashing etc?)

functions/typed/Typed.Tests/GreetingTest.cs
```diff
@@ -0,0 +1,47 @@
+// Copyright 2020 Google LLC
```
amanda-tarafa, line 1:
```suggestion
// Copyright 2023 Google LLC
```

amanda-tarafa (approved):
Tiny nit but otherwise it looks good.

PR #2301, opened 2023-08-08
chore: Update gcs samples

hemanshv:
updated GenerateV4SignedReadUrlSample and GenerateV4UploadSignedUrlSample to use UrlSigner.FromCredentialFile(credentialFilePath) instead of the deprecated UrlSigner.FromServiceAccountPath(credentialFilePath).


hemanshv:
Hi @amanda-tarafa ,
Please share your feedback on the same.
This PR is to resolve b/281105553.

Thanks,

amanda-tarafa (changes requested):
I think there's a second point the user is making that we should consider, and that is to use Application Default Credentials. This should work, but you'll need to modify the tests as well.

When these samples were first written, only service account credentials were supported, but we now support other types of credentials so it makes more sense to use ADC.

amanda-tarafa (approved):

PR #2316, opened 2023-08-29
DLP: Added sample for calculating k-anonymity on dataset

jainamtrivedi-crest:
- Added sample for [Calculating k-anonymity on the dataset](https://cloud.google.com/dlp/docs/visualizing_re-id_risk.md?g=0&l=119#step_1_calculate_k-anonymity_on_the_dataset) and test cases

### DIFF
--- a/dlp/api/Snippets/CalculateKAnonymityOnDataset.cs
+++ b/dlp/api/Snippets/CalculateKAnonymityOnDataset.cs
@@ -1,0 +1,100 @@
+﻿// Copyright 2023 Google Inc.
+//
+// Licensed under the Apache License, Version 2.0 (the "License"); you may not
+// use this file except in compliance with the License. You may obtain a copy of
+// the License at
+//
+// http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS, WITHOUT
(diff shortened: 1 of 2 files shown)


jakubrauch (approved):

dlp/api/Snippets/CalculateKAnonymityOnDataset.cs
```diff
@@ -0,0 +1,104 @@
...
+        var request = new CreateDlpJobRequest
+        {
+            ParentAsLocationName = new LocationName(projectId, "global"),
+            RiskJob = riskJob
+        };
+
+        // Send the job request.
+        DlpJob response = dlp.CreateDlpJob(request);
+
+        return response;
```
annadata, line 94: The [java snippet](https://github.com/GoogleCloudPlatform/java-docs-samples/blob/fafc6dca9dad437b9de9062d8f54168396fb0109/dlp/snippets/src/main/java/dlp/snippets/RiskAnalysisKAnonymityWithEntityId.java#L155) unpacks and prints the results. This can be helpful. 
Any constraints on including that?

jainamtrivedi-crest: Initially we were waiting for the job to get completed but as per Amanda's suggestion we decided to remove the waiting time in the sample because job might take more amount of time to get completed which leads to timeout in test suite. That is why we have not waited for the job to get completed in other C# samples as well.

amanda-tarafa (approved):

PR #2329, opened 2023-09-12
fix: refactor error reporting code sample

minherz:
align code sample for error reporting with code in other languages (b/275730159):
1. use an exception with "fixed" message
1. report error context
1. follow best practices for the language to report service context

### DIFF
--- a/error-reporting/api/ErrorReportingSample/ErrorReportingSample.cs
+++ b/error-reporting/api/ErrorReportingSample/ErrorReportingSample.cs
@@ -57,32 +57,47 @@ public class ErrorReportingSample
 
         // Add a service context to the report. For more details see:
         // https://cloud.google.com/error-reporting/reference/rest/v1beta1/ServiceContext
-        var assemblyName = System.Reflection.Assembly.GetExecutingAssembly().GetName();
+        var assemblyName = System.Reflection.Assembly.GetEntryAssembly().GetName();
         var serviceContext = new ServiceContext()
         {
             Service = assemblyName.Name,
             Version = assemblyName.Version.ToString(),
         };
-        // Add an error context to the report. For more details see:
-        // https://cloud.google.com/error-reporting/reference/rest/v1beta1/ErrorContext
-        var stackTrace = new System.Diagnostics.StackTrace(e, true);
-        var stackFrame = stackTrace.GetFrame(stackTrace.FrameCount - 1);
-        var errorContext = new ErrorContext()
-        {
-            ReportLocation = new SourceLocation()
-            {
-                FilePath = stackFrame.GetFileName(),
-                LineNumber = stackFrame.GetFileLineNumber(),
-                FunctionName = stackFrame.GetMethod().Name,
-            },
-        };
         var errorEvent = new ReportedErrorEvent()
         {
             Message = e.ToString(),
             ServiceContext = serviceContext,
-            Context = errorContext,
+            Context = getErrorContext(e),
         };
         reporter.ReportErrorEvent(projectName, errorEvent);
+    }
+
+    /// <summary>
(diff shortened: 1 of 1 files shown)


error-reporting/api/ErrorReportingSample/ErrorReportingSample.cs
```diff
@@ -56,16 +56,31 @@ private static void report(Exception e)
...
+        var stackFrame = stackTrace.GetFrame(stackTrace.FrameCount - 1);
+        var errorContext = new ErrorContext()
+        {
+            ReportLocation = new SourceLocation()
+            {
+                FilePath = stackFrame.GetFileName(),
+                LineNumber = stackFrame.GetFileLineNumber(),
+                FunctionName = stackFrame.GetMethod().Name,
+            },
+        };
```
amanda-tarafa, line 78: A couple of comments here:

This information is not always available as it depends on what debug information you have available (.pdb files). So you need to protect against it not being present.

It's better to use the bottom frame as that's the frame where the exception was actually thrown, the rest show how the exception was bubbled up.

See what we do in [Google.Cloud.Diagnostics](https://github.com/googleapis/google-cloud-dotnet/blob/8509295870d00b72bc70b8d3a63a10bcacd8dafd/apis/Google.Cloud.Diagnostics.Common/Google.Cloud.Diagnostics.Common/ErrorReporting/ErrorReportingContextExceptionLogger.cs#L127-L143) for reference.

error-reporting/api/ErrorReportingSample/ErrorReportingSample.cs
```diff
@@ -56,16 +56,31 @@ private static void report(Exception e)
             Environment.GetEnvironmentVariable("GOOGLE_PROJECT_ID"));
 
         // Add a service context to the report. For more details see:
-        // https://cloud.google.com/error-reporting/reference/rest/v1beta1/projects.events#ServiceContext
-        ServiceContext serviceContext = new ServiceContext()
+        // https://cloud.google.com/error-reporting/reference/rest/v1beta1/ServiceContext
+        var assemblyName = System.Reflection.Assembly.GetExecutingAssembly().GetName();
+        var serviceContext = new ServiceContext()
```
amanda-tarafa, line 61: I'm not certain you'd want this to be used as the service context. This is the assembly that contains the code currently executing, for instance, if your organization has a library for logging, then this will be that library's assembly.

You might want to look at [GetEntryAssembly](https://learn.microsoft.com/en-us/dotnet/api/system.reflection.assembly.getentryassembly?view=net-7.0) instead?

@jskeet In case you have other ideas.

error-reporting/api/ErrorReportingSample/ErrorReportingSample.cs
minherz: The code is refactored to reflect on the best practices.

error-reporting/api/ErrorReportingSample/ErrorReportingSample.cs
minherz: Replaced. :+1:

amanda-tarafa (approved):
LGTM.

Let me know if this is ready for merging and I'll do so.

minherz:
@amanda-tarafa it looks like it is gated by the @jskeet review.

amanda-tarafa:
Nope, we require 2 reviews in the repo, one from language (me or Jon) and one from product. For the products we don't have reviewers, and since the 2 reviews requirement cannot be apply to specific folders just to the entire repo, the we fallback to both Jon and myself but only one review is required.

You can update the OWNERS file for Error Reporting adding a team that could do product side reviews.

I'll merge now, thanks!!

PR #2336, opened 2023-09-15
chore: Move idiomatic reviewers to be from dotnet-samples-reviewers

amanda-tarafa:


amanda-tarafa:
Self merging, this is trivial and only affects process.

PR #2337, opened 2023-09-15
chore: refactor error reporting sample

minherz:
this is a refactoring of the change made in #2329 .
it removes the explicit source location as [recommended] by Error Reporting API. align the method name to the language guidelines.

[recommended]: https://cloud.google.com/error-reporting/reference/rest/v1beta1/projects.events/report#ReportedErrorEvent.FIELDS.message

### DIFF
--- a/error-reporting/api/ErrorReportingSample/ErrorReportingSample.cs
+++ b/error-reporting/api/ErrorReportingSample/ErrorReportingSample.cs
@@ -49,7 +49,7 @@ public class ErrorReportingSample
         }
         catch (Exception e)
         {
-            report(e);
+            ReportError(e);
             Console.WriteLine("Stackdriver Error Report Sent");
         }
     }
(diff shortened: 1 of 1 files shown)


error-reporting/api/ErrorReportingSample/ErrorReportingSample.cs
```diff
@@ -76,37 +76,8 @@ private static void report(Exception e)
         {
             Message = e.ToString(),
             ServiceContext = serviceContext,
-            Context = getErrorContext(e),
```
NimJay, line 79: **Question:** Why do we no longer need to explicitly specify the error context?

NimJay (approved):
Approved!
Left a question, but no a blocker.

Thanks, @minherz, for working on these adjustments.

error-reporting/api/ErrorReportingSample/ErrorReportingSample.cs
NimJay: Ah, you answered my question in https://github.com/GoogleCloudPlatform/java-docs-samples/pull/8648:

> "delegate error context processing to Error Reporting backend"

Sounds good!

amanda-tarafa (approved):

PR #2368, opened 2023-10-30
docs(aiplatform): Added Predict Text Prompt GenAI code sample

holtskinner:
- Based on this Java sample: https://github.com/GoogleCloudPlatform/java-docs-samples/blob/HEAD/aiplatform/src/main/java/aiplatform/PredictTextPromptSample.java
- Belongs in this spot in the docs: https://cloud.google.com/vertex-ai/docs/generative-ai/text/test-text-prompts#test_text_prompts


amanda-tarafa:
I've requested a review from @meteatamel as he was working on some related samples as well.

amanda-tarafa (approved):

meteatamel:
@holtskinner @amanda-tarafa I know this is already merged but one lingering question I have is: Is it better to use protobuf constructs or more familiar JSON in these samples? For example, this is what we currently have:

```
        var parameters = new wkt::Value
        {
            StructValue = new wkt::Struct
            {
                Fields =
                {
                    { "temperature", new wkt::Value { NumberValue = 0.2 } },
                    { "maxOutputTokens", new wkt::Value { NumberValue = 256 } },
                    { "topP", new wkt::Value { NumberValue = 0.95 } },
                    { "topK", new wkt::Value { NumberValue = 40 } }
                }
            }
        };
```

Which is fine but it's quite unintuitive to me as a C# dev. An alternative way would be like this:

```
        var parameters = new
        {
            temperature = 0.2,
            maxOutputTokens = 256,
            topK = 40,
            topP = 0.95
        };

// later convert to Protobuf to use in the request
Parameters = wkt::Value.Parser.ParseJson(JsonConvert.SerializeObject(parameters)),
```

The latter is definitely more intuitive to me, it's mainly JSON and only convert to Protobuf when I'm sending it as show in my sample: https://github.com/meteatamel/genai-samples/blob/main/vertexai/palm2-text/console/csharp/sdk/Program.cs

jskeet:
While parsing a `Value` from arbitrary JSON *should* be okay, it still makes me nervous... because it might mean users expect to be able to use "regular" JSON serialization for other messages.

Eventually I hope we'll have genuine protobuf messages for this so it's type safe anyway - until then, I'd personally prefer us to stick with the explicit initialization.

Issue #2382, opened 2023-11-03
docs(aiplatform): Add PredictChatPromptSample


meteatamel:
Based on this Java sample: https://github.com/GoogleCloudPlatform/java-docs-samples/blob/HEAD/aiplatform/src/main/java/aiplatform/PredictChatPromptSample.java
Belongs in this spot in the docs: https://cloud.google.com/vertex-ai/docs/generative-ai/chat/test-chat-prompts

PR #2398, opened 2023-11-10
docs(aiplatform): Add PredictChatPromptSample #2382

meteatamel:


aiplatform/api/AIPlatform.Samples/PredictChatPromptSample.cs
```diff
@@ -0,0 +1,151 @@
...
+using System;
+using System.Collections.Generic;
+
+public class PredictChatPromptSample
+{
+    public string PredictChatPrompt(
+        string projectId = "your-project-id",
+        string locationId = "us-central1",
+        string publisher = "google",
+        string model = "chat-bison@001"
```
Rishabh-V, line 31: Suggestion: Just thinking as an end user, I will very much appreciate if I know all the different models that I can use, so a link  to documentation pointing to different models available will be helpful.

aiplatform/api/AIPlatform.Samples/PredictChatPromptSample.cs
```diff
@@ -0,0 +1,151 @@
...
+                {
+                    author = "user",
+                    content = prompt
+                }
+            }
+        });
+        // var instance = wkt::Value.Parser.ParseJson(instanceJson);
+
+        // Or, you can construct Protobuf directly.
+        var instance = new wkt::Value
```
Rishabh-V, line 72: My preference would be to show just one approach here to construct Protobuf directly. 
We can add a TODO stating, this would be greatly simplified when the new version of library is available and link to the documentation showing a possible second approach using JSON.

aiplatform/api/AIPlatform.Samples/PredictChatPromptSample.cs
```diff
@@ -0,0 +1,151 @@
...
+        {
+            StructValue = new()
+            {
+                Fields =
+                {
+                    ["context"] = wkt::Value.ForString("My name is Miles. You are an astronomer, knowledgeable about the solar system."),
```
Rishabh-V, line 78: We can avoid repeating `wkt::Value.` multiple times by adding a static using directive 

```
``` 

aiplatform/api/AIPlatform.Samples/PredictChatPromptSample.cs
```diff
@@ -0,0 +1,151 @@
...
+
+        // You can construct Protobuf from JSON.
+        var parametersJson = JsonConvert.SerializeObject(new
+        {
+            temperature = 0.3,
+            maxDecodeSteps = 200,
+            topP = 0.8,
+            topK = 40
+        });
+        // var parameters = wkt::Value.Parser.ParseJson(parametersJson);
```
Rishabh-V, line 124: Same for parameters - Only one way to construct Protobuf.

aiplatform/api/AIPlatform.Samples/PredictChatPromptSample.cs
```diff
@@ -0,0 +1,151 @@
...
+                    { "topK", new wkt::Value { NumberValue = 40 } }
+                }
+            }
+        };
+
+        // Make the request.
+        var response = client.Predict(endpoint, instances, parameters);
+
+        // Parse the response and return the content.
+        var content = response.Predictions[0].StructValue.Fields["candidates"].ListValue.Values[0].StructValue.Fields["content"].StringValue;
```
Rishabh-V, line 145: This should be simplified after the library is generated from new protos. May be a TODO here as well. So, that we don't miss out on it later.

Rishabh-V (commented):

aiplatform/api/AIPlatform.Samples/PredictChatPromptSample.cs
meteatamel: I mentioned already in other PR but these samples will be embedded within the docs pages, so given that, I'm not sure if we want to explain more or link to docs.

aiplatform/api/AIPlatform.Samples/PredictChatPromptSample.cs
meteatamel: Ack. I still think showing JSON in one of the samples would be useful, since Protobuf is so hard to work with. However, I'll let Amanda make the final call.

aiplatform/api/AIPlatform.Samples/PredictChatPromptSample.cs
meteatamel: I'd prefer Value.ForString (after using alias) but I'll follow whatever the team convention is after Amanda chimes in.

aiplatform/api/AIPlatform.Samples/PredictChatPromptSample.cs
Rishabh-V: Right Mete. I had prefixed it as a suggestion. If the sample is tied to the docs already, it is fine. Thanks

aiplatform/api/AIPlatform.Samples/PredictChatPromptSample.cs
Rishabh-V: Right. Given we have another type named Value, alias would be better.

aiplatform/api/AIPlatform.Samples/PredictChatPromptSample.cs
amanda-tarafa: @meteatamel Can you pick one sample to show JSON, have the rest with protobuf, and then explain in docs something like "This can also be done using JSON, please refer to sample xyz who demonstrates this"?

I find that the samples showing both are very long and for an unfamiliar user could end up being confusing.

I wouldn't mention a new version of the library as we don't have any ETA on that.

aiplatform/api/AIPlatform.Samples/PredictChatPromptSample.cs
amanda-tarafa: This is sample code, that users will see, we shouldn't leave TODOs for ourselves. Only TODOs, if any, might be those directed at developers  consuming the samples like this hypothetical example: "TODO: change this location to the location where you are storing your data.". And not even those are usual.

Also, we shouldn't comment on possible changes for which we don't have ETAs, etc.

aiplatform/api/AIPlatform.Samples/PredictChatPromptSample.cs
amanda-tarafa: Yes, I also prefer Value.ForString with alias. For people that are not familiar with Protobuf just ForString might not be self evident.

amanda-tarafa (commented):
@meteatamel looks good, except for the "use one single approach proto vs. JSON per sample suggestions". I'll wait for those changes (or arguments against) before LGTMing.

I'll also review the other samples, this suggestion applies to those as well so I won't repeat it on each. I'll just comment on particulars of each sample, if any.

aiplatform/api/AIPlatform.Samples/PredictChatPromptSample.cs
meteatamel: Fixed

aiplatform/api/AIPlatform.Samples/PredictChatPromptSample.cs
meteatamel: Ok, I chose the JSON one in this sample and Protobuf in the other samples, so we can show both

PR #2399, opened 2023-11-10
docs(aiplatform): Add PredictTextExtractionSample #2388

meteatamel:


aiplatform/api/AIPlatform.Samples/PredictTextExtractionSample.cs
```diff
@@ -0,0 +1,115 @@
...
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+// [START aiplatform_sdk_extraction]
+
+using Google.Cloud.AIPlatform.V1;
+using wkt = Google.Protobuf.WellKnownTypes;
```
anuragsrivstv, line 20: I have same comments on this sample as other samples. Please address them here as well if you change anything in other samples.

aiplatform/api/AIPlatform.Samples/PredictTextExtractionSample.cs
```diff
@@ -0,0 +1,115 @@
...
+        {
+            Endpoint = $"{locationId}-aiplatform.googleapis.com"
+        }.Build();
+
+        // Configure the parent resource.
+        var endpoint = EndpointName.FromProjectLocationPublisherModel(projectId, locationId, publisher, model);
+
+        // Initialize request argument(s).
+        var content = @"
+Background: There is evidence that there have been significant changes
```
amanda-tarafa, line 46: Is this sample different from the one that's supposed to be summarizing text?

If not, then just keep one and you can add more than one region tag to a single sample.

Also, hoping that the different behaviour of the text-bison model with the different input is explained somewhere.

aiplatform/api/AIPlatform.Samples/PredictTextExtractionSample.cs
meteatamel: Fixed

aiplatform/api/AIPlatform.Samples/PredictTextExtractionSample.cs
meteatamel: Yes, it's slightly different. I'm merely mirroring Java and other samples, so they have this sample as a separate thing, and I think we should do the same in C# for consistency in docs.

amanda-tarafa (approved):

PR #2400, opened 2023-11-10
docs(aiplatform): Add PredictCodeChatSample #2389

meteatamel:


aiplatform/api/AIPlatform.Samples/PredictCodeChatSample.cs
```diff
@@ -0,0 +1,106 @@
...
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+// [START aiplatform_sdk_code_chat]
+
+using Google.Cloud.AIPlatform.V1;
+using wkt = Google.Protobuf.WellKnownTypes;
```
Rishabh-V, line 20: Nit: Sort using directives alphabetically.  (wkt should be the last one)

aiplatform/api/AIPlatform.Samples/PredictCodeChatSample.cs
```diff
@@ -0,0 +1,106 @@
...
+using System;
+using System.Collections.Generic;
+
+public class PredictCodeChatSample
+{
+    public string PredictCodeChat(
+        string projectId = "your-project-id",
+        string locationId = "us-central1",
+        string publisher = "google",
+        string model = "codechat-bison@001"
```
Rishabh-V, line 30: Suggestion: Just thinking as an end user, I will very much appreciate if I know all the different models that I can use, so a link  to documentation pointing to different models available will be helpful.

aiplatform/api/AIPlatform.Samples/PredictCodeChatSample.cs
```diff
@@ -0,0 +1,106 @@
...
+        // Configure the parent resource.
+        var endpoint = EndpointName.FromProjectLocationPublisherModel(projectId, locationId, publisher, model);
+
+        var instance = new wkt::Value
+        {
+            StructValue = new()
+            {
+                Fields =
+                {
+                    ["messages"] = wkt::Value.ForList(
```
Rishabh-V, line 50: We can avoid repeating `wkt::Value.` multiple times by adding a static using directive 

```
``` 

aiplatform/api/AIPlatform.Samples/PredictCodeChatSample.cs
```diff
@@ -0,0 +1,106 @@
...
+                    { "maxOutputTokens", new wkt::Value { NumberValue = 1024 } }
+                }
+            }
+        };
+
+        // Make the request.
+        var response = client.Predict(endpoint, instances, parameters);
+
+        // Parse and return the content.
+        var content = response.Predictions[0].StructValue.Fields["candidates"].ListValue.Values[0].StructValue.Fields["content"].StringValue;
```
Rishabh-V, line 100: This should be simplified after the library is generated from new protos. May be a TODO here as well. So, that we don't miss out on it later.

Rishabh-V (commented):

aiplatform/api/AIPlatform.Samples/PredictCodeChatSample.cs
meteatamel: These samples will be part of specific doc pages where the model is basically set. As such, I'm not sure if we want to list all different models in the code as comments.

aiplatform/api/AIPlatform.Samples/PredictCodeChatSample.cs
meteatamel: As I mentioned in the other PR, I'm not sure if this works, because there's another Value in the AI library.

Rishabh-V: We can use [alias](https://learn.microsoft.com/en-us/dotnet/csharp/language-reference/keywords/using-directive#using-alias) to disambiguate the Value if there is conflict in Value.

However, here I had made a suggestion to use [static using](https://learn.microsoft.com/en-us/dotnet/csharp/language-reference/keywords/using-directive#static-modifier) because using static directive, wkt::Value.ForString etc which are repeated number of times can be replaced with just `ForString` directly. I think using alias would be neat (and it applies to all other PRs as well)

meteatamel: I see. If we use alias, then I kind of prefer Value.ForString instead of ForString but I'll let Amanda chime in.

aiplatform/api/AIPlatform.Samples/PredictCodeChatSample.cs
meteatamel: I'll do another pass on all of these samples once the new library is out..so it's not really just here but the whole sample will be refactored.

aiplatform/api/AIPlatform.Samples/PredictCodeChatSample.cs
Rishabh-V: Ack.

aiplatform/api/AIPlatform.Samples/PredictCodeChatSample.cs
amanda-tarafa: Yep, Value as an alias seems best. Just ForString might confuse some users.

aiplatform/api/AIPlatform.Samples/PredictCodeChatSample.cs
```diff
@@ -0,0 +1,106 @@
...
+                                ["author"] = wkt::Value.ForString("system"),
+                                ["content"] = wkt::Value.ForString("I am doing good. What can I help you in the coding world?"),
+                            }
+                        }),
+                        wkt::Value.ForStruct(new()
+                        {
+                            Fields =
+                            {
+                                ["author"] = wkt::Value.ForString("user"),
+                                ["content"] = wkt::Value.ForString("Please help write a function to calculate the min of two numbers."),
```
amanda-tarafa, line 72: I'm really hoping the result of this prompt is correct, right? What language is the function in, and can we make that be C# for our samples?

aiplatform/api/AIPlatform.Samples/PredictCodeChatSample.cs
meteatamel: Fixed

aiplatform/api/AIPlatform.Samples/PredictCodeChatSample.cs
meteatamel: I actually changed the prompt to say "C# function" and that seems to generate a function correctly

amanda-tarafa (approved):

PR #2401, opened 2023-11-10
docs(aiplatform): Add PredictCodeGenerationUnitTestSample #2390

meteatamel:


aiplatform/api/AIPlatform.Samples/PredictCodeGenerationUnitTestSample.cs
```diff
@@ -0,0 +1,92 @@
...
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+// [START aiplatform_sdk_code_generation_unittest]
+
+using Google.Cloud.AIPlatform.V1;
+using wkt = Google.Protobuf.WellKnownTypes;
```
Rishabh-V, line 20: Nit: Sort using directives alphabetically.  (wkt should be the last one)

aiplatform/api/AIPlatform.Samples/PredictCodeGenerationUnitTestSample.cs
```diff
@@ -0,0 +1,92 @@
...
+            Endpoint = $"{locationId}-aiplatform.googleapis.com"
+        }.Build();
+
+        // Configure the parent resource.
+        var endpoint = EndpointName.FromProjectLocationPublisherModel(projectId, locationId, publisher, model);
+
+        // Learn how to create prompts to work with a code model to generate code:
+        // https://cloud.google.com/vertex-ai/docs/generative-ai/code/code-generation-prompts
+        var prefix = @"
+Write a unit test for this function:
```
Rishabh-V, line 46: Question: Does this work for C# code as well. I understand that we may want to show the same code across language samples but can this code be in C#?

aiplatform/api/AIPlatform.Samples/PredictCodeGenerationUnitTestSample.cs
```diff
@@ -0,0 +1,92 @@
...
+                else:
+                    return False
+            else:
+                return True
+        else:
+            return False";
+
+        var instances = new List<wkt::Value>
+        {
+            wkt::Value.ForStruct(new()
```
Rishabh-V, line 61: We can avoid repeating `wkt::Value.` multiple times by adding a static using directive 

```
``` 

aiplatform/api/AIPlatform.Samples.Tests/PredictCodeGenerationUnitTestTest.cs
```diff
@@ -0,0 +1,37 @@
...
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+using Xunit;
+
+[Collection(nameof(AIPlatformFixture))]
+public class PredictCodeGenerationUnitTestTest
```
Rishabh-V, line 20: This name sounds a bit odd to me, ending in TestTest

Rishabh-V (commented):

aiplatform/api/AIPlatform.Samples/PredictCodeGenerationUnitTestSample.cs
meteatamel: But it's referring to `Google.Protobuf.WellknownTypes` so that's why I have it here :-)

aiplatform/api/AIPlatform.Samples/PredictCodeGenerationUnitTestSample.cs
meteatamel: I'm not sure if that works because there's another `Value` in the `AIPlatform.V1` library. Someone else suggested using `using Value = Google.Protobuf.WellKnownTypes.Value;` and that works. I'm happy to use whatever the team decides

aiplatform/api/AIPlatform.Samples.Tests/PredictCodeGenerationUnitTestTest.cs
meteatamel: Yes it's indeed weird but it's really a test of a unit test generation class. So, in that sense, TestTest makes sense.

Rishabh-V: I am okay if that name is fine with all. I was thinking if we name something like <PredictUnitTestCodeGeneration>Test then we can avoid this unfortunate name.  

That said, this is a change in naming convention that is used for other samples which roughly follows the region tag.

aiplatform/api/AIPlatform.Samples/PredictCodeGenerationUnitTestSample.cs
Rishabh-V: Yes, but still it should be sorted alphabetically.

aiplatform/api/AIPlatform.Samples/PredictCodeGenerationUnitTestSample.cs
meteatamel: I think it can be C#, yes. I'm simply following the same code for other languages (Java, Node.js, etc.) in the docs, so that's why it's on Python.

aiplatform/api/AIPlatform.Samples/PredictCodeGenerationUnitTestSample.cs
meteatamel: Ok, are you suggesting we use the *alias* or *static using* then?

Rishabh-V: Given that we another type named Value, I will suggest alias as mentioned in another PRs

aiplatform/api/AIPlatform.Samples/PredictCodeGenerationUnitTestSample.cs
amanda-tarafa: I left this comment elsewhere, I think it would be better if we showed the C# sample showing the models work with C# code.

aiplatform/api/AIPlatform.Samples.Tests/PredictCodeGenerationUnitTestTest.cs
amanda-tarafa: I think the name is weird, but it's fine, it's a test of a sample after all so not as bad as if it was the name of a public method in the library or even shown in the sample.

aiplatform/api/AIPlatform.Samples/PredictCodeGenerationUnitTestSample.cs
meteatamel: Done

aiplatform/api/AIPlatform.Samples/PredictCodeGenerationUnitTestSample.cs
meteatamel: Ok, I changed to C# and it works fine

aiplatform/api/AIPlatform.Samples/PredictCodeGenerationUnitTestSample.cs
meteatamel: Ack

aiplatform/api/AIPlatform.Samples/PredictCodeGenerationUnitTestSample.cs
```diff
@@ -0,0 +1,104 @@
...
+        var client = new PredictionServiceClientBuilder
+        {
+            Endpoint = $"{locationId}-aiplatform.googleapis.com"
+        }.Build();
+
+        // Configure the parent resource.
+        var endpoint = EndpointName.FromProjectLocationPublisherModel(projectId, locationId, publisher, model);
+
+        // Learn how to create prompts to work with a code model to generate code:
+        // https://cloud.google.com/vertex-ai/docs/generative-ai/code/code-generation-prompts
```
amanda-tarafa, line 45:
```suggestion
```

amanda-tarafa (approved):

amanda-tarafa (approved):

PR #2403, opened 2023-11-10
docs(aiplatform): Add PredictTextSummarizationSample #2392

meteatamel:


aiplatform/api/AIPlatform.Samples/PredictTextSummarizationSample.cs
```diff
@@ -0,0 +1,116 @@
...
+                }
+            })
+        };
+
+        var parameters = wkt::Value.ForStruct(new()
+        {
+            Fields =
+            {
+                { "temperature", new wkt::Value { NumberValue = 0 } },
+                { "maxDecodeSteps", new wkt::Value { NumberValue = 32 } },
```
anuragsrivstv, line 100: It's not clear to me which part of code in this sample summarizes the text rather that extracting sentiments from the text. I believe we should add comments to make this clear for the user. same goes for other similar samples as well.

meteatamel: I agree it's not really summary but rather Q&A on the provided text. However, this is what they call it in Java, Node.js, Python samples, so I'm just following their naming.

aiplatform/api/AIPlatform.Samples/PredictTextSummarizationSample.cs
```diff
@@ -0,0 +1,116 @@
...
+                { "topP", new wkt::Value { NumberValue = 0 } },
+                { "topK", new wkt::Value { NumberValue = 1 } }
+            }
+        });
+
+        // Make the request.
+        var response = client.Predict(endpoint, instances, parameters);
+
+        // Parse and return the content
+        var responseContent = response.Predictions[0].StructValue.Fields["content"].StringValue;
```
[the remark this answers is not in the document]
meteatamel: Ack

aiplatform/api/AIPlatform.Samples/PredictTextSummarizationSample.cs
amanda-tarafa: Is this explained in the docs where the sample is shown? It's really confusing.

anuragsrivstv: What actually changed apart from content ? Also what I'm trying to understand is that how can we say that this sample will summarize the provided text. Is it temperature or maxDecodeSteps value or model name or anything else ?

meteatamel: Content is the main change (also some minor params change to match other languages). It summarizes because the content (prompt) tells the model to summarize.

amanda-tarafa (approved):

Issue #2393, opened 2023-11-09
docs(aiplatform): Add PredictCodeCompletionTestFunctionSample


meteatamel:
Based on this Java sample:https://github.com/GoogleCloudPlatform/java-docs-samples/blob/HEAD/aiplatform/src/main/java/aiplatform/PredictCodeCompletionTestFunctionSample.java
Belongs in this spot in the docs: https://cloud.google.com/vertex-ai/docs/samples/aiplatform-sdk-code-completion-test-function

PR #2404, opened 2023-11-10
docs(aiplatform): Add PredictCodeCompletionTestFunctionSample #2393

meteatamel:


aiplatform/api/AIPlatform.Samples/PredictCodeCompletionTestFunctionSample.cs
```diff
@@ -0,0 +1,82 @@
...
+    return s[::-1]
+def test_empty_input_string()";
+
+        var instances = new List<wkt::Value>
+        {
+            wkt::Value.ForStruct(new()
+            {
+                Fields =
+                {
+                    ["prefix"] = wkt::Value.ForString(prefix),
```
anuragsrivstv, line 56: Same comment as on other sample. please see if we can do `using Value = Google.Protobuf.WellKnownTypes.Value;`

aiplatform/api/AIPlatform.Samples/PredictCodeCompletionTestFunctionSample.cs
```diff
@@ -0,0 +1,82 @@
...
+                    { "maxOutputTokens", new wkt::Value { NumberValue = 64 } }
+                }
+            }
+        };
+
+        // Make the request.
+        var response = client.Predict(endpoint, instances, parameters);
+
+        // Parse and return the content.
+        var content = response.Predictions[0].StructValue.Fields["content"].StringValue;
```
anuragsrivstv, line 77: I'm assuming that were are fetching first response out of all predictions as it will be most recommended answer provided by the model. If that is the case shouldn't we also add comment also why are we fetching value at index `0` , this will give user an idea how to alter sample for their use case ?

meteatamel: In reality, predictions array is always a single item..Technically, it's supposed to be a list of predictions, but I haven't seen that in my runs. So, here, we're simply picking the only element from the array.

aiplatform/api/AIPlatform.Samples/PredictCodeCompletionTestFunctionSample.cs
```diff
@@ -0,0 +1,82 @@
...
+
+        // Configure the parent resource.
+        var endpoint = EndpointName.FromProjectLocationPublisherModel(projectId, locationId, publisher, model);
+
+        // Learn how to create prompts to work with a code model to create code completion suggestions:
+        // https://cloud.google.com/vertex-ai/docs/generative-ai/code/code-completion-prompts
+        var prefix = @"
+def reverse_string(s):
+    return s[::-1]
+def test_empty_input_string()";
```
amanda-tarafa, line 48: Could this be C# code? Or does the model does not work for C#?

aiplatform/api/AIPlatform.Samples/PredictCodeCompletionTestFunctionSample.cs
meteatamel: Ack

aiplatform/api/AIPlatform.Samples/PredictCodeCompletionTestFunctionSample.cs
meteatamel: I changed to C# code. The model does not do great but it's not worse than Python.

amanda-tarafa (approved):

PR #2405, opened 2023-11-10
docs(aiplatform): Add PredictCodeGenerationFunctionSample #2394

meteatamel:


aiplatform/api/AIPlatform.Samples/PredictCodeGenerationFunctionSample.cs
```diff
@@ -0,0 +1,78 @@
...
+
+        // Learn how to create prompts to work with a code model to generate code:
+        // https://cloud.google.com/vertex-ai/docs/generative-ai/code/code-generation-prompts
+        var instances = new List<wkt::Value>
+        {
+            wkt::Value.ForStruct(new()
+            {
+                Fields =
+                {
+                    ["prefix"] = wkt::Value.ForString("Write a function that checks if a year is a leap year."),
```
hemanshv, line 51: May be we should refer to the docs that describes the usage of these constant fields like prefix, temperature etc.

aiplatform/api/AIPlatform.Samples/PredictCodeGenerationFunctionSample.cs
```diff
@@ -0,0 +1,78 @@
...
+                    { "maxOutputTokens", new wkt::Value { NumberValue = 256 } }
+                }
+            }
+        };
+
+        // Make the request.
+        var response = client.Predict(endpoint, instances, parameters);
+
+        // Parse and return the content.
+        var content = response.Predictions[0].StructValue.Fields["content"].StringValue;
```
hemanshv, line 72: Use linq instead of using magic number 0 eg `response.Predictions.First.`

hemanshv (changes requested):

aiplatform/api/AIPlatform.Samples/PredictCodeGenerationFunctionSample.cs
```diff
@@ -0,0 +1,78 @@
...
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+// [START aiplatform_sdk_code_generation_function]
+
+using Google.Cloud.AIPlatform.V1;
+using wkt = Google.Protobuf.WellKnownTypes;
```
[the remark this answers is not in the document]
meteatamel: @hemanshv The problem is there's also another `Value` in `Google.Cloud.AIPlaform.V1`, so if I dont have `wkt`, I need to refer to these with `Google.Protobuf.WellKnownTypes.Value` which is even more verbose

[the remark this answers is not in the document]
hemanshv: Ok it make sense then. Thanks for clarification.

aiplatform/api/AIPlatform.Samples/PredictCodeGenerationFunctionSample.cs
meteatamel: We already refer to the docs a little earlier in the sample

```
        // Learn how to create prompts to work with a code model to generate code:
        // https://cloud.google.com/vertex-ai/docs/generative-ai/code/code-generation-prompts
```

Let me know if you have something else in mind?

aiplatform/api/AIPlatform.Samples/PredictCodeGenerationFunctionSample.cs
meteatamel: I don't think you can use Linq against the RepeatField that Predictions are? I tried but I got some compilation error. Let me know if it works for you and if so, how.

meteatamel: Ok, fixed in the latest commit

aiplatform/api/AIPlatform.Samples.Tests/PredictCodeGenerationFunctionTest.cs
```diff
@@ -0,0 +1,37 @@
...
+    private readonly PredictCodeGenerationFunctionSample _sample;
+
+    public PredictCodeGenerationFunctionTest(AIPlatformFixture fixture)
+    {
+        _fixture = fixture;
+        _sample = new PredictCodeGenerationFunctionSample();
+    }
+
+    [Fact]
+    public void Runs()
```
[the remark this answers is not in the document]
hemanshv: I am talking about the name of the function `Runs()`. We should rename it to `TestPredictCodeGeneration()`  like the tests of other samples eg .https://github.com/GoogleCloudPlatform/dotnet-docs-samples/blob/main/spanner/api/Spanner.Samples.Tests/AddColumnAsyncPostgresTest.cs#L32

aiplatform/api/AIPlatform.Samples/PredictCodeGenerationFunctionSample.cs
hemanshv: its fine then, added so that user are aware about what else they can do. Thanks

aiplatform/api/AIPlatform.Samples.Tests/PredictCodeGenerationFunctionTest.cs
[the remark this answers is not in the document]
meteatamel: Ah, I see. Yes, that makes sense and in that case, I'd suggest to rename `Runs` to `TestPredictFunction` because that's the method that's being called. I'll see what Amanda says before making changes.

meteatamel:
Assigning to @amanda-tarafa for her review

aiplatform/api/AIPlatform.Samples.Tests/PredictCodeGenerationFunctionTest.cs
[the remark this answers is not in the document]
amanda-tarafa: Yes, it should be `TestPredictFunction` here and elsewhere accordingly.

aiplatform/api/AIPlatform.Samples/PredictCodeGenerationFunctionSample.cs
```diff
@@ -0,0 +1,79 @@
...
+        var client = new PredictionServiceClientBuilder
+        {
+            Endpoint = $"{locationId}-aiplatform.googleapis.com"
+        }.Build();
+
+        // Configure the parent resource.
+        var endpoint = EndpointName.FromProjectLocationPublisherModel(projectId, locationId, publisher, model);
+
+        // Learn how to create prompts to work with a code model to generate code:
+        // https://cloud.google.com/vertex-ai/docs/generative-ai/code/code-generation-prompts
```
amanda-tarafa, line 45: Another link to docs here. If the docs this is shown in already have explanations or links to relevant docs, this is unnecessary.

aiplatform/api/AIPlatform.Samples.Tests/PredictCodeGenerationFunctionTest.cs
```diff
@@ -0,0 +1,37 @@
+/*
```
hemanshv, line 1: placeholder to undo request changes

aiplatform/api/AIPlatform.Samples.Tests/PredictCodeGenerationFunctionTest.cs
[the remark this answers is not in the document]
meteatamel: Ack

aiplatform/api/AIPlatform.Samples/PredictCodeGenerationFunctionSample.cs
meteatamel: Removed

amanda-tarafa (approved):

PR #2449, opened 2024-01-24
docs(pubsub): Add CreateCloudStorageSubscription sample

jskeet:
First commit is just tidy-up.

Currently the test fails due to permissions ("Cloud Pub/Sub did not have the necessary permissions configured to access the provided bucket [...]") - but so does the BigQuery test (in a different way).

I thought I'd create the PR to get feedback on whether this is the right approach and I should add an IAM call to grant permission for the Pub/Sub service account to the bucket, or something else.

### DIFF
--- a/pubsub/api/Pubsub.Samples.Tests/CreateCloudStorageSubscriptionTest.cs
+++ b/pubsub/api/Pubsub.Samples.Tests/CreateCloudStorageSubscriptionTest.cs
@@ -1,0 +1,46 @@
+// Copyright 2024 Google Inc.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+using System;
+using Xunit;
+
+[Collection(nameof(PubsubFixture))]
+public class CreateCloudStorageSubscriptionTest
+{
+    private readonly PubsubFixture _pubsubFixture;
+    private readonly CreateCloudStorageSubscriptionSample _createCloudStorageSubscriptionSample;
+
+    public CreateCloudStorageSubscriptionTest(PubsubFixture pubsubFixture)
+    {
+        _pubsubFixture = pubsubFixture;
+        _createCloudStorageSubscriptionSample = new CreateCloudStorageSubscriptionSample();
+    }
+
+    [Fact]
+    public void TestCreateCloudStorageSubscription()
+    {
+        string randomName = _pubsubFixture.RandomName().Replace("-", "_");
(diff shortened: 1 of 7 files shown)


amanda-tarafa (commented):
We have this [TestBucket](https://github.com/GoogleCloudPlatform/dotnet-docs-samples/blob/main/testutil/TestBucket.cs) utility class for creating temporary buckets, you can probably use that instead of creating the buckets.

As for the permissions, and unless granting them is part of the sample we usually just manually grant them to the relevant credentials. Not sure if these permissions need to be granted on a bucket by bucket basis, or can be granted at a higher level.

jskeet:
Got it, thanks. Will have a look tomorrow. (Hadn't thought that of course we can just grant permission to all buckets instead of the specific one.)

jskeet:
Okay, the new test now passes for me locally. I'm not sure it's even running on Kokoro though - there should be 52 tests, not 51. The test that's failing is unrelated to this one, I think.

I haven't changed the code to use TestBucket - I couldn't find other examples that used it, and the code here is really pretty straightforward. Happy to adapt it if you'd like though.

amanda-tarafa (commented):
I think your test is not running because you haven't added the new files :). On the PR I only see the fixture and the tidy up changes. (When I reviewed earlier that was the case as well, I just thought you wanted me to check the fixture first?)

The failura is a service side Internal Error and that's bound to happen every so often, so yes, unrelated and not a concern.

jskeet:
Doh! Added now. Let's see what happens. (If it fails with an admin error, I can fix that - I'll just need to see which account to make a storage admin.)

amanda-tarafa:
I'll give the relevant account the needed permissions and I'll rerun the tests.

pubsub/api/Pubsub.Samples.Tests/CreateCloudStorageSubscriptionTest.cs
```diff
@@ -0,0 +1,46 @@
...
+        string randomName = _pubsubFixture.RandomName().Replace("-", "_");
+        string topicId = $"testTopicForCreateCloudStorageSubscription{randomName}";
+        string subscriptionId = $"testSubscriptionForCreateCloudStorageSubscription{randomName}";
+        string bucket = _pubsubFixture.CloudStorageBucketName;
+
+        _pubsubFixture.CreateTopic(topicId);
+
+        var subscription = _createCloudStorageSubscriptionSample.CreateCloudStorageSubscription(
+            _pubsubFixture.ProjectId, topicId, subscriptionId, bucket, "prefix", "suffix", TimeSpan.FromMinutes(5));
+        _pubsubFixture.TempSubscriptionIds.Add(subscriptionId);
```
amanda-tarafa, line 42: We usually do this before calling the sample, in case the subscription is effectively created but there's an error after, in sample code maybe. That way we guarantee that we delete the subscrition no matter what. The Fixture ignores errors that happen on dispose.

amanda-tarafa (dismissed):
LGTM, a tiny change.

pubsub/api/Pubsub.Samples.Tests/CreateCloudStorageSubscriptionTest.cs
jskeet: I copied this from a BQ-based subscription - I'll make the same change in that one as well.

jskeet: Done.

PR #2469, opened 2024-02-09
docs(pubsub): Add samples for schema revisions

jskeet:
This PR is large due to the first commit which is a wide refactoring (but makes everything so much nicer).

### DIFF
--- a/pubsub/api/Pubsub.Samples.Tests/CommitAvroSchemaTest.cs
+++ b/pubsub/api/Pubsub.Samples.Tests/CommitAvroSchemaTest.cs
@@ -1,0 +1,41 @@
+// Copyright 2024 Google Inc.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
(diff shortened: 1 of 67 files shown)


pubsub/api/Pubsub.Samples.Tests/CommitAvroSchemaTest.cs
```diff
@@ -0,0 +1,41 @@
+// Copyright 2021 Google Inc.
```
amanda-tarafa, line 1: nit, 2024

amanda-tarafa (dismissed):
Thanks for the first commit!

pubsub/api/Pubsub.Samples.Tests/CommitAvroSchemaTest.cs
jskeet: Will fix - and check the others

PR #2485, opened 2024-02-14
docs(bigquery): Add bigquery_create_table_range_partitioned sample 

jskeet:


amanda-tarafa (approved):
The failure is an unrelated sample which I haven't seen fail before, probably transient. If it happens again I'll create an issue to track it.

Issue #2511, opened 2024-02-29
[dlp]: InspectImagesForSensitiveDataWithInfoTypesTests has started failing


amanda-tarafa:
```
Failed GoogleCloudSamples.InspectImagesForSensitiveDataWithInfoTypesTests.TestInspectImage [294 ms]
  Error Message:
   Assert.Equal() Failure: Values differ
Expected: 3
Actual:   4
  Stack Trace:
     at GoogleCloudSamples.InspectImagesForSensitiveDataWithInfoTypesTests.TestInspectImage() in C:\tmpfs\src\github\dotnet-docs-samples\dlp\api\Snippets.Tests\InspectImagesForSensitiveDataWithInfoTypesTests.cs:line 30
```

It can be seen [here](https://source.cloud.google.com/results/invocations/30a186cb-ea9f-4003-9971-96916658c3e6/log) but it has failed on nightly CI as well.

amanda-tarafa:
@annadata FYI

PR #2527, opened 2024-03-06
Fix failing InspectImage test from issue #2511

patrickmoy:
Test as written was somewhat flaky, verifying only on the number of findings. Location detector gave an unlikely but valid additional finding which increased the response finding count. 

Fixed by setting specific InfoTypes to search for (thus ignoring unlikely fringe findings) and updating test assertions.


amanda-tarafa:
@patrickmoy Tests are green and these changes look good to me. Any reason this is still in draft? Thanks.

patrickmoy:
Oh, the last run there were some failures, but they looked unrelated to the changes, so I wanted to wait for the next run.

I did get one odd timeout, so I tried increasing the time limit, but the next run finished in a few minutes so I'll revert that one change before sending it in.

amanda-tarafa:
Yep, I think the "too many buckets creation" errors was because there were some dependency-updating PRs running at the same time. I'll review and we can wait for tests to go green again before merging. Thanks!

amanda-tarafa (approved):
Thanks!

amanda-tarafa:
This is the Storage quota again. I think it's safe to merge. I'll keep an eye on nightly CI just in case, but I think this is just a coincidence of a few PRs running tests at the same time.

PR #2590, opened 2024-04-05
tests[Spanner]: Fix flake in UpdateUsingDmlReturningAsyncTest

amanda-tarafa:
- We had an `async void` inserting data that we needed for the test. The data was not always available when needed.
- Also, simplified the tests.
- I searched for others `async void`, there are none.


jskeet (approved):
Don't really understand why this was written that way before, but looks fine now...

PR #2614, opened 2024-04-30
Added samples for Gemini 1.5

meteatamel:
* Added some samples for Gemini 1.5. 
* They're all very similar samples with slight differences, that's why I'm doing it in a single PR. 
* Fixes b/336757387, b/336757211, b/336757534, b/336757434, b/336758092

### DIFF
--- a/aiplatform/api/AIPlatform.Samples/VideoInputWithAudio.cs
+++ b/aiplatform/api/AIPlatform.Samples/VideoInputWithAudio.cs
@@ -1,0 +1,73 @@
+/*
+ * Copyright 2024 Google LLC
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     https://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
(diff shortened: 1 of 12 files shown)


jskeet:
Reassigned to Amanda as a better samples reviewer - but feel free to assign it back, @amanda-tarafa

aiplatform/api/AIPlatform.Samples/VideoInputWithAudio.cs
```diff
@@ -0,0 +1,73 @@
...
+        content.Parts.AddRange(new List<Part>()
+        {
+            new() {
+                Text = prompt
+            },
+            new() {
+                FileData = new() {
+                    MimeType = "video/mp4",
+                    FileUri = "gs://cloud-samples-data/generative-ai/video/pixel8.mp4"
+                }
```
amanda-tarafa, line 54: nit: This braces should be on the next line.

amanda-tarafa (approved):
LGTM

But I noticed that for the anonymous types initialization you are using something like JSON syntax by keeping the openening bracket on the same line as `new`? Also using parentheses on all the `new`s.

In a separate PR, can you change that? So when you have things like:

```csharp
Field = new() {
    Innerfield = value
}
```

do this instead?

```csharp
Field = new
{
    Innerfield = value
}
```

On these samples and all the others. Thanks!

aiplatform/api/AIPlatform.Samples/VideoInputWithAudio.cs
meteatamel: @amanda-tarafa I'm handling this in a separate PR

PR #2621, opened 2024-05-06
feat(storage): add quickstart sample for storage control

JesseLovelace:
Adds a quickstart sample for storage control

### DIFF
--- a/storage/api/Storage.Samples/StorageControlQuickstart.cs
+++ b/storage/api/Storage.Samples/StorageControlQuickstart.cs
@@ -1,0 +1,33 @@
+// Copyright 2024 Google LLC
+//
+// Licensed under the Apache License, Version 2.0 (the "License").
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+// https://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+// [START storage_control_quickstart_sample]
+using Google.Cloud.Storage.Control.V2;
+using System;
+
+public class StorageControlQuickstartSample
+{
+    public StorageLayout StorageControlQuickstart(string bucketName = "your-unique-bucket-name")
+    {
+        StorageControlClient storageControl = StorageControlClient.Create();
+        GetStorageLayoutRequest request = new GetStorageLayoutRequest
+        {
+            Name = StorageLayoutName.Format("_", bucketName), // Set project to "_" for global bucket"
+        };
+        StorageLayout response = storageControl.GetStorageLayout(request);
+        Console.WriteLine($"Bucket {bucketName} has location type {response.LocationType}");
+        return response;
+    }
+}
+// [END storage_control_quickstart_sample]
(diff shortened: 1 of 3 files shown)


storage/api/Storage.Samples/StorageControlQuickstart.cs
```diff
@@ -0,0 +1,33 @@
...
+using System;
+
+public class StorageControlQuickstartSample
+{
+    public StorageLayout StorageControlQuickstart(string bucketName = "your-unique-bucket-name")
+    {
+        StorageControlClient storageControl = StorageControlClient.Create();
+        GetStorageLayoutRequest request = new GetStorageLayoutRequest
+        {
+            Name = StorageLayoutName.Format("_", bucketName), // Set project to "_" for global bucket"
```
jskeet, line 26: Rather than explicitly formatting to a string, I'd set the `StorageLayoutName` property instead (if you really need a request object - see below):

```csharp
GetStorageLayoutRequest request = new GetStorageLayoutRequest
{
    StorageLayoutName = new StorageLayoutName("_", bucketName");
}
```

storage/api/Storage.Samples/StorageControlQuickstart.cs
```diff
@@ -0,0 +1,33 @@
...
+        };
+        StorageLayout response = storageControl.GetStorageLayout(request);
```
jskeet, line 28: This is a relatively long-winded way of doing things. We generate an overload accepting the name:

```csharp
StorageLayout response = storageControl.GetStorageLayout(new StorageLayoutName("_", bucket));
```

(By all means separate out the StorageLayoutName initialization to a separate variable.)

amanda-tarafa (commented):
This looks good, but I wonder if we should create a separate folder for Storage control? I think we should.

jskeet (commented):
I don't know whether @amanda-tarafa has different preferences in terms of resource name usage and overloads vs request objects.

storage/api/Storage.Samples/StorageControlQuickstart.cs
amanda-tarafa: Yes +1 to this.

storage/api/Storage.Samples/StorageControlQuickstart.cs
amanda-tarafa: There is a sample guideline, or at least there used to be one, about showing the request object overload for most samples. This is meant to be a hint for users that "there's likely more you can do".

I'm usually lenient with whatever approach is used, so I'm fine either way. I agree that it is more legible to use the simpler overload though, if that's enough.

storage/api/Storage.Samples/StorageControlQuickstart.cs
```diff
@@ -0,0 +1,32 @@
...
+using System;
+
+public class StorageControlQuickstartSample
+{
+    public StorageLayout StorageControlQuickstart(string bucketName = "your-unique-bucket-name")
+    {
+        StorageControlClient storageControl = StorageControlClient.Create();
+
+        // Using "_" for projectId means global bucket namespace
+        StorageLayout response = storageControl.GetStorageLayout(new StorageLayoutName("_", bucketName));
```
amanda-tarafa, line 26:
```suggestion
        var layoutName = new StorageLayoutName("_", bucketName);
        StorageLayout response = storageControl.GetStorageLayout(layoutName);
```

amanda-tarafa: I'll apply this suggestion myself.

amanda-tarafa (approved):

amanda-tarafa (approved):

PR #2630, opened 2024-05-09
docs(aiplatform): Change all samples to use inline initialization #2626

meteatamel:
Updated all samples to initialize requests inline per @jskeet's suggestion

### DIFF
--- a/aiplatform/api/AIPlatform.Samples/MultimodalMultiImage.cs
+++ b/aiplatform/api/AIPlatform.Samples/MultimodalMultiImage.cs
@@ -19,7 +19,6 @@
 using Google.Api.Gax.Grpc;
 using Google.Cloud.AIPlatform.V1;
 using Google.Protobuf;
-using System.Collections.Generic;
 using System.Net.Http;
 using System.Text;
 using System.Threading.Tasks;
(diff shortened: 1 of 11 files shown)


aiplatform/api/AIPlatform.Samples/MultimodalMultiImage.cs
```diff
@@ -49,60 +46,30 @@ public class MultimodalMultiImage
...
             }
-        });
-
-        var generateContentRequest = new GenerateContentRequest
-        {
-            Model = $"projects/{projectId}/locations/{location}/publishers/{publisher}/models/{model}"
         };
-        generateContentRequest.Contents.Add(content);
 
-        // Make the request, returning a streaming response
```
jskeet, line 100: Were the comments here and at line 105 deliberately removed?

jskeet: (Ditto other samples)

jskeet (approved):
Not sure it was worth removing the comments, but I'm fine with it if that was deliberate.

Personally I find the new style *much* easier to understand in terms of the structure of the request - thanks!

aiplatform/api/AIPlatform.Samples/MultimodalMultiImage.cs
meteatamel: Yes. I only kept comments in the GeminiQuickstart but the rest wasn't really adding much value.

amanda-tarafa (approved):
Oh yes, nice!

Issue #2652, opened 2024-05-17
docs(aiplatform): Add GenerateImage sample


meteatamel:
Based on this Python sample: https://github.com/GoogleCloudPlatform/python-docs-samples/blob/main/generative_ai/imagen/generate_image.py

Place in docs: https://cloud.google.com/vertex-ai/generative-ai/docs/samples/generativeaionvertexai-imagen-generate-image

meteatamel:
Added and documented

PR #2653, opened 2024-05-17
docs(aiplatform): Add GenerateImage sample #2652

meteatamel:


aiplatform/api/AIPlatform.Samples.Tests/GenerateImageTest.cs
```diff
@@ -0,0 +1,39 @@
+/*
+ * Copyright 2023 Google LLC
```
jskeet, line 2: 2024

aiplatform/api/AIPlatform.Samples/GenerateImage.cs
```diff
@@ -0,0 +1,74 @@
...
+            {
+                Fields =
+                {
+                    ["sampleCount"] = Value.ForNumber(1)
+                }
+            })
+        };
+
+        PredictResponse response = await predictionServiceClient.PredictAsync(predictRequest);
+        byte[] bytesBase64Encoded = Convert.FromBase64String(response.Predictions.First().StructValue.Fields["bytesBase64Encoded"].StringValue);
```
jskeet, line 64: Eek, is there no way of persuading the API to give the bytes directly as a bytestring? That's unfortunate.

(But I'd argue that the variable name should just be `imageData` or something similar - it's *not* base64-encoded, because you've done the decoding in the initialization expression.)

meteatamel: good point, changed to `imageBytes`

jskeet (approved):

amanda-tarafa (approved):

PR #2697, opened 2024-06-24
[Gemini] Create C# samples for Controlled Generation

meteatamel:


jskeet (approved):
Can't say I fully understand what the samples are for without doing more research, but they look okay in and of themselves.

meteatamel:
> Can't say I fully understand what the samples are for without doing more research, but they look okay in and of themselves.

The samples are about "controlled generation", a way to enforce the output from LLM to have a certain format (JSON) or schema. The corresponding Python samples are [here](https://github.com/GoogleCloudPlatform/python-docs-samples/tree/main/generative_ai/controlled_generation). 

After I implemented, I compared the output of the tests from the two languages and they're almost the same (minus the LLM  randomness differences).

amanda-tarafa (approved):
Can I ask you to split these into multiple files, one per sample and one per test? If you need these sooner I'm happy to merge as is and then split, LGTMing for that.

amanda-tarafa:
We've talked about this offline and because the only changes between samples is the schema, it makes sense to have them all on the same file. Merging...

PR #2708, opened 2024-07-03
feat(storagecontrol): managed folders samples

JesseLovelace:
Samples for managed folders feature

### DIFF
--- a/storage/api/Storage.Samples/StorageControlCreateManagedFolder.cs
+++ b/storage/api/Storage.Samples/StorageControlCreateManagedFolder.cs
@@ -1,0 +1,40 @@
+// Copyright 2024 Google LLC
+//
+// Licensed under the Apache License, Version 2.0 (the "License").
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+// https://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+// [START storage_control_managed_folder_create]
+using Google.Cloud.Storage.Control.V2;
+using System;
+
+public class StorageControlCreateManagedFolderSample
+{
+    public ManagedFolder StorageControlCreateManagedFolder(string bucketName = "your-unique-bucket-name",
+        string managedFolderId = "your_managed_folder_id")
+    {
+        StorageControlClient storageControl = StorageControlClient.Create();
+
+        var request = new CreateManagedFolderRequest()
+        {
+            // Set project to "_" to signify globally scoped bucket
+            Parent = BucketName.FormatProjectBucket("_", bucketName),
+            ManagedFolder = new ManagedFolder(),
+            ManagedFolderId = managedFolderId
+        };
+
+        ManagedFolder managedFolder = storageControl.CreateManagedFolder(request);
+
+        Console.WriteLine($"Managed Folder {managedFolderId} created in bucket {bucketName}");
+        return managedFolder;
+    }
+}
+// [END storage_control_managed_folder_create]
--- a/storage/api/Storage.Samples/StorageControlDeleteManagedFolder.cs
+++ b/storage/api/Storage.Samples/StorageControlDeleteManagedFolder.cs
@@ -1,0 +1,35 @@
+// Copyright 2024 Google LLC
+//
+// Licensed under the Apache License, Version 2.0 (the "License").
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+// https://www.apache.org/licenses/LICENSE-2.0
+//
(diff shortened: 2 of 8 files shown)


storage/api/Storage.Samples/StorageControlDeleteManagedFolder.cs
```diff
@@ -0,0 +1,35 @@
...
+using System;
+
+public class StorageControlDeleteManagedFolderSample
+{
+    public void StorageControlDeleteManagedFolder(string bucketName = "your-unique-bucket-name",
+        string managedFolderId = "your_managed_folder_id")
+    {
+        StorageControlClient storageControl = StorageControlClient.Create();
+
+        string managedFolderResourceName =
```
jskeet, line 26: I think it would be better to just create a new ManagedFolderName and call DeleteManagedFolder(ManagedFolderName). IMO we should avoid the Format calls as far as possible, using the resource name classes.

storage/api/Storage.Samples/StorageControlCreateManagedFolder.cs
```diff
@@ -0,0 +1,40 @@
...
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+// [START storage_control_managed_folder_create]
+using Google.Cloud.Storage.Control.V2;
+using System;
+
+public class StorageControlCreateManagedFolderSample
+{
+    public ManagedFolder StorageControlCreateManagedFolder(string bucketName = "your-unique-bucket-name",
```
jskeet, line 21: I think this would be better as `bucketId` instead of `bucketName` - it means we can be consistent that a bucket resource name (as represented by a BucketName) consists of a project ID (or _) and a bucket ID.

storage/api/Storage.Samples/StorageControlCreateManagedFolder.cs
```diff
@@ -0,0 +1,40 @@
...
+        string managedFolderId = "your_managed_folder_id")
+    {
+        StorageControlClient storageControl = StorageControlClient.Create();
+
+        var request = new CreateManagedFolderRequest()
```
jskeet, line 26: Any reason for creating a request rather than just calling

```csharp
ManagedFolder managedFolder = storageControl.CreateManagedFolder(
    new BucketName("_", bucketName),
    new ManagedFolder(),
    managedFolderId);
```

storage/api/Storage.Samples/StorageControlGetManagedFolder.cs
```diff
@@ -0,0 +1,36 @@
...
+using System;
+
+public class StorageControlGetManagedFolderSample
+{
+    public ManagedFolder StorageControlGetManagedFolder(string bucketName = "your-unique-bucket-name",
+        string managedFolderId = "your_managed_folder_Id")
+    {
+        StorageControlClient storageControl = StorageControlClient.Create();
+
+        string managedFolderResourceName =
```
jskeet, line 26: Again, let's use a ManagedFolderName rather than formatting as a string.

storage/api/Storage.Samples/StorageControlListManagedFolders.cs
```diff
@@ -0,0 +1,37 @@
...
+using System.Collections.Generic;
+
+public class StorageControlListManagedFoldersSample
+{
+    public IEnumerable<ManagedFolder> StorageControlListManagedFolders(string bucketName = "your-unique-bucket-name")
+    {
+        StorageControlClient storageControl = StorageControlClient.Create();
+
+        // Use "_" for project ID to signify globally scoped bucket
+        string bucketResourceName = BucketName.FormatProjectBucket("_", bucketName);
```
jskeet, line 27: And again (but BucketName this time).

storage/api/Storage.Samples/StorageControlCreateManagedFolder.cs
jskeet: (The same goes for all samples.)

storage/api/Storage.Samples/StorageControlDeleteManagedFolder.cs
```diff
@@ -0,0 +1,35 @@
...
+using System;
+
+public class StorageControlDeleteManagedFolderSample
+{
+    public void StorageControlDeleteManagedFolder(string bucketId = "your-unique-bucket-name",
+        string managedFolderId = "your_managed_folder_id")
+    {
+        StorageControlClient storageControl = StorageControlClient.Create();
+
+        ManagedFolderName managedFolderResourceName =
```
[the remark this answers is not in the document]
JesseLovelace: In my opinion, when a user thinks about something like "managedFodlerName", they're probably just thinking about the name that they gave the folder, like "myFolder'. But since it's not that, it's something like "projects/_/buckets/mybucket/managedFolders/myFolder", calling it a resourceName makes it a little more clear that it's something more qualified than just the name they gave the folder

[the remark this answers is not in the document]
jskeet: Sure. I could go back and forth on that - we *want* users to start thinking that "name = resource name, ID = last part" but I'm not going to push back on this :)

amanda-tarafa:
Merging now...

PR #2714, opened 2024-07-09
build: CI supports .NET 8

amanda-tarafa:
And KMS samples and tests are migrated to .NET 8 to confirm things work.


jskeet (dismissed):
Yup, this looks encouraging.

amanda-tarafa:
Windows ran KMS tests in .NET, but Linux didn't. I'm looking at this.

amanda-tarafa:
Both Linux and Windows are now targetting .NET 8 for running KMS tests.
I'm making a few more changes that I will push shortly and we should be good.

amanda-tarafa:
(I'm looking at these failures)

bigtable/api/Snippets/Reads.cs
```diff
@@ -137,7 +144,10 @@ public string readRowRanges(string projectId = "YOUR-PROJECT-ID", string instanc
             ReadRowsStream readRowsStream = bigtableClient.ReadRows(tableName, rowSet);
```
jskeet, line 144: We should probably have using statements for all of these streams...

bigtable/api/Snippets/Writes/WriteBatch.cs
```diff
@@ -37,7 +37,7 @@ public class WriteBatch
 
             TableName tableName = new TableName(projectId, instanceId, tableId);
             BigtableVersion timestamp = new BigtableVersion(DateTime.UtcNow);
-            String COLUMN_FAMILY = "stats_summary";
+            string COLUMN_FAMILY = "stats_summary";
```
jskeet, line 40: As a follow-up we should probably rename this and make it a const (even if it's still as a local variable)

jskeet (approved):

bigtable/api/Snippets/Reads.cs
amanda-tarafa: Oh didn't see this before merging, but yep, I'll do that.

bigtable/api/Snippets/Writes/WriteBatch.cs
amanda-tarafa: Yes, I think we shoulc rewrite a lot of these anyway. Let's what we can get done medium term.

amanda-tarafa (commented):
Sorry, I merged before seeing your comments. I'll send a PR with usings for the streams shortly.

jskeet:
> Sorry, I merged before seeing your comments.

That's fine, I'd intended them to be "for future refinement" anyway :)

bigtable/api/Snippets/Reads.cs
amanda-tarafa: ReadRowStream is not disposable, I've created b/352471768.

PR #2722, opened 2024-07-12
ci: Add AIPlatform projects to the existing solution

amanda-tarafa:
This is so the test script can pick up the projects to restore and run tests from.

Also, upgrading to .NET 8


jskeet (approved):

amanda-tarafa:
Errors are permissions, I gave permissions but it seems like its taking a little while to propagate, I'll try in a few minutes.

jskeet (approved):

amanda-tarafa:
These are working now, excepto for a couple of "INTERNAL" errors, we sould maybe look at retrying if those end up being commonplace, let's keep and eye out. Thanks @meteatamel

PR #2726, opened 2024-07-12
chore: Stop automatic upgrade of some samples to .NET 8

amanda-tarafa:
For these, we need to confirm that the environments they are meant for support .NET 8 and documentation needs to be updated also.


amanda-tarafa:
I'll sefl merge this and test it by retrying #2719

PR #2739, opened 2024-07-23
feat(Video Stitcher): Add VOD config samples and tests; VOD sessions …

irataxy:
…now use VOD configs

See https://cloud.google.com/video-stitcher/docs/how-to/managing-vod-configs

b:347726038

### DIFF
--- a/media/stitcher/api/Stitcher.Samples/GetVodConfig.cs
+++ b/media/stitcher/api/Stitcher.Samples/GetVodConfig.cs
@@ -1,5 +1,5 @@
 ﻿/*
- * Copyright 2022 Google LLC
+ * Copyright 2024 Google LLC
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
@@ -14,34 +14,28 @@
  * limitations under the License.
  */
 
-// [START videostitcher_create_vod_session]
+// [START videostitcher_get_vod_config]
 
-using Google.Api.Gax.ResourceNames;
 using Google.Cloud.Video.Stitcher.V1;
 
-public class CreateVodSessionSample
+public class GetVodConfigSample
 {
-    public VodSession CreateVodSession(
+    public VodConfig GetVodConfig(
         string projectId, string location, string vodConfigId)
     {
         // Create the client.
         VideoStitcherServiceClient client = VideoStitcherServiceClient.Create();
 
-        CreateVodSessionRequest request = new CreateVodSessionRequest
+        GetVodConfigRequest request = new GetVodConfigRequest
         {
-            ParentAsLocationName = LocationName.FromProjectLocation(projectId, location),
-            VodSession = new VodSession
-            {
-                VodConfig = VodConfigName.FormatProjectLocationVodConfig(projectId, location, vodConfigId),
-                AdTracking = AdTracking.Server
-            }
+            VodConfigName = VodConfigName.FromProjectLocationVodConfig(projectId, location, vodConfigId)
         };
 
         // Call the API.
-        VodSession session = client.CreateVodSession(request);
+        VodConfig response = client.GetVodConfig(request);
 
         // Return the result.
-        return session;
+        return response;
     }
 }
-// [END videostitcher_create_vod_session]
+// [END videostitcher_get_vod_config]
--- a/media/stitcher/api/Stitcher.Samples/ListVodConfigs.cs
+++ b/media/stitcher/api/Stitcher.Samples/ListVodConfigs.cs
@@ -19,6 +19,7 @@
 using Google.Api.Gax;
 using Google.Api.Gax.ResourceNames;
 using Google.Cloud.Video.Stitcher.V1;
+using System;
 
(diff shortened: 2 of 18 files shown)


media/stitcher/api/Stitcher.Samples/ListVodConfigs.cs
```diff
@@ -0,0 +1,43 @@
...
+        // Create the client.
+        VideoStitcherServiceClient client = VideoStitcherServiceClient.Create();
+
+        ListVodConfigsRequest request = new ListVodConfigsRequest
+        {
+            ParentAsLocationName = LocationName.FromProjectLocation(projectId, regionId)
+        };
+
+        // Make the request.
+        PagedEnumerable<ListVodConfigsResponse, VodConfig> response = client.ListVodConfigs(request);
```
amanda-tarafa, line 37: Is this how all the other list samples are for Video Stitcher? Just return the response? We usually like to show how to iterate with a foreach or similar.

media/stitcher/api/Stitcher.Samples/GetVodConfig.cs
```diff
@@ -0,0 +1,41 @@
...
+        // Create the client.
+        VideoStitcherServiceClient client = VideoStitcherServiceClient.Create();
+
+        GetVodConfigRequest request = new GetVodConfigRequest
+        {
+            VodConfigName = VodConfigName.FromProjectLocationVodConfig(projectId, location, vodConfigId)
+        };
+
+        // Call the API.
+        VodConfig response = client.GetVodConfig(request);
```
amanda-tarafa, line 35: For simple requests we usually recommend the overload that takes the resource name (instead of creating te request object). I can't remember if we agreed to something else for Video Stitcher?

amanda-tarafa (commented):
A couple of comments only, and mostly to confirm things match other Video Stitcher samples.

media/stitcher/api/Stitcher.Samples/GetVodConfig.cs
irataxy: The rest of the samples create a request object. Same for Live Stream API.

media/stitcher/api/Stitcher.Samples/ListVodConfigs.cs
irataxy: We just return the response in the other list samples. I added an iterator for VOD-related resources. If that looks good, I'll add to all of the other list samples.

amanda-tarafa: This looks good. You may want to change the tests so that the collection is no iterated over twice, but because we won't have much to list, that's really not a problem, so I'll leave it up to you.

media/stitcher/api/Stitcher.Samples/GetVodConfig.cs
amanda-tarafa: Acked, that's fine then.

amanda-tarafa (approved):
The changes look good, but the tests are failing because of quota. I've seen this happen before, when several test runs happen at the same time. Can you double check (in a separate PR) that we are not forgetting to delete CDN Keys anywhere?

```
[xUnit.net 00:02:43.72]     Stitcher.Samples.Tests.CreateCdnKeyAsyncTest.CreatesCdnKeyAsync_MediaCdn [FAIL]
  Failed Stitcher.Samples.Tests.CreateCdnKeyAsyncTest.CreatesCdnKeyAsync_MediaCdn [107 ms]
  Error Message:
   Grpc.Core.RpcException : Status(StatusCode="ResourceExhausted", Detail="Quota limit 'CdnKeysPerProjectPerRegion' has been exceeded. Limit: 10 in region us-central1.")
```

irataxy:
> The changes look good, but the tests are failing because of quota. I've seen this happen before, when several test runs happen at the same time. Can you double check (in a separate PR) that we are not forgetting to delete CDN Keys anywhere?
> 
> ```
> [xUnit.net 00:02:43.72]     Stitcher.Samples.Tests.CreateCdnKeyAsyncTest.CreatesCdnKeyAsync_MediaCdn [FAIL]
>   Failed Stitcher.Samples.Tests.CreateCdnKeyAsyncTest.CreatesCdnKeyAsync_MediaCdn [107 ms]
>   Error Message:
>    Grpc.Core.RpcException : Status(StatusCode="ResourceExhausted", Detail="Quota limit 'CdnKeysPerProjectPerRegion' has been exceeded. Limit: 10 in region us-central1.")
> ```

I looked into this and ran it locally several times. All of the test resources are cleaned up, but the test will create 5 CDN keys every run (2 in the fixture and 3 create samples). I can request an increased CDN key quota for the test project if this is a recurring issue.

amanda-tarafa:
> I looked into this and ran it locally several times. All of the test resources are cleaned up, but the test will create 5 CDN keys every run (2 in the fixture and 3 create samples). I can request an increased CDN key quota for the test project if this is a recurring issue.

Thanks, I'll keep an eye out and let you know. I really doesn't happen that often, so let's leave it as is for now.

PR #2744, opened 2024-07-29
chore: cleanup unused region tags

msampathkumar:
Adding `generativeaionvertexai_` region tags to replace old `aiplatform_`

In near future GenAI code samples, will only use `generativeaionvertexai_` region tag prefix.


amanda-tarafa:
(Will review when it's ready)

amanda-tarafa (approved):
Thanks!

PR #2748, opened 2024-08-01
Add regional endpoints code samples

jakubrauch:
Tested new code in `dlp/api` with `(cd Snippets.Tests && pwsh runTest.ps1)`.

Please merge when approved.


dlp/api/Snippets/InspectStringRep.cs
```diff
@@ -0,0 +1,90 @@
...
+        {
+            Parent = new LocationName(projectId, repLocation).ToString(),
+            Item = new ContentItem
+            {
+                Value = dataValue
+            },
+            InspectConfig = inspectConfig
+        };
+
+        var dlp = new DlpServiceClientBuilder{
```
amanda-tarafa, line 57:
```suggestion
        var dlp = new DlpServiceClientBuilder
        {
```

dlp/api/Snippets.Tests/InspectStringRepTests.cs
```diff
@@ -0,0 +1,139 @@
...
+            string repLocation = "us-west1";
+            var infoTypes = new InfoType[] { new InfoType { Name = "PERSON_NAME" } };
+            var customInfoTypes = new CustomInfoType[] { };
+            var response = InspectStringRep.Inspect(Fixture.ProjectId, repLocation, testData, Likelihood.Possible.ToString(), 5, true, infoTypes, customInfoTypes);
+
+            Assert.Contains(response.Result.Findings, f => f.Quote == "Robert");
+        }
+
+        [Fact]
+        public void TestInspectString_NoResults()
```
amanda-tarafa, line 43: We usually don't test samples so extensively. Sample tests are just for testing that the sample itself works, not for testing Client Library or API functionality, which should be tested elsewhere. These many tests are a maintenance burden and may provoke quota issues, etc.

I think we should be fine with just the first test on this file.

amanda-tarafa (changes requested):
Thanks!

dlp/api/Snippets/InspectStringRep.cs
jakubrauch: Updated

dlp/api/Snippets.Tests/InspectStringRepTests.cs
jakubrauch: Thanks, removed other four.

amanda-tarafa (approved):

PR #2765, opened 2024-08-21
chore: Revert to Google.Cloud.Spanner.Data version 5.0.0-beta03

jskeet:
Assuming we want the samples to refer to GA library versions, we won't be able to update the dependency until 5.0.0 has been released.


jskeet:
(Will self-merge.)

PR #2831, opened 2024-10-26
feat[Spanner]: Add edition samples and tests

amanda-tarafa:
And creates the Instances we use for testing in the Enterprise Plus edition.

Minor refactoring of fixture code.

Closes b/331523930
Fixes b/371116134

### DIFF
--- a/spanner/api/Spanner.Samples.Tests/UpdateInstanceAsyncTest.cs
+++ b/spanner/api/Spanner.Samples.Tests/UpdateInstanceAsyncTest.cs
@@ -1,0 +1,38 @@
+// Copyright 2024 Google LLC
+//
+// Licensed under the Apache License, Version 2.0 (the "License").
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at 
+//
+// https://www.apache.org/licenses/LICENSE-2.0 
+//
+// Unless required by applicable law or agreed to in writing, software 
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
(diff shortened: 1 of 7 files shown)


spanner/api/Spanner.Samples.Tests/UpdateInstanceAsyncTest.cs
```diff
@@ -0,0 +1,38 @@
...
+public class UpdateInstanceAsyncTest
+{
+    private readonly SpannerFixture _spannerFixture;
+
+    public UpdateInstanceAsyncTest(SpannerFixture spannerFixture) => _spannerFixture = spannerFixture;
+
+    [Fact]
+    public async Task TestUpdateInstanceAsync()
+    {
+        UpdateInstanceAsyncSample updateInstanceSample = new UpdateInstanceAsyncSample();
```
jskeet, line 31: If there's only one instance used across all these tests, we're not really testing that the sample actually updates it - it feels like it might already be an Enterprise Plus instance.

Presumably we're not being *too* rigorous about this because these are just samples, not integration tests?

jskeet (approved):
One question around what exactly we're doing in terms of test instances (and I may be missing something) but otherwise fine.

spanner/api/Spanner.Samples.Tests/UpdateInstanceAsyncTest.cs
amanda-tarafa: Yes, you are right with everything. We keep one instance only (except for two cases where we create specific instances). And yes, we don't treat samples as integration tests, so we just test that the sample code doesn't fail, etc.

Creating instances is costly, the quota for admin operations is low etc.

PR #2832, opened 2024-10-29
feat[Spanner]: Add Multi Region Encryption samples

amanda-tarafa:


amanda-tarafa:
@panerorenn9541 FYI

spanner/api/Spanner.Samples.Tests/CopyBackupWithMultiRegionEncryptionAsyncTest.cs
```diff
@@ -0,0 +1,51 @@
...
+    public CopyBackupWithMultiRegionEncryptionAsyncTest(SpannerFixture fixture)
+    {
+        _fixture = fixture;
+    }
+
+    [SkippableFact]
+    public async Task TestCopyBackupWithMultiRegionEncryptionAsync()
+    {
+        CopyBackupWithMultiRegionEncryptionAsyncSample copyBackupSample = new CopyBackupWithMultiRegionEncryptionAsyncSample();
+        string source_Project_id = _fixture.ProjectId;
```
jskeet, line 35: Any reason these are using underscores instead of `sourceProjectId` etc?

You could potentially introduce a record type to make this simpler:

```csharp
var source = new BackupName(_fixture.ProjectId, _fixture.InstanceId, _fixture.BackupId);
var target = new BackupName(_fixture.ProjectId, _fixture.InstanceId, SpannerFixture.GenerateId("test_", 16));
...
var backup = await copyBackupSample.CopyBackupWithMultiRegionEncryptionAsync(
    source.InstanceId, source.ProjectId, source.BackupId,
    target.InstanceId, target.ProjectId, target.BackupId,
    expireTime, kmsKeyNames);
...
}

// As a nested type
public record BackupName(string ProjectId, string InstanceId, string BackupId)
```

jskeet: This is assuming there isn't a good BackupName class already, of course - I'd really have hoped there would be...

spanner/api/Spanner.Samples.Tests/CreateBackupWithMultiRegionEncryptionAsyncTest.cs
```diff
@@ -0,0 +1,43 @@
...
+
+    public CreateBackupWithMultiRegionEncryptionAsyncTest(SpannerFixture fixture)
+    {
+        _fixture = fixture;
+    }
+
+    [SkippableFact]
+    public async Task TestCreateBackupWithMultiRegionEncryptionAsync()
+    {
+        Skip.If(!_fixture.RunCmekBackupSampleTests, SpannerFixture.SkipCmekBackupSamplesMessage);
```
jskeet, line 37: Perhaps change RunCmekBackupSampleTests to SkipCmekBackupSampleTests for consistency between the two properties, and to avoid the negation?

spanner/api/Spanner.Samples.Tests/CopyBackupWithMultiRegionEncryptionAsyncTest.cs
jskeet: It looks like we *do* have BackupName, because it's used later. Just use that?

spanner/api/Spanner.Samples/CopyBackupWithMultiRegionEncryptionAsync.cs
```diff
@@ -0,0 +1,70 @@
...
+        {
+            Console.WriteLine($"Error while copying backup: {completedResponse.Exception}");
+            throw completedResponse.Exception;
+        }
+
+        Backup backup = completedResponse.Result;
+
+        Console.WriteLine($"Backup copied successfully.");
+        Console.WriteLine($"Backup with Id {sourceBackupId} has been copied from {sourceProjectId}/{sourceInstanceId} to {targetProjectId}/{targetInstanceId} Backup {targetBackupId}");
+        Console.WriteLine($"Backup {backup.Name} of size {backup.SizeBytes} bytes was created with encryption keys {string.Join(", ", (object[]) kmsKeyNames)} at {backup.CreateTime} from {backup.Database} and is in state {backup.State} and has version time {backup.VersionTime.ToDateTime()}");
```
jskeet, line 64: Why the cast to `object[]`? I wouldn't expect that to be necessary. (I assume I'm missing something, but I'd like to know what...)

jskeet: I don't think we need to call ToDateTime() either - I believe we'll end up with an ISO-8601-formatted timestamp if you just use `{backup.VersionTime}`. (You're doing that later with CreateTime, so I'd assume it would work the same?)

spanner/api/Spanner.Samples/CreateBackupWithMultiRegionEncryptionAsync.cs
```diff
@@ -0,0 +1,64 @@
...
+        // Poll until the returned long-running operation is complete.
+        var completedResponse = await operation.PollUntilCompletedAsync();
+        if (completedResponse.IsFaulted)
+        {
+            Console.WriteLine($"Error while creating backup: {completedResponse.Exception}");
+            throw completedResponse.Exception;
+        }
+
+        var backup = completedResponse.Result;
+        Console.WriteLine($"Backup {backup.Name} of size {backup.SizeBytes} bytes was created with encryption keys {string.Join(", ", (object[]) kmsKeyNames)} at {backup.CreateTime}");
```
jskeet, line 60: Ditto on the cast here.

spanner/api/Spanner.Samples/CreateDatabaseWithMultiRegionEncryptionAsync.cs
```diff
@@ -0,0 +1,74 @@
...
+public class CreateDatabaseWithMultiRegionEncryptionAsyncSample
+{
+    public async Task<Database> CreateDatabaseWithMultiRegionEncryptionAsync(string projectId, string instanceId, string databaseId, CryptoKeyName[] kmsKeyNames)
+    {
+        // Create a DatabaseAdminClient instance that can be used to execute a
+        // CreateDatabaseRequest with custom encryption configuration options.
+        DatabaseAdminClient databaseAdminClient = DatabaseAdminClient.Create();
+        // Define create table statement for table #1.
+        var createSingersTable =
+        @"CREATE TABLE Singers (
```
jskeet, line 31: Nit: indent this line? (Ditto for createAlbumsTable)

spanner/api/Spanner.Samples/CreateBackupWithMultiRegionEncryptionAsync.cs
jskeet: (It's in later samples too.)

jskeet (approved):
Approving so that you can merge after addressing the comments, which are all nits really.

spanner/api/Spanner.Samples.Tests/CopyBackupWithMultiRegionEncryptionAsyncTest.cs
amanda-tarafa: I've also fixed the code from where this was copied.

spanner/api/Spanner.Samples/CopyBackupWithMultiRegionEncryptionAsync.cs
amanda-tarafa: > Why the cast to object[]? I wouldn't expect that to be necessary.

This is an array of `CryptoKeyNames` so the call to `string.Join` is ambiguos between `object[]` and `IEnumerable<T>`. I'm changing this to be an `IEnumerable<CryptoKeyNames>` instead.

amanda-tarafa: > I don't think we need to call ToDateTime() either

Yes, removed the call to ToDateTime, here and in the cody from where this was copied.

spanner/api/Spanner.Samples/CreateBackupWithMultiRegionEncryptionAsync.cs
amanda-tarafa: Turned the parameter to `IEnumerable<CryptoKeyName>` everywhere. And removed the cast.

spanner/api/Spanner.Samples/CreateDatabaseWithMultiRegionEncryptionAsync.cs
amanda-tarafa: Done here and in all the create database samples.

amanda-tarafa (commented):
@jskeet all comments addressed and fixed elsewhere. Changes in the second commit. PTAL.

spanner/api/Spanner.Samples.Tests/CopyBackupTest.cs
```diff
@@ -38,9 +38,11 @@ public void CopyBackup()
         string target_backupId = SpannerFixture.GenerateId("test_", 16);
         DateTimeOffset expireTime = DateTimeOffset.UtcNow.AddHours(12);
 
-        Backup backup = copyBackupSample.CopyBackup(source_Instance_id, source_Project_id, source_backupId, 
-            target_Instance_id, target_Project_id, target_backupId, expireTime);
+        Backup backup = copyBackupSample.CopyBackup(
```
jskeet, line 41: The variables are still declared above, just not used any more.

jskeet (dismissed):
One simple nit - rest looks good.

spanner/api/Spanner.Samples.Tests/CopyBackupTest.cs
amanda-tarafa: Done, thanks!

jskeet (approved):

PR #2929, opened 2025-02-28
feat(parametermanager): Added samples for creating params

vatsal-vora-crestdata:
## Description
Created samples for creating parameter and parameter version using Parameter Manager SDK

Sample List:

[parameter-structured-data](https://cloud.google.com/secret-manager/parameter-manager/docs/create-parameter#parameter-structured-data)
[parameter-unstructured-data](https://cloud.google.com/secret-manager/parameter-manager/docs/create-parameter#parameter-unstructured-data)
[add-parameter-version](https://cloud.google.com/secret-manager/parameter-manager/docs/add-parameter-version#add-parameter-version) (structured, unstructured)
[create-parameter-secret-reference](https://cloud.google.com/secret-manager/parameter-manager/docs/reference-secrets-in-parameter#create-parameter-secret-reference)

### DIFF
--- a/parametermanager/api/ParameterManager.Samples.Tests/CreateParamTests.cs
+++ b/parametermanager/api/ParameterManager.Samples.Tests/CreateParamTests.cs
@@ -1,0 +1,42 @@
+/*
+ * Copyright 2025 Google LLC
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     https://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+using Google.Cloud.ParameterManager.V1;
+
+[Collection(nameof(ParameterManagerFixture))]
+public class CreateParamTests
+{
+    private readonly ParameterManagerFixture _fixture;
+    private readonly CreateParamSample _sample;
+
+    public CreateParamTests(ParameterManagerFixture fixture)
+    {
+        _fixture = fixture;
+        _sample = new CreateParamSample();
+    }
+
+    [Fact]
+    public void CreateParam()
+    {
+        ParameterName parameterName = _fixture.ParameterName;
+        Parameter result = _sample.CreateParam(
+          projectId: parameterName.ProjectId, parameterId: parameterName.ParameterId);
+
+        Assert.NotNull(result);
+        Assert.Equal(result.Name, parameterName.ToString());
+        Assert.Equal(result.Format.ToString(), ParameterFormat.Unformatted.ToString());
+    }
+}
--- a/parametermanager/api/ParameterManager.Samples.Tests/ParameterManagerFixture.cs
+++ b/parametermanager/api/ParameterManager.Samples.Tests/ParameterManagerFixture.cs
@@ -1,0 +1,255 @@
+// Copyright(c) 2025 Google Inc.
+//
+// Licensed under the Apache License, Version 2.0 (the "License"); you may not
+// use this file except in compliance with the License. You may obtain a copy of
+// the License at
+//
+// http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS, WITHOUT
+// WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied. See the
+// License for the specific language governing permissions and limitations under
+// the License.
+
+using Google.Api.Gax.ResourceNames;
+using Google.Cloud.Iam.V1;
+using Google.Cloud.ParameterManager.V1;
+using Google.Cloud.SecretManager.V1;
+using Google.Protobuf;
+using System.Text;
+
+[CollectionDefinition(nameof(ParameterManagerFixture))]
+public class ParameterManagerFixture : IDisposable, ICollectionFixture<ParameterManagerFixture>
+{
+    public string ProjectId { get; }
+    public string LocationId = "global";
+
+    public string Payload = "test123";
+    public string JsonPayload = "{\"username\": \"test-user\", \"host\": \"localhost\"}";
+    public string SecretReference { get; }
+    public string SecretId = "projects/project-id/secrets/secret-id/versions/latest";
+
+    public ParameterName ParameterName { get; }
+    public ParameterName ParameterNameWithFormat { get; }
+
+    public string ParameterId { get; }
+    public Parameter Parameter { get; }
+    public ParameterVersionName ParameterVersionName { get; }
+
+    public Parameter ParameterWithFormat { get; }
+    public ParameterVersionName ParameterVersionNameWithFormat { get; }
+
+    public Parameter ParameterWithSecretReference { get; }
+    public ParameterVersionName ParameterVersionNameWithSecretReference { get; }
+
+    public Parameter ParameterToDelete { get; }
+    public ParameterName ParameterNameToDelete { get; }
+
+    public string ParameterVersionId { get; }
+    public Parameter ParameterToDeleteVersion { get; }
+    public ParameterVersion ParameterVersionToDelete { get; }
+    public ParameterVersionName ParameterVersionNameToDelete { get; }
+
+    public ParameterName ParameterNameForQuickstart { get; }
+    public ParameterVersionName ParameterVersionNameForQuickstart { get; }
+
+    public Parameter ParameterToRender { get; }
+    public ParameterVersion ParameterVersionToRender { get; }
+    public ParameterVersionName ParameterVersionNameToRender { get; }
+    public Secret Secret { get; }
+    public SecretVersion SecretVersion { get; }
+    public Policy Policy { get; }
+    public ParameterManagerFixture()
+    {
+        ProjectId = Environment.GetEnvironmentVariable("GOOGLE_PROJECT_ID");
+        if (String.IsNullOrEmpty(ProjectId))
+        {
+            throw new Exception("missing GOOGLE_PROJECT_ID");
+        }
+
+        ParameterName = new ParameterName(ProjectId, LocationId, RandomId());
+        ParameterNameWithFormat = new ParameterName(ProjectId, LocationId, RandomId());
+
+        ParameterId = RandomId();
+        Parameter = CreateParameter(ParameterId, ParameterFormat.Unformatted);
+        ParameterVersionName = new ParameterVersionName(ProjectId, LocationId, ParameterId, RandomId());
+
+        ParameterId = RandomId();
+        ParameterWithFormat = CreateParameter(ParameterId, ParameterFormat.Json);
+        ParameterVersionNameWithFormat = new ParameterVersionName(ProjectId, LocationId, ParameterId, RandomId());
+
+        ParameterId = RandomId();
+        ParameterWithSecretReference = CreateParameter(ParameterId, ParameterFormat.Unformatted);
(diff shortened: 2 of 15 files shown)


amanda-tarafa:
Thanks for separating the PRs. I'll wait for #2921 to be fixed and for these PRs to each be moved out of draft.

arpangoswami:
Since 2921 is merged, Can we take a look at the following PRs?
#2929
#2930 
#2931 
#2932
#2933
#2934
#2935

arpangoswami (approved):

amanda-tarafa:
Yes, I'll review these one by one.

amanda-tarafa (changes requested):
The tests are not running properly because the project is targetting .NET 9.

We target the latest LTS .NET version, which is .NET 8. Please change to that. I'll run the tests again after.

vatsal-vora-crestdata:
@amanda-tarafa I have updated the .Net version to 8 in all the PRs.

amanda-tarafa:
These are now blocked by b/405387323.

arpangoswami (approved):

parametermanager/api/ParameterManager.Samples/ParameterManager.Samples.csproj
```diff
@@ -0,0 +1,15 @@
...
+
+  <PropertyGroup>
+    <TargetFramework>net8.0</TargetFramework>
+    <ImplicitUsings>enable</ImplicitUsings>
+    <Nullable>enable</Nullable>
+    <IsPackable>false</IsPackable>
+  </PropertyGroup>
+
+  <ItemGroup>
+    <PackageReference Include="Google.Cloud.Common" Version="2.2.0" />
```
amanda-tarafa, line 11: Do you really need this?

parametermanager/api/ParameterManager.Samples/CreateParam.cs
```diff
@@ -0,0 +1,53 @@
...
+using Google.Api.Gax.ResourceNames;
+using Google.Cloud.ParameterManager.V1;
+
+/// <summary>
+/// This function creates a parameter of the format type "unformatted" using the Parameter Manager SDK for GCP.
+/// </summary>
+/// <param name="projectId">The ID of the project where the parameter is to be created.</param>
+/// <param name="parameterId">The ID to assign to the new parameter. This ID must be unique within the project.</param>
+/// <returns>The created Parameter object.</returns>
+public class CreateParamSample
```
amanda-tarafa, line 28: Can you use parameter through instead of param?

parametermanager/api/ParameterManager.Samples/CreateParam.cs
```diff
@@ -0,0 +1,53 @@
...
+
+using Google.Api.Gax.ResourceNames;
+using Google.Cloud.ParameterManager.V1;
+
+/// <summary>
+/// This function creates a parameter of the format type "unformatted" using the Parameter Manager SDK for GCP.
+/// </summary>
+/// <param name="projectId">The ID of the project where the parameter is to be created.</param>
+/// <param name="parameterId">The ID to assign to the new parameter. This ID must be unique within the project.</param>
+/// <returns>The created Parameter object.</returns>
```
amanda-tarafa, line 27: This should be in the method and not on the class.

parametermanager/api/ParameterManager.Samples.Tests/ParameterManagerFixture.cs
```diff
@@ -0,0 +1,255 @@
...
+using Google.Protobuf;
+using System.Text;
+
+[CollectionDefinition(nameof(ParameterManagerFixture))]
+public class ParameterManagerFixture : IDisposable, ICollectionFixture<ParameterManagerFixture>
+{
+    public string ProjectId { get; }
+    public string LocationId = "global";
+
+    public string Payload = "test123";
```
amanda-tarafa, line 28: These should be constants.

parametermanager/api/ParameterManager.Samples.Tests/ParameterManagerFixture.cs
```diff
@@ -0,0 +1,255 @@
...
+
+    public ParameterName ParameterNameForQuickstart { get; }
+    public ParameterVersionName ParameterVersionNameForQuickstart { get; }
+
+    public Parameter ParameterToRender { get; }
+    public ParameterVersion ParameterVersionToRender { get; }
+    public ParameterVersionName ParameterVersionNameToRender { get; }
+    public Secret Secret { get; }
+    public SecretVersion SecretVersion { get; }
+    public Policy Policy { get; }
```
amanda-tarafa, line 62: Are all of these being shared by two or more tests? As a general rule everything that's test specific should be on the test itself, the fixture should only contain data that's going to be sherad across tests and helper methods.

parametermanager/api/ParameterManager.Samples.Tests/CreateParamTests.cs
```diff
@@ -0,0 +1,42 @@
...
+
+    [Fact]
+    public void CreateParam()
+    {
+        ParameterName parameterName = _fixture.ParameterName;
+        Parameter result = _sample.CreateParam(
+          projectId: parameterName.ProjectId, parameterId: parameterName.ParameterId);
+
+        Assert.NotNull(result);
+        Assert.Equal(result.Name, parameterName.ToString());
```
amanda-tarafa, line 39: There should be a `NameAsParameterName` property that allows you to compare without converting to string.

parametermanager/api/ParameterManager.Samples.Tests/CreateParamTests.cs
```diff
@@ -0,0 +1,42 @@
...
+        Assert.Equal(result.Format.ToString(), ParameterFormat.Unformatted.ToString());
```
amanda-tarafa, line 40: You shouldn't need to convert these to string.

parametermanager/api/ParameterManager.Samples.Tests/ParameterManagerFixture.cs
```diff
@@ -0,0 +1,255 @@
...
+using Google.Cloud.ParameterManager.V1;
+using Google.Cloud.SecretManager.V1;
+using Google.Protobuf;
+using System.Text;
+
+[CollectionDefinition(nameof(ParameterManagerFixture))]
+public class ParameterManagerFixture : IDisposable, ICollectionFixture<ParameterManagerFixture>
+{
+    public string ProjectId { get; }
+    public string LocationId = "global";
```
amanda-tarafa, line 26: You are hardcoding "global" in the samples already, which means you don't need this here.

amanda-tarafa (changes requested):
I've reviewed some of the files and left comments. Please apply those changes to all files in this PR and in the rest.

Also, let's do the fixture incrementally on each of the PRs, that is, on each PR only add to the fixture what's needed for that PR.

parametermanager/api/ParameterManager.Samples/ParameterManager.Samples.csproj
vatsal-vora-crestdata: No, not needed. Removed this package reference.

parametermanager/api/ParameterManager.Samples/CreateParam.cs
vatsal-vora-crestdata: Updated the docstring.

parametermanager/api/ParameterManager.Samples.Tests/ParameterManagerFixture.cs
vatsal-vora-crestdata: Removed additional data that is not used in this PR's test cases.

parametermanager/api/ParameterManager.Samples.Tests/CreateParamTests.cs
vatsal-vora-crestdata: I've updated assetions.

parametermanager/api/ParameterManager.Samples.Tests/CreateParamTests.cs
vatsal-vora-crestdata: Removed ToString().

parametermanager/api/ParameterManager.Samples.Tests/ParameterManagerFixture.cs
vatsal-vora-crestdata: Yes, in samples, `global` is hardcoded, but this variable is used for creating and deleting parameters and parameter versions in fixtures.

parametermanager/api/ParameterManager.Samples.Tests/ParameterManagerFixture.cs
vatsal-vora-crestdata: Not using `const` for these fields because they belong to an instance class (`ParameterManagerFixture`) and need to be accessed through an instance (`_fixture`). Making them `const` would require the class to be `static`, which is not desirable in this case.

arpangoswami (dismissed):

parametermanager/api/ParameterManager.Samples.Tests/ParameterManagerFixture.cs
amanda-tarafa: Can't you parameterize the location in samples instead and always pass global or something else from tests?

parametermanager/api/ParameterManager.Samples.Tests/ParameterManagerFixture.cs
amanda-tarafa: > Making them const would require the class to be static

This is not true, consts can be defined in non static classes. See https://learn.microsoft.com/en-us/dotnet/csharp/programming-guide/classes-and-structs/constants

> Not using const for these fields because they belong to an instance class (ParameterManagerFixture) and need to be accessed through an instance (_fixture).

There's absoluteyl nothing wrong with accessing consts through the class ParameterManagerFixture.

amanda-tarafa (changes requested):

parametermanager/api/ParameterManager.Samples.Tests/ParameterManagerFixture.cs
vatsal-vora-crestdata: For global parameter samples, a location is not required. However, since `ParameterName(ProjectId, LocationId, RandomId())` requires a location, `global` must be explicitly provided. Therefore, it has been declared.

parametermanager/api/ParameterManager.Samples.Tests/ParameterManagerFixture.cs
vatsal-vora-crestdata: Thank you for pointing that out. You're absolutely correct—const fields can be declared within non-static classes and accessed using the class name. The issue stemmed from referencing the constants through an instance, which led to the error. This has been corrected by updating the variables to constants and referencing them appropriately.

parametermanager/api/ParameterManager.Samples.Tests/ParameterManagerFixture.cs
```diff
@@ -0,0 +1,124 @@
...
+
+    public string ParameterId { get; }
+    public Parameter Parameter { get; }
+    public ParameterVersionName ParameterVersionName { get; }
+
+    public Parameter ParameterWithFormat { get; }
+    public ParameterVersionName ParameterVersionNameWithFormat { get; }
+
+    public Parameter ParameterWithSecretReference { get; }
+    public ParameterVersionName ParameterVersionNameWithSecretReference { get; }
```
amanda-tarafa, line 43: All of these seem to be used in a single test each. If that's the case, they shouldn't be on the fixture, they should be on the specific test that uses them. In the fixture, you should only have data that you share across tests (and that is not modified by tests) and helper methods.

For cleanup, you can have a collection of `_parametersToDelete`, etc. that you iterate over on dispose to delete the parameters. Each test would call the helper methods it needs, for instance  `CreateParameters`, and it's those helper methods the ones that add paremeter names to that collection.

durgesh-ninave-crest:
@amanda-tarafa Raised a duplicate of this PR https://github.com/GoogleCloudPlatform/dotnet-docs-samples/pull/2981 because Vatsal's CLA has expired, which is blocking the merge. Once the new PR is merged, will close the original one.

amanda-tarafa:
Closing in favor of #2981

PR #2930, opened 2025-02-28
feat(parametermanager): Added quickstart samples for both global and regional

vatsal-vora-crestdata:
## Description
Created quickstart samples for global and regional parameter using Parameter Manager SDK

Sample List:
1. Global Quickstart
2. Regional Quickstart

Also added README file and Tests for the same.

### DIFF
--- a/parametermanager/README.md
+++ b/parametermanager/README.md
@@ -1,0 +1,1 @@
+### Initial placeholder README file for folder creation
--- a/parametermanager/README.md
+++ b/parametermanager/README.md
@@ -1,1 +1,0 @@
-### Initial placeholder README file for folder creation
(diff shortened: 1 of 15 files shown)


arpangoswami (dismissed):

arpangoswami (dismissed):

arpangoswami (approved):

durgesh-ninave-crest:
@amanda-tarafa Raised a duplicate of this PR https://github.com/GoogleCloudPlatform/dotnet-docs-samples/pull/2986 because Vatsal's CLA has expired, which is blocking the merge. Once the new PR is merged, will close the original one.

amanda-tarafa:
@durgesh-ninave-crest Please close all the old PRs now, so we don't have to sort through each of the many PRs to figure out which we need to review and which we don't need to review.

Adittionally, it's probably better to just have one or two small PRs open at any given moment, as it's likely that change requests in those will need to be applied to the rest. Also we won't have this problem of having to duplicate PRs if licensenses expire or similar. Please conside doing that now. Thanks.

PR #2994, opened 2025-04-11
feat(parametermanager): Added samples for kms_key field in global parameter

durgesh-ninave-crest:
## Description

Added samples for global parameter manager kms_key field using Parameter manager SDK

#### Samples List:

1. create_param_with_kms_key
2. remove_param_kms_key
3. update_param_kms_key

Also added tests for the same.


parametermanager/api/ParameterManager.Samples.Tests/ParameterManagerFixture.cs
```diff
@@ -115,6 +139,68 @@ private void DeleteParameter(ParameterName name)
         }
     }
 
+    public KeyRing CreateKeyRing(string projectId, string keyRingId)
```
amanda-tarafa, line 142: Add to some delete list?

parametermanager/api/ParameterManager.Samples.Tests/ParameterManagerFixture.cs
```diff
@@ -115,6 +139,68 @@ private void DeleteParameter(ParameterName name)
...
+        {
+            return KmsClient.CreateKeyRing(new CreateKeyRingRequest
+            {
+                ParentAsLocationName = parent,
+                KeyRingId = keyRingId,
+            });
+        }
+    }
+
+    public CryptoKey CreateHsmKey(string projectId, string keyId, string keyRingId)
```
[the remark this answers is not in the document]
amanda-tarafa: This one I think really needs to be cleaned.

parametermanager/api/ParameterManager.Samples.Tests/ParameterManagerFixture.cs
amanda-tarafa: OK, I think I see what you are doing. You want to have a permanent key ring for these tests and just create it if it's not there. What's the reason for that?

I think it would be less flake prone if you create a key ring with a random name and fxiture initializarion and dispose of it at the end. It's OK to share the key ring across a single test run but I don't think we should have a permanent key ring that is used by all tests run (those may run in parallel on CI, when PRs are submitted, etc.)

amanda-tarafa (changes requested):

parametermanager/api/ParameterManager.Samples.Tests/ParameterManagerFixture.cs
durgesh-ninave-crest: You're right that using a random name for the key ring would reduce flakiness. However, in Google Cloud, key rings cannot be deleted once created. This means that if we create a new key ring with a random name each time, they will accumulate over time, since we can't clean them up. We can only destroy key versions, not the key ring itself.

amanda-tarafa: Acked. Then let's remove the code that creates the KeyRing. I can create that resource statically and you can reference it directly. You can even add a check on fixture initialization, and if the key ring with the expected name does not exist fail with a descriptive error. What do you think?

durgesh-ninave-crest: Yes, that’s a good idea. I'll remove the KeyRing creation code and instead reference the expected statically-created resource. I'll also add a check during fixture initialisation to fail with a clear error if the KeyRing doesn't exist.

gptSanyam (approved):

gptSanyam (approved):

gptSanyam (approved):

parametermanager/api/ParameterManager.Samples.Tests/ParameterManagerFixture.cs
durgesh-ninave-crest: @amanda-tarafa I've updated the testcases and fixture. Please create a keyring named `csharp-test-key-ring`.

parametermanager/api/ParameterManager.Samples.Tests/CreateParameterWithKmsKeyTests.cs
```diff
@@ -0,0 +1,53 @@
...
+using Google.Cloud.ParameterManager.V1;
+using Google.Protobuf;
+using System.Text;
+
+[Collection(nameof(ParameterManagerFixture))]
+public class CreateParameterWithKmsKeyTests
+{
+    private readonly ParameterManagerFixture _fixture;
+    private readonly CreateParameterWithKmsKeySample _sample;
+    readonly string keyRingId = "csharp-test-key-ring";
```
amanda-tarafa, line 27: Make this a property on the fixture, so you don't have to repeat it everywhere.

parametermanager/api/ParameterManager.Samples.Tests/CreateParameterWithKmsKeyTests.cs
```diff
@@ -0,0 +1,53 @@
...
+
+    public CreateParameterWithKmsKeyTests(ParameterManagerFixture fixture)
+    {
+        _fixture = fixture;
+        _sample = new CreateParameterWithKmsKeySample();
+        _fixture.GetKeyRing(_fixture.ProjectId, keyRingId);
```
amanda-tarafa, line 33: Why do you need to call GetKeyRing here?

parametermanager/api/ParameterManager.Samples.Tests/CreateParameterWithKmsKeyTests.cs
```diff
@@ -0,0 +1,53 @@
...
+    }
+
+    [Fact]
+    public void CreateParameterWithKmsKey()
+    {
+        string KeyId = _fixture.RandomId();
+        CryptoKey cryptoKey = _fixture.CreateHsmKey(_fixture.ProjectId, KeyId, keyRingId);
+        CryptoKeyVersionName cryptoKeyVersionName = new CryptoKeyVersionName(_fixture.ProjectId, ParameterManagerFixture.LocationId, keyRingId, KeyId, "1");
+        _fixture.CryptoKeyVersionsToDelete.Add(cryptoKeyVersionName);
```
amanda-tarafa, line 42: Can you include these two lines within the CreateHsmKey method in the fixture?

parametermanager/api/ParameterManager.Samples/RemoveParameterKmsKey.cs
```diff
@@ -0,0 +1,64 @@
...
+
+        // Build the parent resource name. 
+        ParameterName name = new ParameterName(projectId, "global", parameterId);
+
+        // Build the parameter.
+        UpdateParameterRequest request = new UpdateParameterRequest
+        {
+            Parameter = new Parameter
+            {
+                Name = name.ToString(),
```
amanda-tarafa, line 46: There should be a field here named `NameAsParameterName`, if you use that you don't have to call ToString.

parametermanager/api/ParameterManager.Samples.Tests/ParameterManagerFixture.cs
```diff
@@ -73,6 +81,21 @@ public string RandomId()
         return (parameterId, versionId, parameter, payload);
```
amanda-tarafa, line 81: I know this is not added on this PR, but what is this method for, `CreateParameterWithVersion`? It's really not doing that, it's just creating a parameter and returning some versionId and payload. Can you delete this method?

parametermanager/api/ParameterManager.Samples.Tests/ParameterManagerFixture.cs
```diff
@@ -115,6 +138,63 @@ private void DeleteParameter(ParameterName name)
         }
     }
 
+    public KeyRing GetKeyRing(string projectId, string keyRingId)
```
amanda-tarafa, line 141: I don't think we need this method, nor to call it anywhere.

parametermanager/api/ParameterManager.Samples.Tests/ParameterManagerFixture.cs
```diff
@@ -115,6 +138,63 @@ private void DeleteParameter(ParameterName name)
...
+                    Algorithm = CryptoKeyVersion.Types.CryptoKeyVersionAlgorithm.GoogleSymmetricEncryption,
+                    ProtectionLevel = ProtectionLevel.Hsm,
+                },
+            },
+        };
+
+        try
+        {
+            string name = $"projects/{projectId}/locations/global/keyRings/{keyRingId}/cryptoKeys/{keyId}";
+            CryptoKey resp = KmsClient.GetCryptoKey(name);
```
amanda-tarafa, line 177: Why do we need to get the key first. Aren't we creating these as we need them and deleting them on dispose? Or is this a similar case as the key ring? If it is a similar case as the key ring let's address it in the same way we addressed the key ring case.

amanda-tarafa (changes requested):
The change requests I've made on a couple of the files apply to all others, where relevant.

Thanks.

parametermanager/api/ParameterManager.Samples.Tests/CreateParameterWithKmsKeyTests.cs
durgesh-ninave-crest: Updated the property in the fixture.

parametermanager/api/ParameterManager.Samples.Tests/CreateParameterWithKmsKeyTests.cs
durgesh-ninave-crest: In the KMS tests, we check whether the key ring is present because If we perform this check in the fixture and the key ring is not present, then all the test cases are skipped.

parametermanager/api/ParameterManager.Samples/RemoveParameterKmsKey.cs
durgesh-ninave-crest: I've removed ToString.

parametermanager/api/ParameterManager.Samples.Tests/ParameterManagerFixture.cs
durgesh-ninave-crest: This method is called in the KMS-related test cases.

amanda-tarafa (approved):

PR #3004, opened 2025-04-17
samples(storage): add samples and test cases to support MoveObject feature for HNS-enabled buckets

mahendra-google:


storage/api/Storage.Samples.Tests/MoveObjectWithMetaGenerationMismatchFailsTest.cs
```diff
@@ -0,0 +1,43 @@
...
+    [Fact]
+    public void MoveObjectWithMetaGenerationMismatchFails()
+    {
+        MoveObjectMetaGenerationMismatchFailsSample moveObjectMetaGenerationMismatchFailsSample = new MoveObjectMetaGenerationMismatchFailsSample();
+        UploadObjectFromMemorySample uploadObjectFromMemory = new UploadObjectFromMemorySample();
+        var originName = Guid.NewGuid().ToString();
+        var originContent = Guid.NewGuid().ToString();
+        var destinationName = Guid.NewGuid().ToString();
+        uploadObjectFromMemory.UploadObjectFromMemory(_fixture.BucketNameHns, originName, originContent);
+        var exception = Assert.Throws<GoogleApiException>(() => moveObjectMetaGenerationMismatchFailsSample.MoveObjectMetaGenerationMismatchFails(_fixture.BucketNameHns, originName, destinationName, 0));
```
amanda-tarafa, line 39: This could just as well be testing that the sample succeeds with the right precondition. The sample does not intrinsically fail, so it is unsual to test for a failure here. If we really need to keep the sample I'd prefer that we test the sample succeeds with the right preconditions.

amanda-tarafa (changes requested):
These look good, thanks. The change requests are tiny but I do have a question about the purpose of the `...Fails` sample. @danielduhh might be able to clarify.

storage/api/Storage.Samples.Tests/MoveObjectWithMetaGenerationMismatchFailsTest.cs
mahendra-google: My understanding was that we can create separate samples for failure cases and test those as well but I agree with you that the sample succeeds with the right preconditions will suffice.

mahendra-google: So in that case  I think I should remove failed samples and test cases ?

amanda-tarafa: That would be my preference yes, and then possibly on the documentation where the sample is included, a note explaining that moving an object with missmatched preconditions fails.
But please coordinate/confirm with technical writers. In this repo we should only have samples that are included on documentation.

mahendra-google: As I am not aware , who will be technical writers to co-ordinate with in this case ?

amanda-tarafa: I'll send an internal email.

mahendra-google: All samples other than storage_move_object are removed

storage/api/Storage.Samples/MoveObjectWithGenerationMismatchFails.cs
```diff
@@ -0,0 +1,42 @@
...
+public class MoveObjectGenerationMismatchFailsSample
+{
+    /// <summary>
+    /// Moves the source object to the destination object within a hierarchical namespace enabled bucket with preconditions set.
+    /// </summary>
+    /// <param name="sourceBucketName">Name of the source bucket containing the object to move.</param>
+    /// <param name="sourceObjectName">The name of the source object to move within the bucket.</param>
+    /// <param name="destinationObjectName">The name of the new object to move to within the bucket.</param>
+    /// <param name="generation">The generation of the source object.</param>
+    public void MoveObjectGenerationMismatchFails(
```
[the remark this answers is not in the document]
mahendra-google: All samples other than storage_move_object are removed as discussed in Google Chat Space with @danielduhh

mahendra-google:
This PR is ready for re-review

amanda-tarafa (approved):
Thanks! I'll merge when tests are green.

PR #3027, opened 2025-05-05
feat(PubSub): Add SubscriberErrorListener Sample

cy-yun:
PubSub SubscriberErrorListener Sample

https://cloud.google.com/pubsub/docs/samples/pubsub-subscriber-error-listener#pubsub_subscriber_error_listener-java

### DIFF
--- a/pubsub/api/Pubsub.Samples.Tests/SubscriberErrorListenerTest.cs
+++ b/pubsub/api/Pubsub.Samples.Tests/SubscriberErrorListenerTest.cs
@@ -1,0 +1,48 @@
+// Copyright 2025 Google LLC
+//
+// Licensed under the Apache License, Version 2.0 (the "License").
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at 
+//
+// https://www.apache.org/licenses/LICENSE-2.0 
+//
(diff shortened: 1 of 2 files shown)


hongalex (approved):

pubsub/api/Pubsub.Samples.Tests/SubscriberErrorListenerTest.cs
```diff
@@ -0,0 +1,48 @@
...
+        _pubsubFixture = pubsubFixture;
+        _publishMessagesAsyncSample = new PublishMessagesAsyncSample();
+        _subscriberErrorListenerSample = new SubscriberErrorListenerSample();
+    }
+
+    [Fact]
+    public async Task SubscribeWithErrorListener()
+    {
+        var (topicId, subscriptionId) = _pubsubFixture.RandomNameTopicSubscriptionId();
+        var messages = Enumerable.Range(0, 100).Select(index => $"Message {index}").ToList();
```
amanda-tarafa, line 37: Same note as before, I don't think we need to test with 100 messages. This could lead to the test being flaky? Testing with a single message should be enough.

amanda-tarafa (changes requested):

amanda-tarafa (approved):

PR #3028, opened 2025-05-06
feat(PubSub): Add OptimisticSubscribe Sample

cy-yun:
https://cloud.google.com/pubsub/docs/samples/pubsub-optimistic-subscribe

### DIFF
--- a/pubsub/api/Pubsub.Samples.Tests/OptimisticSubscribeTest.cs
+++ b/pubsub/api/Pubsub.Samples.Tests/OptimisticSubscribeTest.cs
@@ -1,0 +1,47 @@
+// Copyright 2025 Google LLC
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+// http://www.apache.org/licenses/LICENSE-2.0
+//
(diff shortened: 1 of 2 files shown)


hongalex (dismissed):

hongalex (dismissed):

pubsub/api/Pubsub.Samples.Tests/OptimisticSubscribeTest.cs
```diff
@@ -0,0 +1,47 @@
...
+        _pubsubFixture = pubsubFixture;
+        _publishMessagesAsyncSample = new PublishMessagesAsyncSample();
+        _optimisticSubscribeSample = new OptimisticSubscribeSample();
+    }
+
+    [Fact]
+    public async Task SubscribeOptimistically()
+    {
+        var (topicId, subscriptionId) = _pubsubFixture.RandomNameTopicSubscriptionId();
+        var messages = Enumerable.Range(0, 100).Select(index => $"Message {index}").ToList();
```
amanda-tarafa, line 37: Let's publish a single message and expect a single message to be read. 5 seconds might be too little to recieve 100 messages, and any way, if we recieve one, we know the sample is right, testing with more than 1 adds nothing.

amanda-tarafa (changes requested):

PR #3032, opened 2025-05-07
samples(storage): add samples and test cases for soft delete objects

mahendra-google:
Add samples and test cases for soft delete objects as follows:-

- [ ] storage_list_soft_deleted_objects
- [ ] storage_list_soft_deleted_object_versions
- [ ] storage_restore_object

### DIFF
--- a/storage/api/Storage.Samples/ListSoftDeletedObjects.cs
+++ b/storage/api/Storage.Samples/ListSoftDeletedObjects.cs
@@ -1,0 +1,39 @@
+// Copyright 2025 Google Inc.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+// [START storage_list_soft_deleted_objects]
+
+using Google.Cloud.Storage.V1;
+using System;
+using System.Collections.Generic;
+
+public class ListSoftDeletedObjectsSample
+{
+    /// <summary>
+    /// List soft deleted objects.
+    /// </summary>
+    /// <param name="bucketName">The name of the bucket.</param>
+    public IEnumerable<Google.Apis.Storage.v1.Data.Object> ListSoftDeletedObjects(string bucketName = "your-unique-bucket-name")
+    {
+        var storage = StorageClient.Create();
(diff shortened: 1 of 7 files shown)


mahendra-google:
This PR is ready for review.

storage/api/Storage.Samples/ListSoftDeletedObjects.cs
```diff
@@ -0,0 +1,39 @@
...
+public class ListSoftDeletedObjectsSample
+{
+    /// <summary>
+    /// List all soft-deleted objects in the bucket.
+    /// </summary>
+    /// <param name="bucketName">The name of the bucket.</param>
+    public IEnumerable<Google.Apis.Storage.v1.Data.Object> ListSoftDeletedObjects(string bucketName = "your-unique-bucket-name")
+    {
+        var storage = StorageClient.Create();
+        var objects = storage.ListObjects(bucketName, null, new ListObjectsOptions { SoftDeletedOnly = true });
```
amanda-tarafa, line 30: Not sure what the null parameter is, but please include its name to make the sample clearer, something like (we don't need to do this for parameters that we pass values to, because the values make it clear enough):
```suggestion
        var objects = storage.ListObjects(bucketName, parameterName: null, new ListObjectsOptions { SoftDeletedOnly = true });
```

storage/api/Storage.Samples/ListSoftDeletedVersionsOfObject.cs
```diff
@@ -0,0 +1,41 @@
...
+    /// <summary>
+    /// List all soft-deleted versions of the object in the bucket.
+    /// </summary>
+    /// <param name="bucketName">The name of the bucket.</param>
+    /// <param name="objectName">The name of the object.</param>
+    public IEnumerable<Google.Apis.Storage.v1.Data.Object> ListSoftDeletedVersionsOfObject(string bucketName = "your-unique-bucket-name",
+        string objectName = "your-object-name")
+    {
+        var storage = StorageClient.Create();
+        var objects = storage.ListObjects(bucketName, null, new ListObjectsOptions { SoftDeletedOnly = true, MatchGlob = objectName });
```
amanda-tarafa, line 32: Same commet as before regarding including the parameter name.

amanda-tarafa (changes requested):
A small change request but otherwise looks good. Thanks.

storage/api/Storage.Samples/ListSoftDeletedObjects.cs
mahendra-google: Changes are made

storage/api/Storage.Samples/ListSoftDeletedVersionsOfObject.cs
mahendra-google: Changes are made

mahendra-google:
This PR is ready for re-review

amanda-tarafa (approved):
Thanks, I'll wait for tests to be green before merging.

PR #3106, opened 2025-06-10
fix(retail): add region tag for Dotnet - SearchSimpleQuery sample

rebeccaellis:
Added region tags in order to update code samples in the retail commerce documentation.


amanda-tarafa:
This was a duplicate of #3110 . #3110 has now being updated to address my change request.

PR #3207, opened 2025-09-16
samples(Storagecontrol): Add samples for anywhere cache

mahendra-google:
This PR adds sample code for the anywhere cache feature, allowing users to:

- [ ] Create anywhere cache
- [ ] Get anywhere cache
- [ ] List anywhere cache
- [ ] Update anywhere cache
- [ ] Pause anywhere cache
- [ ] Disable anywhere cache
- [ ] Resume anywhere cache

> [!NOTE]
> Sample tests are not included because of the long operation time.

### DIFF
--- a/storage/api/Storage.Samples.Tests/PrintFileAclForUserTest.cs
+++ b/storage/api/Storage.Samples.Tests/PrintFileAclForUserTest.cs
@@ -1,4 +1,4 @@
-﻿// Copyright 2020 Google Inc.
+// Copyright 2020 Google Inc.
 //
 // Licensed under the Apache License, Version 2.0 (the "License");
 // you may not use this file except in compliance with the License.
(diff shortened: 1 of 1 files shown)


storage/api/Storage.Samples/StorageControlListAnywhereCaches.cs
```diff
@@ -0,0 +1,50 @@
...
+    public IEnumerable<AnywhereCache> StorageControlListAnywhereCaches(string bucketName = "your-unique-bucket-name")
+    {
+        StorageControlClient storageControlClient = StorageControlClient.Create();
+
+        // Set project to "_" to signify globally scoped bucket.
+        string parent = $"projects/_/buckets/{bucketName}";
+
+        var request = new ListAnywhereCachesRequest
+        {
+            Parent = parent
```
krishnamd-jkp, line 35: Should we also add PageSize parameter here so users are educated about this field also?

mahendra-google: Yes, We should add PageSize parameter so that the user will be educated to control the  number of Caches returned in the list Cache ( i.e either pageSize number or 1000 whichever is smaller).

mahendra-google:
Hi @krishnamd-jkp, I've successfully synced the PR with the new code ownership rules. The **@GoogleCloudPlatform/gcs-sdk-team** is now assigned. Ready for your review whenever you have a moment

cpriti-os (approved):

PR #3273, opened 2025-11-17
feat(genai): add text generation samples (3)

jdomingr:
Add new samples for text generation:

- Text generation with system instructions
- Text generation with mute video
- Text generation with text and image input
- Text generation with streaming responses

### DIFF
--- a/genai/api/GenAI.Samples/TextGeneration/TextGenWithTxtStream.cs
+++ b/genai/api/GenAI.Samples/TextGeneration/TextGenWithTxtStream.cs
@@ -45,7 +45,7 @@ public class TextGenWithTxtStream
             Console.WriteLine(text);
             fullResponse.Append(text);
         }
-        // Example reponse:
+        // Example response:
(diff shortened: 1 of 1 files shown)


genai/api/GenAI.Samples/TextGeneration/TextGenSysInstrWithTxt.cs
```diff
@@ -0,0 +1,59 @@
...
+            project: projectId,
+            location: location,
+            vertexAI: true,
+            httpOptions: new HttpOptions { ApiVersion = "v1" });
+
+        var systemInstruction = new Content
+        {
+            Parts = new List<Part>
+            {
+                new Part { Text = "You're a language translator." },
```
yyyu-google: why do we need two part objects here? we only need one part

yyyu-google:
look good to me after addressing the system instruction comment

gericdong (approved):

PR #3274, opened 2025-11-20
feat(genai): add text generation samples (4)

jdomingr:
Add new samples for text generation:

- Text generation with local video.
- Text generation with multiple local images.
- Text generation with multiple images (local image and Gcs image)


genai/api/GenAI.Samples/TextGeneration/TextGenWithMultiLocalImg.cs
```diff
@@ -0,0 +1,72 @@
+// Copyright 2024 Google LLC
```
NimJay: Is `2024` intentional here? If not, consider correcting is a later/future PR.

NimJay (commented):
Looks good. Thank you for adding tests.
Approved. Merging.

Aside: The "Google GitHub Admin: Actions Workflow Security Scan" check is stuck on "Expected — Waiting for workflow to run". But docs about this check state that this check only scans PRs that modify `.github/workflows/` or `.github/actions/` — which this PR doesn't. Safe to merge without awaiting this check.

PR #3275, opened 2025-11-21
feat(genai): add thinking samples

jdomingr:
Add samples for thinking:

- Thinking with text
- Thinking budget with text
- Thinking including thoughts

### DIFF
--- a/genai/api/GenAI.Samples/Thinking/ThinkingBudgetWithTxt.cs
+++ b/genai/api/GenAI.Samples/Thinking/ThinkingBudgetWithTxt.cs
@@ -52,10 +52,6 @@ public class ThinkingBudgetWithTxt
         Console.WriteLine($"Token count for thinking: {response.UsageMetadata.ThoughtsTokenCount}");
         // Example response:
         // Token count for thinking: 804
-
-        Console.WriteLine($"Total token count: {response.UsageMetadata.TotalTokenCount}");
-        // Example response:
-        // Total token count: 1450
         return responseText;
     }
 }


gericdong (approved):

genai/api/GenAI.Samples/Thinking/ThinkingBudgetWithTxt.cs
```diff
@@ -0,0 +1,62 @@
...
+        // It follows the pattern $a ^ 2 + 2ab + b ^ 2 = (a + b) ^ 2$.
+        // ...
+        // All methods yield the same solution.
+        // The solution is $x = -2$.
+
+        Console.WriteLine($"Token count for thinking: {response.UsageMetadata.ThoughtsTokenCount}");
+        // Example response:
+        // Token count for thinking: 804
+
+        Console.WriteLine($"Total token count: {response.UsageMetadata.TotalTokenCount}");
```
yyyu-google: I think printing out thinking token count is reasonable, but we don't need to print total token count. printing total token counts confuses the intend of the sample to users. We have dedicated CountTokensAsync https://github.com/googleapis/dotnet-genai/blob/a3b8c21a2bb54de8707e4919834a9ef40984f574/Google.GenAI/Models.cs#L4130 , so can create a separate sample to demonstrate this feature to users.

genai/api/GenAI.Samples/Thinking/ThinkingIncludeThoughtsWithTxt.cs
```diff
@@ -0,0 +1,65 @@
...
+            config: new GenerateContentConfig
+            {
+                ThinkingConfig = new ThinkingConfig { IncludeThoughts = true }
+            });
+
+        foreach (Part part in response.Candidates[0].Content.Parts)
+        {
+            if (part.Thought.GetValueOrDefault(false))
+            {
+                Console.WriteLine(part.Text);
```
yyyu-google: same comment here, I'd suggest to print out the thinking signature in the response.

yyyu-google (approved):

genai/api/GenAI.Samples/Thinking/ThinkingIncludeThoughtsWithTxt.cs
```diff
@@ -0,0 +1,65 @@
...
+        // 2. **Coefficients in Hand: **  `a` is 1, `b` is 4, and `c` is also 4.Easy enough.
+        // 3. **Method Selection - The Decision Point:**Now, I've got to choose the most efficient path.
+        // I could *always* use the quadratic formula, but that might be overkill. Factoring is always worth a
+        // look, especially when dealing with smaller coefficients. Completing the square is a possibility, but
+        // less elegant here. Graphing... well, that's more for visualization, not direct algebraic solution.
+        // 4. **Evaluate for *This * Case:**Let's see if factoring holds up.
+        // ...
+        // * End with a concise summary of the key takeaway, emphasizing that it's a repeated root.
+        // Perfect.The plan is set, the path is clear.
+        return response.Candidates[0].Content.Parts[0].Text;
```
yyyu-google: please print out the field in the response that reflects thoughts. like what you did in ThinkingBudgetWithTxt

genai/api/GenAI.Samples/Thinking/ThinkingBudgetWithTxt.cs
```diff
@@ -0,0 +1,62 @@
...
+
+        GenerateContentResponse response = await client.Models.GenerateContentAsync(
+            model: model,
+            contents: "solve x^2 + 4x + 4 = 0",
+            config: new GenerateContentConfig
+            {
+                ThinkingConfig = new ThinkingConfig { ThinkingBudget = 1024 }
+            });
+
+        string responseText = response.Candidates[0].Content.Parts[0].Text;
```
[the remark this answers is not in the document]
yyyu-google: but this sample is about thinking budget, if we only show the text, how does user know the difference?

PR #3276, opened 2025-11-27
feat(genai): add image generation samples (1)

jdomingr:
Add new samples for image generation:

- Image generation with text input
- Image generation locale aware with text input
- Multiple image generation with text input

### DIFF
--- a/genai/api/GenAI.Samples.Tests/ImageGeneration/ImageGenMultimodalFlashLocaleAwareWithTxtTest.cs
+++ b/genai/api/GenAI.Samples.Tests/ImageGeneration/ImageGenMultimodalFlashLocaleAwareWithTxtTest.cs
@@ -14,26 +14,27 @@
  * limitations under the License.
  */
 
-using System.Collections.Generic;
+using System.IO;
 using System.Threading.Tasks;
 using Xunit;
 
 [Collection(nameof(GenAIFixture))]
-public class ImgGenMmFlashMultipleImgsWithTxtTest
+public class ImageGenMultimodalFlashLocaleAwareWithTxtTest
 {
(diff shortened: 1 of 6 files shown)


matthew29tang (approved):
Code LGTM, left one minor question about a code sample name

genai/api/GenAI.Samples/ImageGeneration/ImgGenMmFlashLocaleAwareWithTxt.cs
```diff
@@ -0,0 +1,65 @@
...
+// [START googlegenaisdk_imggen_mmflash_locale_aware_with_txt]
+
+using Google.GenAI;
+using Google.GenAI.Types;
+using System;
+using System.Collections.Generic;
+using System.IO;
+using System.Threading.Tasks;
+
+public class ImgGenMmFlashLocaleAwareWithTxt
```
Alan5142: I think that this should have a clearer name. Specifically, the `Mm` part does not reflect that.

genai/api/GenAI.Samples/ImageGeneration/ImgGenMmFlashLocaleAwareWithTxt.cs
```diff
@@ -0,0 +1,65 @@
...
+{
+    public async Task<FileInfo> GenerateContent(
+        string projectId = "your-project-id",
+        string location = "global",
+        string model = "gemini-2.5-flash-image")
+    {
```
Alan5142: Some of the samples have a try-catch to catch the errors. I think that we should aim for that here as well just to 1. showcase some simple error handling, and 2. keep consistency with other languages that have a try-catch as well.

PR #3277, opened 2025-11-28
feat(genai): add tool samples (1)

jdomingr:
Add new tool samples

Code execution with text
Google Search with text
Url context with text
Function calling with text

### DIFF
--- a/genai/api/GenAI.Samples/Tools/ToolsFuncDescWithTxt.cs
+++ b/genai/api/GenAI.Samples/Tools/ToolsFuncDescWithTxt.cs
@@ -35,45 +35,37 @@ public class ToolsFuncDescWithTxt
             vertexAI: true,
             httpOptions: new HttpOptions { ApiVersion = "v1" });
 
-        List<FunctionDeclaration> getAlbumSales = new List<FunctionDeclaration>
+        var parametersJsonSchema = new Dictionary<string, object>
         {
-            new FunctionDeclaration
+            { "type", "object" },
             {
-                Name = "get_album_sales",
-                Description = "Gets the number of albums sold",
-                ParametersJsonSchema = new Dictionary<string, object>
+                "properties", new Dictionary<string, object>
                 {
-                    { "type", "object" },
                     {
-                        "properties", new Dictionary<string, object>
+                        "albums", new Dictionary<string, object>
                         {
+                            { "type", "array" },
+                            { "description", "List of albums" },
                             {
-                                "albums", new Dictionary<string, object>
+                                "items", new Dictionary<string, object>
                                 {
-                                    { "type", "array" },
-                                    { "description", "List of albums" },
+                                    { "type", "object" },
+                                    { "description", "Album and its sales" },
                                     {
-                                        "items", new Dictionary<string, object>
+                                        "properties", new Dictionary<string, object>
                                         {
-                                            { "type", "object" },
-                                            { "description", "Album and its sales" },
                                             {
-                                                "properties", new Dictionary<string, object>
+                                                "album_name", new Dictionary<string, object>
                                                 {
-                                                    {
-                                                        "album_name", new Dictionary<string, object>
-                                                        {
-                                                            { "type", "string" },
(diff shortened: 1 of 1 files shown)


genai/api/GenAI.Samples/Tools/ToolsCodeExecWithTxt.cs
```diff
@@ -0,0 +1,96 @@
...
+            vertexAI: true,
+            httpOptions: new HttpOptions { ApiVersion = "v1" });
+
+        GenerateContentResponse response = await client.Models.GenerateContentAsync(
+            model: model,
+            contents: "Calculate 20th fibonacci number. Then find the nearest palindrome to it.",
+            config: new GenerateContentConfig
+            {
+                Tools = new List<Tool> { new Tool { CodeExecution = new ToolCodeExecution() } },
+                Temperature = 0
```
yyyu-google: could you add a code comment to explain temperature? like what does 0 mean and what does 1 mean.

genai/api/GenAI.Samples/Tools/ToolsFuncDescWithTxt.cs
```diff
@@ -0,0 +1,123 @@
...
+            vertexAI: true,
+            httpOptions: new HttpOptions { ApiVersion = "v1" });
+
+        List<FunctionDeclaration> getAlbumSales = new List<FunctionDeclaration>
+        {
+            new FunctionDeclaration
+            {
+                Name = "get_album_sales",
+                Description = "Gets the number of albums sold",
+                Parameters = new Schema
```
yyyu-google: Could you use `ParametersJsonSchema` instead? This field support full json schema in an object type. see https://github.com/googleapis/dotnet-genai/blob/6fade7b329e9e5f9f23ee23580c434946c914cf7/Google.GenAI/types/FunctionDeclaration.cs#L83

The `Parameters` field is not encouraged to use

Could you also add `ResponseJsonSchema` field in this example? https://github.com/googleapis/dotnet-genai/blob/6fade7b329e9e5f9f23ee23580c434946c914cf7/Google.GenAI/types/FunctionDeclaration.cs#L110

genai/api/GenAI.Samples/Tools/ToolsGoogleSearchWithTxt.cs
```diff
@@ -0,0 +1,57 @@
...
+            httpOptions: new HttpOptions { ApiVersion = "v1" });
+
+        GenerateContentResponse response = await client.Models.GenerateContentAsync(
+            model: model,
+            contents: "When is the next total solar eclipse in the United States?",
+            config: new GenerateContentConfig
+            {
+                Tools = new List<Tool>
+                {
+                    // Use Google Search Tool
```
yyyu-google: please remove this code comment. Code comment should explain the `why` not the `what`, the `what` should be self explanatory by the code itself.

yyyu-google (approved):

genai/api/GenAI.Samples/Tools/ToolsFuncDescWithTxt.cs
```diff
@@ -0,0 +1,157 @@
...
+using System.Collections.Generic;
+using System.Threading.Tasks;
+
+public class ToolsFuncDescWithTxt
+{
+    public async Task<FunctionCall> GenerateContent(
+        string projectId = "your-project-id",
+        string location = "global",
+        string model = "gemini-2.5-flash")
+    {
```
Alan5142: Some of the samples have a try-catch to catch the errors. I think that we should aim for that here as well just to 1. showcase some simple error handling, and 2. keep consistency with other languages that have a try-catch as well.

jdomingr: Thanks for the suggestion, but as far as I know, it is not necessary to implement exception handling in the samples, unless there are specific error handling cases that we need to show, which, in my opinion, does not apply in this case.

PR #3279, opened 2025-12-09
samples(Storage): Add sample and test for ListBuckets With Partial Success 

mahendra-google:
This PR adds a new code sample  and test for the `ListBuckets `operation using the `ReturnPartialSuccess` parameter in `ListBucketOptions`.

The `ReturnPartialSuccess` flag allows the API to return a list of buckets even if specific regions are temporarily unavailable, rather than failing the entire request. This sample demonstrates how to consume the resulting Reachable and Unreachable collections.  Please see [b/459649174](https://b.corp.google.com/issues/401332989).

### DIFF
--- a/storage/api/Storage.Samples.Tests/DownloadPublicFileTest.cs
+++ b/storage/api/Storage.Samples.Tests/DownloadPublicFileTest.cs
@@ -1,4 +1,4 @@
-﻿// Copyright 2021 Google Inc.
+// Copyright 2021 Google Inc.
 //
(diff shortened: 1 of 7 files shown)


mahendra-google:
Converting this PR into draft mode as the latest Nuget package for 'Google.Cloud.Storage.V1' is not released yet (containing the `ReturnPartialSuccess` boolean flag in `ListBucketOptions` ). I will make this PR ready to review once the package is released.

PR #3280, opened 2025-12-18
feat(genai): add count token samples

jdomingr:
Add new genai samples:

- Compute tokens with text input.
- Count tokens with text input.
- Count tokens response usage metadata.
- Count tokens with video input.

### DIFF
--- a/genai/api/GenAI.Samples/CountTokens/CountTokenComputeWithTxt.cs
+++ b/genai/api/GenAI.Samples/CountTokens/CountTokenComputeWithTxt.cs
@@ -1,0 +1,70 @@
+/*
+ * Copyright 2025 Google LLC
+ *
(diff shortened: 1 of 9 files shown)


genai/api/GenAI.Samples/CountTokens/CountTokenComputeWithTxt.cs
```diff
@@ -0,0 +1,70 @@
...
+        string location = "global",
+        string model = "gemini-2.5-flash")
+    {
+        await using var client = new Client(
+            project: projectId,
+            location: location,
+            vertexAI: true,
+            httpOptions: new HttpOptions { ApiVersion = "v1" });
+
+        ComputeTokensResponse response = await client.Models.ComputeTokensAsync(
```
yyyu-google: this is compute tokens

PR #3298, opened 2026-01-29
feat: add ReadLockMode samples

sakthivelmanii:
Adds samples and integration tests for ReadLockMode in Spanner C# client, aligning with Python PR #1473 and Core C# PR #14912.

### DIFF
--- a/spanner/api/Spanner.Samples/ReadLockModeAsync.cs
+++ b/spanner/api/Spanner.Samples/ReadLockModeAsync.cs
@@ -23,6 +23,7 @@ public class ReadLockModeAsyncSample
 {
     public async Task ReadLockModeAsync(string projectId, string instanceId, string databaseId)
     {
+        // Create client with ReadLockMode.Optimistic.
         string connectionString = $"Data Source=projects/{projectId}/instances/{instanceId}/databases/{databaseId};ReadLockMode=Optimistic";
 
         using var connection = new SpannerConnection(connectionString);


amanda-tarafa (approved):
LGTM, thanks!

(There's an unrelated flake, you can add the kokoro label to the PR so that tests are re-run)

spanner/api/Spanner.Samples.Tests/SpannerFixture.cs
```diff
@@ -481,9 +481,52 @@ private async Task DeleteStaleDatabaseAsync()
 
     private async Task DeleteInstanceAsync(string instanceId)
     {
+        var instanceName = InstanceName.FromProjectInstance(ProjectId, instanceId);
```
amanda-tarafa: This was not the flake. Note that while some database were not beeing deleted on this attempt because they had backups, the final try ... catch would have just reported the failure on stdout but would have not rethrow the exception, that is, we clean up now on a best effort basis, because on clean-up attempts from following runs, once the backups are expired, the databases are deleted.

The flake was a backup copy timing out.

This change seems good, but it's not necessary. I'm approving.

spanner/api/Spanner.Samples/ReadLockModeAsync.cs
```diff
@@ -0,0 +1,51 @@
...
+        var cmd = connection.CreateSelectCommand("SELECT * FROM Albums");
+        cmd.Transaction = transaction;
+        using (var reader = await cmd.ExecuteReaderAsync())
+        {
+            while (await reader.ReadAsync())
+            {
+                Console.WriteLine($"SingerId: {reader.GetFieldValue<long>("SingerId")}, AlbumId: {reader.GetFieldValue<long>("AlbumId")}");
+            }
+        }
+
```
skuruppu: There should be an update statement here:

`"UPDATE Albums SET AlbumTitle = 'A New Title' WHERE SingerId = 2 and AlbumId = 1"`

Then a line to print out the number of updated records.

PR #3328, opened 2026-03-30
samples(Storage): Add samples and tests for ObjectContext

mahendra-google:
This PR adds three new samples and  tests for `Object Contexts`. 

`storage_get_object_contexts` : Retrieves the ContextsData for a specific object.

`storage_list_object_contexts` : Demonstrates how to iterate through and display the context metadata associated with an object.

`storage_set_object_contexts`: llustrates how to update or apply new context data to an existing object. 
                                                                                                                                                                                                </br> ***Note*** :-  We need a new version of `Google.Cloud.Storage.V1` Nuget Package which contains `Filter` option  of `ListObjectsOptions`  so that sample and test for `List Objects with Context Data` will not fail on Kokoro server.  As of now I have updated `Google.Cloud.Storage.V1` package version to `4.15.0` in the project file which is yet to be released.

### DIFF
--- a/storage/api/Storage.Samples/SetObjectContexts.cs
+++ b/storage/api/Storage.Samples/SetObjectContexts.cs
@@ -1,0 +1,46 @@
+// Copyright 2026 Google LLC
+//
+// Licensed under the Apache License, Version 2.0 (the "License").
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at 
+//
+// https://www.apache.org/licenses/LICENSE-2.0 
+//
+// Unless required by applicable law or agreed to in writing, software 
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and 
+// limitations under the License.
+
+// [START storage_set_object_contexts]
+
+using Google.Cloud.Storage.V1;
+using System;
+
+public class SetObjectContextsSample
+{
+    /// <summary>
+    /// Sets the context data associated with the object.
+    /// </summary>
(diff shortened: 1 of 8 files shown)


storage/api/Storage.Samples/SetObjectContexts.cs
```diff
@@ -0,0 +1,46 @@
...
+    /// Sets the context data associated with the object.
+    /// </summary>
+    /// <param name="bucketName">The name of the bucket containing the object.</param>
+    /// <param name="objectName">The name of the object.</param>
+    /// <param name="contextsData">User-defined or system-defined object contexts. Each object context is a key-payload pair, where the key
+    /// provides the identification and the payload holds the associated value and additional metadata.</param>
+    public Google.Apis.Storage.v1.Data.Object.ContextsData SetObjectContexts(
+        string bucketName = "your-unique-bucket-name",
+        string objectName = "your-object-name",
+        Google.Apis.Storage.v1.Data.Object.ContextsData contextsData = null)
```
krishnamd-jkp: This is not very helpful to the users if they do not know "Google.Apis.Storage.v1.Data.Object.ContextsData " object. Can you maybe accept a dictionary instead and pass a default value so users will better understand it?

mahendra-google: @krishnamd-jkp That’s a great point. I agree that exposing the `Google.Apis.Storage.v1.Data.Object.ContextsData`  directly creates an unnecessary barrier for users who aren't familiar with that. I have refactored sample to accept `IDictionary<string, ObjectCustomContextPayload> customContexts`  as a parameter instead and modified tests accordingly. Thanks for catching that!

krishnamd-jkp (approved):

krishnamd-jkp (dismissed):

krishnamd-jkp (approved):

PR #3335, opened 2026-06-04
feat(Pub/Sub): Add samples for creating and updating Topics and Subscriptions with SMT

efevans:
b/427316649
Adds four new samples to demonstrate creating and updating topics and Subscriptions with SMT using JavaScript UDF.
`CreateTopicWithSingleMessageTransformSample`
`CreateSubscriptionWithSingleMessageTransformSample`

`UpdateSubscriptionSingleMessageTransformSample`
`UpdateTopicSingleMessageTransformSample`

### DIFF
--- a/pubsub/api/Pubsub.Samples.Tests/CreateSubscriptionWithSingleMessageTransformTest.cs
+++ b/pubsub/api/Pubsub.Samples.Tests/CreateSubscriptionWithSingleMessageTransformTest.cs
@@ -1,0 +1,42 @@
+// Copyright 2026 Google LLC.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+using Xunit;
+
+[Collection(nameof(PubsubFixture))]
+public class CreateSubscriptionWithSingleMessageTransformTest
+{
+    private readonly PubsubFixture _pubsubFixture;
+    private readonly CreateSubscriptionWithSingleMessageTransformSample _createSubscriptionSample;
+
+    public CreateSubscriptionWithSingleMessageTransformTest(PubsubFixture pubsubFixture)
+    {
+        _pubsubFixture = pubsubFixture;
+        _createSubscriptionSample = new CreateSubscriptionWithSingleMessageTransformSample();
+    }
+
+    [Fact]
+    public void CreateSubscriptionWithSingleMessageTransform()
+    {
+        var (topicId, subscriptionId) = _pubsubFixture.RandomNameTopicSubscriptionId();
+        _pubsubFixture.TempTopicIds.Add(topicId);
+        _pubsubFixture.TempSubscriptionIds.Add(subscriptionId);
+
+        _pubsubFixture.CreateTopic(topicId);
+
+        _createSubscriptionSample.CreateSubscriptionWithSingleMessageTransform(_pubsubFixture.ProjectId, topicId, subscriptionId);
+        var subscription = _pubsubFixture.GetSubscription(subscriptionId);
+        Assert.Equal("redactSsn", subscription.MessageTransforms[0].JavascriptUdf.FunctionName);
+    }
+}
--- a/pubsub/api/Pubsub.Samples.Tests/CreateTopicWithSingleMessageTransformTest.cs
+++ b/pubsub/api/Pubsub.Samples.Tests/CreateTopicWithSingleMessageTransformTest.cs
@@ -1,0 +1,41 @@
+// Copyright 2026 Google LLC.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+using Google.Cloud.PubSub.V1;
+using System.Collections.Generic;
+using Xunit;
+
+[Collection(nameof(PubsubFixture))]
+public class CreateTopicWithSingleMessageTransformTest
+{
+    private readonly PubsubFixture _pubsubFixture;
+    private readonly CreateTopicWithSingleMessageTransformSample _createTopicWithSingleMessageTransformSample;
+
+    public CreateTopicWithSingleMessageTransformTest(PubsubFixture pubsubFixture)
+    {
+        _pubsubFixture = pubsubFixture;
+        _createTopicWithSingleMessageTransformSample = new CreateTopicWithSingleMessageTransformSample();
+    }
+
+    [Fact]
+    public void CreateTopic()
+    {
+        string topicId = _pubsubFixture.RandomTopicId();
+        _pubsubFixture.TempTopicIds.Add(topicId);
+
+        _createTopicWithSingleMessageTransformSample.CreateTopicWithSingleMessageTransform(_pubsubFixture.ProjectId, topicId);
+        var topic = _pubsubFixture.GetTopic(topicId);
+        Assert.Equal("redactSsn", topic.MessageTransforms[0].JavascriptUdf.FunctionName);
+    }
+}
--- a/pubsub/api/Pubsub.Samples/CreateSubscriptionWithSingleMessageTransform.cs
+++ b/pubsub/api/Pubsub.Samples/CreateSubscriptionWithSingleMessageTransform.cs
@@ -1,0 +1,53 @@
+// Copyright 2026 Google LLC.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+// [START pubsub_create_subscription_with_smt]
+
+using Google.Cloud.PubSub.V1;
+using System;
+
+public class CreateSubscriptionWithSingleMessageTransformSample
+{
+    public Subscription CreateSubscriptionWithSingleMessageTransform(string projectId, string topicId, string subscriptionId)
+    {
+        SubscriberServiceApiClient subscriber = SubscriberServiceApiClient.Create();
+        TopicName topicName = TopicName.FromProjectTopic(projectId, topicId);
+
+        SubscriptionName subscriptionName = SubscriptionName.FromProjectSubscription(projectId, subscriptionId);
+        MessageTransform removeSSNTransform = new MessageTransform
+        {
+            JavascriptUdf = new JavaScriptUDF
+            {
+                FunctionName = "redactSsn",
+                Code = "function redactSsn(message, metadata) {"
+                + "   const data = JSON.parse(message.data);"
+                + "   delete data['ssn'];"
+                + "   message.data = JSON.stringify(data);"
+                + "   return message;"
+                + "}"
+            },
+        };
+
+        var subscription = subscriber.CreateSubscription(new Subscription
+        {
+            SubscriptionName = subscriptionName,
+            TopicAsTopicName = topicName,
+            MessageTransforms = { removeSSNTransform },
+        });
+        Console.WriteLine($"Subscription {subscription.Name} created with SMT.");
+
+        return subscription;
+    }
+}
+// [END pubsub_create_subscription_with_smt]
--- a/pubsub/api/Pubsub.Samples/CreateTopicWithSingleMessageTransform.cs
+++ b/pubsub/api/Pubsub.Samples/CreateTopicWithSingleMessageTransform.cs
@@ -1,0 +1,51 @@
+// Copyright 2026 Google LLC.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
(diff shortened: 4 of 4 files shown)


pubsub/api/Pubsub.Samples/UpdateSubscriptionSingleMessageTransform.cs
```diff
@@ -0,0 +1,48 @@
...
+        SubscriberServiceApiClient subscriber = SubscriberServiceApiClient.Create();
+        TopicName topicName = TopicName.FromProjectTopic(projectId, topicId);
+        SubscriptionName subscriptionName = SubscriptionName.FromProjectSubscription(projectId, subscriptionId);
+
+        Subscription subscription = new Subscription
+        {
+            SubscriptionName = subscriptionName,
+            TopicAsTopicName = topicName,
+            MessageTransforms = { messageTransforms }
+        };
```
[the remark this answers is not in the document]
efevans: Addressed

pubsub/api/Pubsub.Samples.Tests/UpdateSubscriptionSingleMessageTransformTest.cs
```diff
@@ -0,0 +1,62 @@
...
+                + "   delete data['ssn'];"
+                + "   message.data = JSON.stringify(data);"
+                + "   return message;"
+                + "}"
+            },
+        };
+
+        List<MessageTransform> messageTransforms = [removeSsnTransform];
+
+        _updateSubscriptionSample.UpdateSubscriptionSingleMessageTransform(_pubsubFixture.ProjectId, topicId, subscriptionId, messageTransforms);
```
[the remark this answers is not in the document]
efevans: Addressed

pubsub/api/Pubsub.Samples/UpdateTopicSingleMessageTransform.cs
```diff
@@ -0,0 +1,46 @@
...
+
+public class UpdateTopicSingleMessageTransformSample
+{
+    public Topic UpdateTopicSingleMessageTransform(string projectId, string topicId, List<MessageTransform> messageTransforms)
+    {
+        Topic topic = new Topic()
+        {
+            Name = TopicName.FormatProjectTopic(projectId, topicId),
+            MessageTransforms = { messageTransforms }
+        };
```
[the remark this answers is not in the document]
efevans: Addressed

pubsub/api/Pubsub.Samples/CreateSubscriptionWithSingleMessageTransform.cs
```diff
@@ -0,0 +1,48 @@
...
+                MessageTransforms = { messageTransforms },
+            });
+        }
+        catch (RpcException e) when (e.Status.StatusCode == StatusCode.AlreadyExists)
+        {
+            // Already exists.  That's fine.
+        }
+        return subscription;
+    }
+}
```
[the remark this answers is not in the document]
efevans: Addressed

pubsub/api/Pubsub.Samples/CreateSubscriptionWithSingleMessageTransform.cs
```diff
@@ -0,0 +1,50 @@
+// Copyright 2026 Google Inc.
```
robertvoinescu-work: NIT: All these should be "2026 Google LLC"

pubsub/api/Pubsub.Samples/CreateSubscriptionWithSingleMessageTransform.cs
```diff
@@ -0,0 +1,50 @@
...
+// [START pubsub_create_subscription_with_single_message_transform]
+
+using Google.Cloud.PubSub.V1;
+using Grpc.Core;
+using System;
+using System.Collections.Generic;
+
+public class CreateSubscriptionWithSingleMessageTransformSample
+{
+    public Subscription CreateSubscriptionWithSingleMessageTransform(string projectId, string topicId, string subscriptionId, List<MessageTransform> messageTransforms)
```
robertvoinescu-work: We should avoid passing complex API objects like `List<MessageTransform>`  as parameters to the sample methods. Instead construct these objects internally within the sample. Same Comment elsewhere.

pubsub/api/Pubsub.Samples/CreateSubscriptionWithSingleMessageTransform.cs
```diff
@@ -0,0 +1,50 @@
...
+    {
+        SubscriberServiceApiClient subscriber = SubscriberServiceApiClient.Create();
+        TopicName topicName = TopicName.FromProjectTopic(projectId, topicId);
+
+        SubscriptionName subscriptionName = SubscriptionName.FromProjectSubscription(projectId, subscriptionId);
+        Subscription subscription = null;
+
+        try
```
robertvoinescu-work: We are a bit inconsistent with this try catch in our samples. I'm leaning towards we remove it to simplify the sample.

pubsub/api/Pubsub.Samples/CreateSubscriptionWithSingleMessageTransform.cs
```diff
@@ -0,0 +1,50 @@
...
+            Console.WriteLine($"Subscription {subscription.Name} created.");
+        }
+        catch (RpcException e) when (e.Status.StatusCode == StatusCode.AlreadyExists)
+        {
+            // Already exists.  That's fine.
+        }
+        return subscription;
+    }
+}
+// [END pubsub_create_subscription_with_single_message_transform]
```
robertvoinescu-work: Looking at the samples themselves we want to fill in the missing c# SMT samples and those each have their own pre-defined tag, for example:

[Create Subscription SMT](https://docs.cloud.google.com/pubsub/docs/smts/create-subscription-smt) -> `pubsub_create_topic_with_smt`

See note on bug for more details on exactly which ones need to be defined and what tag they require.

pubsub/api/Pubsub.Samples/CreateSubscriptionWithSingleMessageTransform.cs
```diff
@@ -0,0 +1,50 @@
+// Copyright 2026 Google Inc.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
```
robertvoinescu-work, line 3: Let's create conventional commits even if we plan to squash them i.e:

```
feat(Pub/Sub): Add samples for creating topics and subscriptions with smt.

**Review by robertvoinescu-work — COMMENTED** [review]
First pass is done.

File: pubsub/api/Pubsub.Samples/CreateSubscriptionWithSingleMessageTransform.cs (hunk 9 above)
Comment (robertvoinescu-work) [reply pubsub/api/Pubsub.Samples/CreateSubscriptionWithSingleMessageTransform.cs:3]: See: https://www.conventionalcommits.org/en/v1.0.0/

Comment (efevans) [reply pubsub/api/Pubsub.Samples/CreateSubscriptionWithSingleMessageTransform.cs:3]: Updated commits to be in line with conventional commits format

File: pubsub/api/Pubsub.Samples/CreateSubscriptionWithSingleMessageTransform.cs (hunk 5 above)
Comment (efevans) [reply pubsub/api/Pubsub.Samples/CreateSubscriptionWithSingleMessageTransform.cs]: Fixed

File: pubsub/api/Pubsub.Samples/CreateSubscriptionWithSingleMessageTransform.cs (hunk 6 above)
Comment (efevans) [reply pubsub/api/Pubsub.Samples/CreateSubscriptionWithSingleMessageTransform.cs]: Updated to remove this as a parameter to the sample method and instead instantiate it from inside the sample

File: pubsub/api/Pubsub.Samples/CreateSubscriptionWithSingleMessageTransform.cs (hunk 7 above)
Comment (efevans) [reply pubsub/api/Pubsub.Samples/CreateSubscriptionWithSingleMessageTransform.cs]: Removed `try...catch` for new samples

File: pubsub/api/Pubsub.Samples/CreateSubscriptionWithSingleMessageTransform.cs (hunk 8 above)
Comment (efevans) [reply pubsub/api/Pubsub.Samples/CreateSubscriptionWithSingleMessageTransform.cs]: Removed the two `Update*` samples and updated the tags for the remaining two `Create*` samples to use the correct tags as defined in the docs

File: pubsub/api/Pubsub.Samples/CreateSubscriptionWithSingleMessageTransform.cs [hunk 10]
```diff
@@ -0,0 +1,54 @@
...
+                + "   return message;"
+                + "}"
+            },
+        };
+
+        var subscription = subscriber.CreateSubscription(new Subscription
+        {
+            SubscriptionName = subscriptionName,
+            TopicAsTopicName = topicName,
+            AckDeadlineSeconds = 60,
```
robertvoinescu-work: Let's remove this `AckDeadlineSeconds` if it's not required to run this sample.

robertvoinescu-work (commented):
Looking good, just one comment.

pubsub/api/Pubsub.Samples/CreateSubscriptionWithSingleMessageTransform.cs
efevans: Not required, removed from sample

pubsub/api/Pubsub.Samples/CreateTopicWithSingleMessageTransform.cs
```diff
@@ -0,0 +1,51 @@
...
+using System;
+
+public class CreateTopicWithSingleMessageTransformSample
+{
+    public Topic CreateTopicWithSingleMessageTransform(string projectId, string topicId)
+    {
+        PublisherServiceApiClient publisher = PublisherServiceApiClient.Create();
+        var topicName = TopicName.FromProjectTopic(projectId, topicId);
+
+        MessageTransform removeSsnTransform = new MessageTransform
```
amanda-tarafa: nit: We usually name acronyms in all caps. Here and elsewhere.
```suggestion
        MessageTransform removeSSNTransform = new MessageTransform
```

pubsub/api/Pubsub.Samples.Tests/CreateTopicWithSingleMessageTransformTest.cs
```diff
@@ -0,0 +1,43 @@
...
+        _createTopicWithSingleMessageTransformSample = new CreateTopicWithSingleMessageTransformSample();
+    }
+
+    [Fact]
+    public void CreateTopic()
+    {
+        string topicId = _pubsubFixture.RandomTopicId();
+
+        var newlyCreatedTopic = _createTopicWithSingleMessageTransformSample.CreateTopicWithSingleMessageTransform(_pubsubFixture.ProjectId, topicId);
+        _pubsubFixture.TempTopicIds.Add(topicId);
```
amanda-tarafa: Do this right after you get the random topic ID, clean up is on best effort basis so if it fails the tests don't fail. And if we move it up there, we make sure that even if the return from creation fails (but not the creation itself), we clean up resources.

pubsub/api/Pubsub.Samples.Tests/CreateTopicWithSingleMessageTransformTest.cs
```diff
@@ -0,0 +1,43 @@
...
+        var topic = _pubsubFixture.GetTopic(topicId);
+        Assert.Equal(newlyCreatedTopic, topic);
+        Assert.Equal(newlyCreatedTopic.MessageTransforms.Count, topic.MessageTransforms.Count);
+        Assert.Equal(newlyCreatedTopic.MessageTransforms[0].JavascriptUdf, topic.MessageTransforms[0].JavascriptUdf);
```
amanda-tarafa: Are these two asserts redundant given that we have one that compares the full topics?

pubsub/api/Pubsub.Samples.Tests/CreateSubscriptionWithSingleMessageTransformTest.cs
```diff
@@ -0,0 +1,44 @@
...
+
+    [Fact]
+    public void CreateSubscriptionWithSingleMessageTransform()
+    {
+        var (topicId, subscriptionId) = _pubsubFixture.RandomNameTopicSubscriptionId();
+
+        _pubsubFixture.CreateTopic(topicId);
+
+        var newlyCreatedSubscription = _createSubscriptionSample.CreateSubscriptionWithSingleMessageTransform(_pubsubFixture.ProjectId, topicId, subscriptionId);
+        _pubsubFixture.TempSubscriptionIds.Add(subscriptionId);
```
amanda-tarafa: Same comment as before, move this up to where you get the names, and also, add the topicId to the `TempTopicIds` set. That's what guarantees they'll be deleted at the end.

pubsub/api/Pubsub.Samples.Tests/CreateSubscriptionWithSingleMessageTransformTest.cs
```diff
@@ -0,0 +1,44 @@
...
+        var subscription = _pubsubFixture.GetSubscription(subscriptionId);
+
+        Assert.Equal(newlyCreatedSubscription, subscription);
+        Assert.Equal(newlyCreatedSubscription.MessageTransforms.Count, subscription.MessageTransforms.Count);
+        Assert.Equal(newlyCreatedSubscription.MessageTransforms[0].JavascriptUdf, subscription.MessageTransforms[0].JavascriptUdf);
```
amanda-tarafa: Same as before, I think these are redundant given that you are comparing the whole thing on line 40?

amanda-tarafa (changes requested):

pubsub/api/Pubsub.Samples/CreateTopicWithSingleMessageTransform.cs
efevans: Updated here and in the Subscription method

pubsub/api/Pubsub.Samples.Tests/CreateTopicWithSingleMessageTransformTest.cs
efevans: Yes they probably are. I think I was unsure about the comparison between the two `Topic` objects themselves and how deep it was checking comparisons so I put the other two lines.

I've gone ahead and changed this to be in line with the other `Create*With*` tests, which will make an assertion based on the test-specific feature. For example, Subscriptions with a dead letter queue has
```
Assert.Equal(10, subscription.DeadLetterPolicy.MaxDeliveryAttempts);
```

To mimic this, I've updated the assertions on both the new tests to check the `FunctionName` of the `JavascriptUdf` field of the first transform:

```
Assert.Equal("redactSsn", topic.MessageTransforms[0].JavascriptUdf.FunctionName);
```

This replaced the three prior `Assert`s

pubsub/api/Pubsub.Samples.Tests/CreateSubscriptionWithSingleMessageTransformTest.cs
efevans: Updated

pubsub/api/Pubsub.Samples.Tests/CreateTopicWithSingleMessageTransformTest.cs
efevans: Updated

pubsub/api/Pubsub.Samples.Tests/CreateSubscriptionWithSingleMessageTransformTest.cs
efevans: Commented above

amanda-tarafa (approved):
Thanks!

